PR #31, opened 2021-03-13
Checkpoint

mansishr:
Added functionality for loading and saving Tensorflow and PyTorch models.

### DIFF
--- a/docs/running_the_federation.saveandload.rst
+++ b/docs/running_the_federation.saveandload.rst
@@ -1,0 +1,37 @@
+.. # Copyright (C) 2020-2021 Intel Corporation
+.. # SPDX-License-Identifier: Apache-2.0
+
+.. _running_saveandload:
+
+Saving and Loading Models
+#########################
+
+The models trained in federation (or other pre-trained models) can be saved once the federation is over so that the next time federation is run, the previously saved model is picked up.
+The functionality of saving and loading models are available in these templates:
+
+ - :code:`keras_cnn_mnist`: workspace with a simple `Keras <http://keras.io/>`_ CNN model that will download the `MNIST <http://yann.lecun.com/exdb/mnist/>`_ dataset and train in a federation.
+ - :code:`torch_cnn_mnist`: workspace with a simple `PyTorch <http://pytorch.org>`_ CNN model that will download the `MNIST <http://yann.lecun.com/exdb/mnist/>`_ dataset and train in a federation.
+
+Saving the model
+~~~~~~~~~~~~~~~~
+:code:`save_native` is used to save a tersorflow or pytorch model at the end of federation. For a pytorch model, the model state dict and optimizer state dict are stored.
+
+.. code-block:: console
+
+    $ model.save_native(filepath)
+
+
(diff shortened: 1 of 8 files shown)


openfl-workspace/torch_cnn_mnist/code/pt_cnn.py
```diff
@@ -42,10 +43,23 @@ def __init__(self, device='cpu', **kwargs):
...
-        self.init_network(device=self.device, **kwargs)
-        self._init_optimizer()
-        self.loss_fn = cross_entropy
-        self.initialize_tensorkeys_for_functions()
+
+        path = 'model/saved_model.pth'
+        if os.path.exists(path):
+            # Load the previously trained model
+            self.load_native(filepath=path)
+            if kwargs['loss']:
```
psfoley, line 51: Should be changed to `if 'loss' in kwargs:`

openfl-workspace/torch_cnn_mnist/code/pt_cnn.py
```diff
@@ -42,10 +43,23 @@ def __init__(self, device='cpu', **kwargs):
...
+                self.loss_fn = kwargs['loss']
```
psfoley, line 52: This will set `self.loss_fn` to a string. You should add additional code to check if the provided function is found within `torch.nn`, and then resolve it with `getattr`

openfl-workspace/torch_cnn_mnist/plan/plan.yaml
```diff
@@ -28,6 +28,8 @@ data_loader :
 task_runner :
   defaults : plan/defaults/task_runner.yaml
   template : code.pt_cnn.PyTorchCNN
+  settings :
+    loss :
```
psfoley, line 32: @mansishr I think it would be worth commenting out `loss`, and also provide an example of acceptable values. i.e.:

```    
       settings :
           # built-in loss functions from torch.nn
           # loss : L1Loss
```

docs/running_the_federation.saveandload.rst
```diff
@@ -0,0 +1,37 @@
...
+
+The models trained in federation (or other pre-trained models) can be saved once the federation is over so that the next time federation is run, the previously saved model is picked up.
+The functionality of saving and loading models are available in these templates:
+
+ - :code:`keras_cnn_mnist`: workspace with a simple `Keras <http://keras.io/>`_ CNN model that will download the `MNIST <http://yann.lecun.com/exdb/mnist/>`_ dataset and train in a federation.
+ - :code:`torch_cnn_mnist`: workspace with a simple `PyTorch <http://pytorch.org>`_ CNN model that will download the `MNIST <http://yann.lecun.com/exdb/mnist/>`_ dataset and train in a federation.
+
+Saving the model
+~~~~~~~~~~~~~~~~
+:code:`save_native` is used to save a tersorflow or pytorch model at the end of federation. For a pytorch model, the model state dict and optimizer state dict are stored.
```
psfoley, line 17: Tensorflow spelling

alexey-gruzdev:
With introduction of Interactive API I believe we could drop this with more simple usage case through Interactive API

PR #36, opened 2021-03-17
Use kroki to create svg

maradionov:
Using only mermaid will cause large white boxes around the diagram. Using Korki to pre-create a mermaid image helps to avoid this.


tonyreina (approved):
Thanks! LTGM.

itrushkin:
Jenkins please retry a build

alexey-gruzdev (approved):

PR #45, opened 2021-03-26
Fixing sdist distribution

sarthakpati:
Added azure devops file for CI and auto-uploads (commented out for now).

Fixes #45


openfl-helper:
Can one of the admins verify this patch?

alexey-gruzdev:
Let's rebase a branch to have no conflicting files - [setup.py](https://github.com/intel/openfl/pull/45/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7)

setup.py
```diff
@@ -55,9 +55,11 @@
         'jupyter',
         'ipykernel',
         'flatten_json',
+        'cryptography==3.4.6',
         'cryptography>=3.4.6',
-        'six'  # Remove this after upgrade to torchvision==0.9
+        'six',  # Remove this after upgrade to torchvision==0.9
         # See https://github.com/pytorch/vision/issues/3497
+        'pytest'
```
alexey-gruzdev, line 62: Could we re-use [requirements-test.txt](https://github.com/intel/openfl/blob/develop/requirements-test.txt), instead of adding _pytest_ dependency to setup.py ?

alexey-gruzdev:
@psfoley @tonyreina here we have Azure Pipelines as additional CI service. We need to divide responsibilities between Azure & our Jenkins CIs.

setup.py
sarthakpati: Sure thing! I was putting stuff into `setup` to reduce the amount of code on the Azure pipelines code. 😄

sarthakpati:
> We need to divide responsibilities between Azure & our Jenkins CIs.

I provided the Azure as an example to show the sdist push. We could do it with Jenkins, as well.

alexey-gruzdev:
@sarthakpati I will close this, since after internal discussion with @psfoley we decided to not extend our CI to include Azure pipelines (maybe in a future)

PR #60, opened 2021-04-26
Interactive API beta

igor-davidyuk:
A new Python API was introduced. It allows defining Federated Learning tasks, data loaders, and models in an interactive environment such as Jupyter Notebook taking advantage of the introduced interactive API component.
Defined objects are then serialized and may be transferred to a remote node.

Fixes https://github.com/intel/openfl/issues/13


igor-davidyuk:

alexey-gruzdev:
@igor-davidyuk could you squash commits which are from Igor Davidyuk user, since CLA check failed. Maybe even squash everything into single commit will help to resolve this.

PR #61, opened 2021-04-26
Interactive API beta

### LINKED ISSUE #13: Redesigning Task Runner
- [ ] Not require explicitly setting TensorKeys to track.
- [ ] PyTorch/TensorFlow code can be imported directly into fx.
Define easy to use python API:
     - [x]	Model Provider
     - [x]	Task registering
     - [ ]	Data repository adapter
- [x]	Implement example new task runner in Jupyter notebook
- [x] Build compatibility with new Data Loader interface
Framework adapters for all currently supported frameworks
     - [x] Tensorflow 2
     - [x] PyTorch
- [x] Well defined Interface for user contributed plugins
- [ ] Extensions to CoreTaskRunner
- [ ] Custom task types (beyond ‘train’, ‘validate’)
- [ ] Allow custom function arguments
- [ ] Break tight coupling between model / dataloader to save native model to central node (aggregator) at end of experiment

igor-davidyuk:
A new Python API was introduced. It allows defining Federated Learning tasks, data loaders, and models in an interactive environment such as Jupyter Notebook taking advantage of the introduced interactive API component.
Defined objects are then serialized and may be transferred to a remote node.

Fixes #13


igor-davidyuk:

openfl/interface/interactive_api/experiment.py
```diff
@@ -0,0 +1,335 @@
+# We need assigner interface
+import os
+from copy import deepcopy
+from logging import getLogger
+import functools
+from collections import defaultdict
+from openfl.federated import Plan
+from pathlib import Path
+from importlib import resources
```
psfoley, line 9: This breaks compatibility with python < 3.7. It would be good to investigate whether there's an alternative to this

openfl/interface/interactive_api/experiment.py
```diff
@@ -0,0 +1,335 @@
...
+        )
+        return tensor_dict
+
+    def _prepare_plan(self, model_provider, task_keeper, data_loader,
+                      rounds_to_train,
+                      delta_updates=False, opt_treatment='RESET',
+                      model_interface_file='model_obj.pkl', tasks_interface_file='tasks_obj.pkl',
+                      dataloader_interface_file='loader_obj.pkl'):
+        # Load the default plan
+        with resources.path('openfl.interface.interactive_api', 'plan.yaml') as plan_path:
```
psfoley, line 123: `plan.yaml` in the interactive_api folder does not get copied to this location during openfl installation. It would be good to either move this file to an artifact directory and rework `setup.py` so this gets copied properly, or have it reside in a default workspace directory under `openfl-workspace`. The advantage of the second approach is openfl-workspace continues to be the central location for all plan templates

openfl/interface/workspace.py
```diff
@@ -108,34 +107,41 @@ def export_():
...
-                # 'fx create workspace' and current one.
-                echo(f'Writing {package}=={version} '
-                     f'to {requirements_filename}...')
-                f.write(f'{package}=={version}\n')
-            elif version is None:  # local dependency
-                warn(f'Could not generate requirements for {package}.'
-                     f' Consider installing it manually after workspace'
-                     f' import.')
-    echo(f'{export_requirements_filename} written.')
+    # requirements_filename = 'requirements.txt'
```
psfoley, line 110: Does the removal of this break the existing export mechanism for the legacy API? If not, this commented out code should be removed

igor-davidyuk:
I would like to make sure that metrics returning from tasks that are not declared in the plan are aggregated and treated correctly by the aggregator.

openfl/interface/workspace.py
igor-davidyuk: The new version works. Removed commented code

openfl/interface/interactive_api/experiment.py
igor-davidyuk: Moved default plan for the Interactive python API to `openfl/openfl-workspace/workspace/plan/plans/default/base_plan_interactive_api.yaml`

openfl/interface/interactive_api/experiment.py
igor-davidyuk: Done

igor-davidyuk:
> I would like to make sure that metrics returning from tasks that are not declared in the plan are aggregated and treated correctly by the aggregator.

setup.py
```diff
@@ -56,6 +60,7 @@
         'ipykernel',
         'flatten_json',
         'cryptography>=3.4.6',
+        'cloudpickle',
```
psfoley, line 63: `dill` should probably be added here as well

igor-davidyuk: I do not think we want to include all plugins dependencies into requirements.

alexey-gruzdev (approved):

PR #62, opened 2021-04-27
Allow to specify custom aggregation function

### LINKED ISSUE #40: Simple aggregation schemes on aggregator: geometric mean + median
It would be great to have not only **_Avg_** operator for tensor aggregation, but also few other simple operations like **_Geometric Mean_**, **_Median_**, etc.

itrushkin:
Contents:
 - Removed multiple aggregation functions support.
 - Support for custom functions:
   - User can define a function wrapper by implementing `openfl.component.AggregationFunctionInterface`.
   - `call` function receives:
     - `local_tensors`: Local versions of tensor to aggregate. List of `LocalTensor` objects with fields:
        - `col_name`: collaborator name
        - `tensor`: tensor
        - `weight`: number of collaborator local data sample size divided by overall sample size.
     -  `db_iterator` : iterator over history of all tensors.
                Columns: `['tensor_name', 'round', 'tags', 'nparray']`
     - `tensor_name`: name of the tensor.
     - `fl_round`: round number.
     - `tags`: tuple of tags for this tensor.

Closes #40.


openfl/databases/tensor_db.py
```diff
@@ -11,6 +11,24 @@
 from openfl.utilities import TensorKey
 
 
+def weighted_average(tensors, weights):
```
alexey-gruzdev, line 14: Could we use type hints to make it more clearer for user how to define such kind of functions?
```
def weighted_average(tensors: np.array, weights: np.array): -> np.array
```

openfl/databases/tensor_db.py
```diff
@@ -11,6 +11,24 @@
...
+    """Compute simple weighted average (FedAvg operation)."""
+    return np.average(tensors, weights=weights, axis=0)
+
+
+def geometric_mean(tensors, *_):
+    """Compute geometric mean."""
+    from scipy.stats import gmean
+    result = gmean(tensors, axis=0)
+    print(f'gmean({tensors})={result}')
```
alexey-gruzdev, line 23: Line, which was used for debug?

openfl/federated/plan/plan.py
```diff
@@ -29,7 +29,7 @@ class Plan(object):
     def Load(yaml_path: Path, default={}):
         """Load the plan from YAML file."""
         if yaml_path and yaml_path.exists():
-            return safe_load(yaml_path.read_text())
+            return yaml.load(yaml_path.read_text(), Loader=yaml.Loader)
```
psfoley, line 32: This should be changed back to `safe_load`

itrushkin: `yaml.Loader` allows to serialize and deserialize Python objects like classes and functions. If we want to change some behavior (aggregation functions), we probably want to override a default function with a custom function by passing it to `override_config` keyword argument of `run_experiment` native method. This argument will replace a default plan setting with `!!python/name...` string that can only be deserialized via `yaml.Loader`. I wonder what was the point of `safe_load` here?

itrushkin: I fixed this by skipping the dumping of the updated plan in the native API.

openfl/native/native.py
```diff
@@ -98,16 +98,16 @@ def update_plan(override_config):
...
-    flat_plan_config = get_plan(return_complete=True)
+    plan, flat_plan_config = get_plan(return_complete=True)
     for k, v in override_config.items():
         if k in flat_plan_config:
-            flat_plan_config[k] = v
             logger.info(f'Updating {k} to {v}... ')
         else:
-            logger.info(f'Key {k} not found in plan. Ignoring... ')
-    plan_config = unflatten(flat_plan_config, '.')
-    Plan.Dump(Path(plan_path), plan_config)
```
itrushkin, line 110: @psfoley The fix for YAML loading is here. The solution is that we don't dump updated plan to disk but just keep it as an object in memory. So even if we use an updated plan in the experiment, `plan/plan.yaml` will always contain default (non-overridden) Plan settings. My concern here is that we will never get an updated version of the Plan during the `fx workspace export` of this workspace. Shall we support exporting of workspaces created for native API calls?

itrushkin:
Jenkins please retry a build

itrushkin:
Jenkins please retry a build

docs/overriding_agg_fn.rst
```diff
@@ -0,0 +1,75 @@
...
+            for record in db_iterator:
+                if (
+                    record['round'] == (fl_round - 1)
+                    and record['tensor_name'] == tensor_name
+                    and 'aggregated' in record['tags']
+                ):
+                    previous_tensor_value = record['nparray']
+            for _, tensor in agg_tensor_dict.items():
+                prev_tensor = previous_tensor_value if previous_tensor_value is not None else tensor
+                delta = prev_tensor - tensor
```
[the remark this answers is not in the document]
maradionov: I think yes, but not fully sure at that moment

alexey-gruzdev (approved):

PR #63, opened 2021-04-28
Enable compression options

psfoley:
This PR:
- Fixes some of the lossy compression logic around how delta updates are applied
- Provides a new example for a KMeans pipeline applied to the Keras CNN template (*keras_cnn_with_compression*)
- Fixes a bug in the sparsity pipelines (applies to the SKC and STC pipelines)
- Adds documentation for compression pipelines under advanced topics

### DIFF
--- a/openfl-workspace/torch_cnn_histology/plan/plan.yaml
+++ b/openfl-workspace/torch_cnn_histology/plan/plan.yaml
@@ -36,3 +36,6 @@ tasks:
 
 assigner:
   defaults: plan/defaults/assigner.yaml
+
+compression_pipeline :
+  defaults : plan/defaults/compression_pipeline.yaml
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -320,9 +320,6 @@ class Aggregator:
         )
         tensor_name, origin, round_number, report, tags = tensor_key
 
-        # send_model_deltas = False
-        #compress_lossless = not self.tensor_codec.compression_pipeline.is_lossy()
-
         if 'aggregated' in tags and 'delta' in tags and round_number != 0:
             # send_model_deltas = True
             agg_tensor_key = TensorKey(
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -11,7 +11,6 @@ from openfl.databases import TensorDB
 
 from openfl.protocols import utils
 from openfl.protocols import ModelProto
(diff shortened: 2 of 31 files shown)


openfl-workspace/torch_cnn_histology/plan/plan.yaml
```diff
@@ -36,3 +36,6 @@ tasks:
 
 assigner:
   defaults: plan/defaults/assigner.yaml
+
+compression_pipeline :
```
tonyreina, line 40: If there's an existing plan.yaml will it fail given that it doesn't have the "compression_pipeline" field?

openfl/component/aggregator/aggregator.py
```diff
@@ -510,7 +512,7 @@ def _process_named_tensor(self, named_tensor, collaborator_name):
             tuple(named_tensor.tags)
         )
         tensor_name, origin, round_number, report, tags = tensor_key
-        assert ('compressed' in tags or 'lossy_decompressed' in tags), (
+        assert ('compressed' in tags or 'lossy_compressed' in tags), (
```
tonyreina, line 515: We use enumerated lists like OptTreatment.RESET. Do you think we should do the same here? It's your choice. I was thinking of CompressionType.Lossy or something like that.

alexey-gruzdev:
@psfoley I believe unit-tests should be fixed  - see [CI output](http://213.221.44.203/job/Federated-Learning/job/precommit/176/execution/node/56/log/)

openfl-workspace/torch_cnn_histology/plan/plan.yaml
psfoley: No, it will default to `NoCompressionPipeline`

openfl/component/aggregator/aggregator.py
psfoley: Fair question. The primary argument that I have against a dedicated enum for compression is that it complicates the tensorkey definition. The reason this label is maintained in tags instead of a dedicated field is that compression is typically only applied during tensor transport. To be consistent with the other tags ('delta', 'model', etc.) it should probably be kept as is for now. I'm not opposed to reformulating this later though

tonyreina: No worries.

alexey-gruzdev:
@psfoley let's rebase it and resolve conflicts in `plan.py` ASAP, since I would like to merge it for upcoming release.

alexey-gruzdev (approved):

PR #79, opened 2021-05-14
Custom aggregation functions in CLI

itrushkin:
We need to support the overriding aggregation function in CLI using the plan YAML file. The idea is to specify module paths to pre-defined `AggregationFunctionInterface` implementations in `aggregation_type` section of specific task.

No changes required for Python native API.
/cc @maradionov

### DIFF
--- a/openfl/component/aggregation_functions/weighted_average.py
+++ b/openfl/component/aggregation_functions/weighted_average.py
@@ -5,7 +5,6 @@
 
 from .interface import AggregationFunctionInterface
 import numpy as np
-from openfl.utilities import Singleton
 
 
 def weighted_average(tensors, weights):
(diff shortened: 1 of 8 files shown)


openfl/component/aggregation_functions/weighted_average.py
```diff
@@ -15,6 +15,15 @@ def weighted_average(tensors, weights):
...
     """Weighted average aggregation."""
 
+    _instance = None
+
+    def __new__(cls, *args, **kwargs):
+        """Use the singleton instance if it has already been created."""
+        if not isinstance(cls._instance, cls):
+            cls._instance = object.__new__(cls, *args, **kwargs)
+
+        return cls._instance
```
dmitryagapov, line 25: It might be better to implement the Singleton metaclass so that it can easily add to any class. What do you think?

openfl/utilities/types.py
```diff
@@ -10,3 +10,15 @@
 
 Metric = namedtuple('Metric', ['name', 'value'])
 LocalTensor = namedtuple('LocalTensor', ['col_name', 'tensor', 'weight'])
+
+
+class Singleton:
```
dmitryagapov, line 15: https://github.com/zcutlip/py-singleton-metaclasses/blob/master/pysingleton/singleton.py
Could you check this implementation?

PR #82, opened 2021-05-17
Initial test experiments for new python API

aleksandr-mokrov:
There are two examples (tensorflow mnist and pytorch kvasir unet) for new interactive API in single file mode (like Jupiter notebook) and two same examples with structured code. 
Also, there are some fixes (use tmp dir outside current project dir to create workspace archive, additional ignore patterns and pattern 'data' was excluded, because it exclude 'data' directory for pytorch example - probably it should be discussed).


dmitryagapov:
Jenkins please retry a build

PR #89, opened 2021-05-21
Logging updates

maradionov:
Add new log level METRIC (between INFO and WARNING) to print only metric information. Send metrics to tensorboard. Add example to PyTorch MNIST notebook.


openfl-workspace/default/plan/plan.yaml
```diff
@@ -9,6 +9,7 @@ aggregator :
     best_state_path : save/best.pbuf
     last_state_path : save/last.pbuf
     rounds_to_train : 10
+    tensor_board    : false
```
itrushkin:
```suggestion
    tensorboard    : false
```
We probably should preserve naming from [tensorflow/tensorboard](https://github.com/tensorflow/tensorboard) and [lanpa/tensorboardX](https://github.com/lanpa/tensorboardx).

openfl-workspace/torch_cnn_mnist/plan/plan.yaml
```diff
@@ -9,6 +9,7 @@ aggregator :
     best_state_path : save/torch_cnn_mnist_best.pbuf
     last_state_path : save/torch_cnn_mnist_last.pbuf
     rounds_to_train : 10
+    tensor_board    : true
```
itrushkin:
```suggestion
    tensorboard    : true
```

openfl/component/aggregator/aggregator.py
```diff
@@ -103,6 +105,7 @@ def __init__(self,
         self.collaborator_tasks_results = {}
         # {TaskResultKey: data_size}
         self.collaborator_task_weight = {}
+        self.tb_writer = SummaryWriter(self.log_dir, flush_secs=10) if tensor_board else None
```
itrushkin: Could users pass `SummaryWriter` instance instead of the flag deciding the TensorBoard usage? It would allow users to define the logging directory, flush frequency, and other keyword arguments of `SummaryWriter`. Moreover, in this case, we do not need `tensorboardX` as a dependency.

openfl/component/aggregator/aggregator.py
```diff
@@ -41,6 +42,7 @@ def __init__(self,
                  assigner,
 
                  rounds_to_train=256,
+                 tensor_board=False,
```
maradionov:
```suggestion
                 tensorboard=False,
```

openfl/component/aggregator/aggregator.py
maradionov:
```suggestion
        self.tb_writer = SummaryWriter(self.log_dir, flush_secs=10) if tensorboard else None
```

openfl/utilities/utils.py
```diff
@@ -3,6 +3,40 @@
...
+    Args:
+        levelName: name of log level.
+        levelNum: log level value.
+        methodName: log method wich will use new log level (default = levelName.lower())
+
+    """
+    if not methodName:
+        methodName = levelName.lower()
+
+    if hasattr(logging, levelName):
```
aleksandr-mokrov: It looks redundant.  We don't catch this exception and I think that it is not a problem if we would try add the same level twiсe into logging. In this case, it is enough to exit.

openfl-workspace/default/plan/plan.yaml
alexey-gruzdev: Agree with @itrushkin for naming

openfl/utilities/utils.py
maradionov: done 3ed9f513adc09160ee04a63eb3e910b3b0fdfa65

openfl-workspace/default/plan/plan.yaml
```diff
@@ -9,6 +9,7 @@ aggregator :
     best_state_path : save/best.pbuf
     last_state_path : save/last.pbuf
     rounds_to_train : 10
+    tensorboard    : false
```
dmitryagapov: align familiare with other strings

openfl-workspace/torch_cnn_mnist/plan/plan.yaml
```diff
@@ -9,6 +9,7 @@ aggregator :
     best_state_path : save/torch_cnn_mnist_best.pbuf
     last_state_path : save/torch_cnn_mnist_last.pbuf
     rounds_to_train : 10
+    tensorboard    : true
```
dmitryagapov: align familiar with other strings

openfl/interface/cli.py
```diff
@@ -7,35 +7,40 @@
 from click import echo, option, pass_context, style
 from sys import argv
 from pathlib import Path
+import os
 
+from openfl.utilities import addLoggingLevel
```
dmitryagapov: https://www.python.org/dev/peps/pep-0008/#imports
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.

The same recommendation for all imports.

openfl/interface/cli.py
```diff
@@ -7,35 +7,40 @@
...
-            'error': ERROR,
-            'critical': CRITICAL
-        }
+    if isinstance(level, str):
+        level = level.upper()
 
-    if level.lower() in ['debug', 'error']:
-        os.environ['GRPC_VERBOSITY'] = level.upper()
+    if level in ['DEBUG', 'ERROR']:
+        os.environ['GRPC_VERBOSITY'] = level
```
dmitryagapov: When you use isinstance it looks like you want to handle both types of level `int` and `str`, as it used in logging module, but here only `str` expected and handeled on line 29.

openfl/native/native.py
```diff
@@ -39,12 +41,13 @@ def setup_plan(log_level='CRITICAL'):
     cols_config = 'plan/cols.yaml'
     data_config = 'plan/data.yaml'
 
+    curent_level = logging.root.level
```
dmitryagapov: *current :)

openfl/utilities/utils.py
```diff
@@ -3,6 +3,33 @@
 """Utilities module."""
 
 import numpy as np
+import logging
+
+
+def addLoggingLevel(levelName, levelNum, methodName=None):
```
dmitryagapov: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

alexey-gruzdev:
@maradionov let's also check how it would work with new interactive API.

openfl/component/aggregator/aggregator.py
maradionov: resolve 747b5c53fee5da2143a617db0311e523c7c937d9

openfl/interface/cli.py
maradionov: solved 95fcf3916295687c77d7d34fc04074389b034e59

openfl/interface/cli.py
maradionov: solved 79e053176690b2e97bcb12e9e5a66314d211dc02

openfl/utilities/utils.py
maradionov: solved 79e053176690b2e97bcb12e9e5a66314d211dc02

openfl/native/native.py
maradionov: solved 79e053176690b2e97bcb12e9e5a66314d211dc02

alexey-gruzdev:
Closed this since we've merged it to #104

PR #91, opened 2021-05-25
issues/75/FX autocomplete feature

### LINKED ISSUE #75: FX autocomplete feature
It would be great to have `fx` + `TAB` combination (like standard Linux autocomplete) feature for current CLI.

dmitryagapov:
Fixes #75


alexey-gruzdev:
@mansishr FYI

openfl/interface/cli.py
```diff
@@ -67,8 +68,14 @@ def format_help(self, ctx, formatter):
             '[args]'
         ]
 
-        formatter.write(style(
-            'CORRECT USAGE\n\n', bold=True, fg='bright_black'))
+        formatter.write(style('BASH COMPLETE ACTIVATION\n\n', bold=True, fg='bright_black'))
+        formatter.write(
+            '1. Run in terminal\n'
+            '   _FX_COMPLETE=source_bash fx >> ~/.bashrc\n'
```
psfoley: @dmitryagapov What we currently recommend in the documentation is that users create a virtual environment for OpenFL. If the virtual environment is not sourced on shell start, the `fx` command will not be found, and autocomplete will not work when the virtual environment is entered in the same shell session. A workaround could be to advise users to create an `fx-autocomplete.sh` file that they can source manually. I.e.:

```
    '1. Run in terminal\n'
    '   _FX_COMPLETE=source_bash fx >> ~/.fx-autocomplete.sh\n'
    ' source ~/.fx-autocomplete.sh'
```

psfoley:
@dmitryagapov This is a great contribution....autocomplete now takes <0.5 seconds in my tests! Once the one open point is addressed, this can be approved and merged.

openfl/interface/cli.py
psfoley: After doing some offline investigation into this, due to the bash completion command change in the latest click library the previous code block should be changed to the following:

```
    '1. Run in terminal\n'
    '   _FX_COMPLETE=bash_source fx >> ~/.fx-autocomplete.sh\n'
    ' source ~/.fx-autocomplete.sh'
```
And the `Click` requirement in [setup.py](https://github.com/intel/openfl/blob/develop/setup.py#L51) should be changed to `'Click>=8.0',`

dmitryagapov: @psfoley Thanks a lot! I didn't notice that `source_bash` was changed to `bash_source`.

PR #95, opened 2021-05-28
Use epochs instead  batch_num. Log current epoch number

maradionov:
Some workspaces used num_batches value for global number of batches. Now all of them use epochs number


openfl/federated/task/runner_keras.py
```diff
@@ -60,7 +60,8 @@ def rebuild_model(self, round_num, input_tensor_dict, validation=False):
         else:
             self.set_tensor_dict(input_tensor_dict, with_opt_vars=False)
 
-    def train(self, col_name, round_num, input_tensor_dict, metrics, num_batches=None, **kwargs):
+    def train(self, col_name, round_num, input_tensor_dict,
+              metrics, epochs=1, batch_size=1, **kwargs):
         """
         Perform the training for a specified number of batches.
```
itrushkin: Docstring should be updated since `num_batches` is not specified anymore.

openfl/federated/task/runner_pt.py
```diff
@@ -129,7 +129,7 @@ def validate(self, col_name, round_num, input_tensor_dict,
         return output_tensor_dict, {}
 
     def train_batches(self, col_name, round_num, input_tensor_dict,
-                      num_batches=None, use_tqdm=False, **kwargs):
+                      num_batches=None, use_tqdm=False, epochs=1, **kwargs):
```
itrushkin: Could `num_batches` be removed here as well?

openfl/federated/task/runner_tf.py
```diff
@@ -84,7 +84,7 @@ def rebuild_model(self, round_num, input_tensor_dict, validation=False):
             self.set_tensor_dict(input_tensor_dict, with_opt_vars=False)
 
     def train_batches(self, col_name, round_num, input_tensor_dict,
-                      num_batches, use_tqdm=False, **kwargs):
+                      epochs=1, use_tqdm=False, **kwargs):
         """
         Perform the training for a specified number of batches.
```
itrushkin: Docstring should be updated since `num_batches` is not specified anymore.

PR #114, opened 2021-06-28
commenting out current saving of best model (based on any val score b…

brandon-edwards:
…eing greater than seen previously for that score. This is due to the fact that we

only want some scores to be considered when saving best model, and some
scores are worse when greater.


brandon-edwards:
Commenting out current saving of best model (based on any val score being greater than seen previously for that score) by aggregator. This is due to the fact that we
    only want some scores to be considered when saving best model, and some
    scores are worse when greater. In the long term, we should come up with a good solution for this, but for the challenge I am implementing best model savings in experiment.py (in FeTS/Challenge, outside of openfl)

PR #115, opened 2021-07-02
Allow IP-based Communication

TortoiseHam:
This PR enables IP-based connection in addition to connecting based on registered FQDNs. It also corrects a bug which was leading to the use of incorrect default plan .yaml files.


openfl-helper:
Can one of the admins verify this patch?

TortoiseHam:

alexey-gruzdev:
Jenkins, please retry a build

alexey-gruzdev:
Jenkins, please retry a build

alexey-gruzdev:
ok to test

alexey-gruzdev:
@TortoiseHam thanks for the contribution! We will review your PR. Could you please fix [flake8 issues](http://213.221.44.203/job/Federated-Learning/job/precommit/400/execution/node/27/log/) found by our CI ?

openfl/cryptography/ca.py
```diff
@@ -112,6 +112,10 @@ def sign_certificate(csr, issuer_private_key, issuer_name, days_to_expiration=36
     builder = builder.add_extension(
         x509.BasicConstraints(ca=ca, path_length=None), critical=True,
     )
+    try:
+        builder = builder.add_extension(csr.extensions.get_extension_for_oid(ExtensionOID.SUBJECT_ALTERNATIVE_NAME).value, critical=False)
+    except ExtensionNotFound as e:
+        pass  # Might not have alternative name
```
itrushkin, line 124: This is the only change required for IP-based communication to work.

openfl/federated/plan/plan.py
```diff
@@ -97,7 +97,7 @@ def parse(plan_config_path: Path, cols_config_path: Path = None,
                 defaults = plan.config[section].get(DEFAULTS)
 
                 if defaults is not None:
-                    defaults = WORKSPACE / 'workspace' / defaults
+                    defaults = join(plan_config_path.absolute().parent.parent, defaults)
```
itrushkin: This path will work in the same way for CLI because `defaults` directory gets copied from `site-packages/openfl/openfl-workspace/workspace/plan/defaults`. But any change in copied files wouldn't affect the behavior because configuration files were parsed from the source `site-packages` directory. This change could be a solution to the problem if our new interactive API copied default config files as well. But it doesn't, so you can check that `openfl-tutorials/interactive_api_tutorials_(experimental)/Pytorch_Kvasir_UNET_workspace/new_python_api_UNET.ipynb` notebook is not working anymore. We are going to skip copying default configuration files during workspace creation in #116 to make things consistent, and this change will not be relevant anymore. This would probably lead to a misunderstanding of what files are actually taken for full plan generation because the workspace folder will no longer store files that `plan.yaml` default values point to.

itrushkin (changes requested):

openfl/cryptography/ca.py
TortoiseHam: The other's are to ensure that the intermediate output during plan generation actually honor the FQDN specified by the user, to reduce confusion when debugging

openfl/federated/plan/plan.py
TortoiseHam: this change ensures that the yaml files which are actually stored in the 'local' folder structure are the ones that are used. It's extremely confusing otherwise since the main yaml file points to, for example, plan/defaults/aggregator.yaml, but any changes to that file locally get ignored in favor of the hidden 'global' aggregator.yaml.

It's convenient having access to all of the default yaml files locally for easier mental modeling of what is going to happen when the final file is constructed, but having them be secretly non-overwritable leads to a lot of wasted debugging time

.gitignore
```diff
@@ -5,3 +5,4 @@ __pycache__
 .vscode
 .ipynb_checkpoints
 venv/*
+.idea
```
itrushkin: Add a newline at the end of the file.

openfl/utilities/utils.py
```diff
@@ -1,10 +1,28 @@
...
+    """
+    Get the system FQDN, with priority given to environment variables.
+
+    Args:
+        name: The name from which to extract the FQDN.
+
+    Returns:
+        The FQDN of the system.
+    """
+    fqdn = os.environ.get('FQDN', None)
```
itrushkin, line 20: What is this environment variable for? If it is used for the simplification process of federation initialization, we need to describe it in our documentation for users to know of this feature.

TortoiseHam: The current system ignores a manually specified FQDN until near the end of the process, which as an end user is not very re-assuring (maybe it is set wrong in a certificate somewhere, who knows?). This way if the user wants to force a particular FQDN they can have confidence that it is used everywhere.

itrushkin: Would you mind adding a description of that behavior to our documentation?

TortoiseHam: added

docs/running_the_federation.baremetal.rst
```diff
@@ -84,7 +84,7 @@ On the Aggregator
...
+    This will initialize the plan and autopopulate the `fully qualified domain name (FQDN) <https://en.wikipedia.org/wiki/Fully_qualified_domain_name>`_ of the aggregator node. This FQDN is embedded within the plan so the collaborators know the externally accessible aggregator server address to connect to. If you face connection issues with the autopopulated FQDN in the plan, this value can be overridden with the :code:`-a` flag, for example :code:`fx plan initialize -a aggregator-hostname.internal-domain.com`. Alternatively, you can override the apparent FQDN of the system by setting an FQDN environment variable (:code:`export FQDN=x.x.x.x`) before invoking :code:`fx plan initialize`.
```
itrushkin, line 87: There are 2 more commands in *Hello Federation* sequence that can use `FQDN` environment variable:
 - `fx aggregator generate-cert-request`
 - `fx aggregator certify`

Notes for them also should be extended. You can find corresponding steps on [Configuring the Federation](https://openfl.readthedocs.io/en/latest/running_the_federation.certificates.html#on-the-aggregator-node) docs page.

alexey-gruzdev:
Jenkins please retry a build

docs/running_the_federation.baremetal.rst
TortoiseHam: Done

alexey-gruzdev:
Jenkins please retry a build

openfl/federated/plan/plan.py
itrushkin: Hi @TortoiseHam! We discussed with the team internally that it would be better to provide the user with full plan specifications by copying a single `plan.yaml` file. It will not have any `default: /path/to/yaml` keys but will have all of the default files' settings. Thus, the user will know what can be overridden. It is going to be added in #116 where further discussion on this issue can be continued. In light of what we described, I believe this change should be reverted in this PR.

TortoiseHam: alternatively, if you're going to be overhauling that system anyway, can you just take this pr and then apply whatever final decisions later?

itrushkin: As I said in 1st comment to this change, applying it will lead to unstable behavior in develop branch due to broken interactive API tutorials where default Plan files are not copied to the workspace at all.

TortoiseHam: alright, done

itrushkin (approved):
Thank you for your contribution! Great work!

alexey-gruzdev:
@TortoiseHam Thanks for the contribution!:)

PR #116, opened 2021-07-06
[CLI] Skip copying default plan files

itrushkin:
Copied files are not used. Default YAML files for Plan generation are taken from OpenFL installation directory.


itrushkin:
I'm going to implement additional Plan parsing logic:
- Only plan.yaml file is copied during workspace creation
- Settings from default files apply first, then settings from workspace template
- User will have a single plan file with full settings specification

All of this is intended to reduce user confusion while dealing with YAML files.

alexey-gruzdev:
@psfoley please take a look.

PR #133, opened 2021-08-02
Change variable disable_tls to tls to avoid twice no like "if not dis…

aleksandr-mokrov:
Variable `disable_tls` was renamed to `tls` to avoid twice no like "if not disable_tls".
For CLI command it is possible to use the same old flag `--disable-tls` to disable mTLS.


openfl/interface/director.py
```diff
@@ -31,15 +31,15 @@ def director(context):
 @director.command(name='start')
 @option('-c', '--director-config-path', default='director.yaml',
         help='The director config file path', type=ClickPath(exists=True))
-@option('--disable-tls', default=False,
-        is_flag=True)
+@option('--tls/--disable-tls', default=True,
```
psfoley, line 34: This should just be `--tls`

openfl/interface/envoy.py
```diff
@@ -33,8 +33,8 @@ def envoy(context):
         help='Current shard name')
 @option('-d', '--director-uri', required=True,
         help='The FQDN of the federation director')
-@option('--disable-tls', default=False,
-        is_flag=True)
+@option('--tls/--disable-tls', default=True,
```
psfoley, line 36: Same comment as above. This should just be `@option('--tls', default=True,`

psfoley (changes requested):
The changes are self explanatory. Once my two comments are addressed, I think this is ready for merge.

openfl/interface/director.py
aleksandr-mokrov: By default you can use `fx director start`  without TLS flag and variable `tls` will be set to `True`. But if you want to disable TLS, you can use `fx director start --disable-tls` and variable `tls` will be set to `False`. In other words, we disable mTLS only by explicitly passing the corresponding flag, by default mTLS is enabled.
[Click boolean flags specification](https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags)

igor-davidyuk (approved):
seems legit

psfoley (approved):
Comments wer addressed. PR is approved.

PR #134, opened 2021-08-02
Chunked implementation of SHA384 hash function

itrushkin:


openfl/utilities/utils.py
```diff
@@ -170,3 +171,26 @@ def split_tensor_dict_for_holdouts(logger, tensor_dict,
...
+            of the file to be opened or an integer file descriptor of the file to be wrapped.
+
+    Returns:
+        str: SHA384 hash value.
+    """
+    h = hashlib.sha384()
+    with open(file_path, 'rb') as file:
+        while True:
+            # Reading is buffered, so we can read smaller chunks.
+            chunk = file.read(h.block_size)
```
aleksandr-mokrov: It's only 128 bytes. It look's like a very small buffer size for reading. It can severely degrade performance. It should be checked, but values are more commonly used a from 4kb to 64kb.

aleksandr-mokrov (approved):

PR #135, opened 2021-08-03
Redesign stream metrics

igor-davidyuk:
Removed async from the director.
Introduced metric_dict structure to the aggregator.
Fixed get_trained_model so only one model is returned

### DIFF
--- a/openfl/component/director/director.py
+++ b/openfl/component/director/director.py
@@ -143,9 +143,10 @@ class Director:
         while True:
             if not aggregator.metric_queue.empty():
                 yield aggregator.metric_queue.get()
+                continue
 
             if aggregator.all_quit_jobs_sent():
-                raise StopIteration
+                return
 
             yield None
 
--- a/openfl/component/director/director.py
+++ b/openfl/component/director/director.py
@@ -81,19 +81,25 @@ class Director:
 
         return True
 
-    async def get_trained_model(self, experiment_name: str, caller: str):
-        """Get trained models."""
+    def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
+        """Get trained model."""
         if experiment_name not in self.experiment_stash[caller]:
-            logger.error('Aggregator has not started yet')
-            return None, None
+            logger.error('No experiment data in the stash')
+            return None
 
         aggregator = self.experiment_stash[caller][experiment_name].aggregator
 
         if aggregator.last_tensor_dict is None:
             logger.error('Aggregator have no aggregated model to return')
-            return None, None
+            return None
 
-        return aggregator.best_tensor_dict, aggregator.last_tensor_dict
+        if model_type == 'best':
+            return aggregator.best_tensor_dict
+        elif model_type == 'last':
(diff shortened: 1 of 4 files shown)


openfl/transport/grpc/director_server.py
```diff
@@ -206,22 +208,29 @@ async def StreamMetrics(self, request, context):  # NOQA:N802
...
 
-        async for message in \
-                self.director.stream_metrics(request.experiment_name, request.header.sender):
-            yield message
+        while True:
+            metric_dict = self.director.get_next_metric(
+                experiment_name=request.experiment_name, caller=request.header.sender
+            )
+            if len(metric_dict) == 0:
+                await asyncio.sleep(5)
```
aleksandr-mokrov, line 218: It looks like we are bringing internal logic to protocol-transport layer. Why are we waiting several seconds when we got empty dict? Why do empty dict and None lead to different logic?

igor-davidyuk: Please refer to the `get_next_metric` method's docstring. The signature of the method is explained there.

aleksandr-mokrov: I mean, these questions are talking about taking out internal logic

openfl/transport/grpc/director_server.py
```diff
@@ -206,22 +208,29 @@ async def StreamMetrics(self, request, context):  # NOQA:N802
...
+                continue
+            if metric_dict is None:
+                break
```
dmitryagapov: We should check that `metric_dict is None` before we are checking `len(metric_dict) == 0` because `len(None)` raise exception `TypeError: object of type 'NoneType' has no len()`

dmitryagapov (changes requested):

openfl/transport/grpc/director_server.py
igor-davidyuk: Changed the get_next_metric method

openfl/component/director/director.py
```diff
@@ -114,33 +120,40 @@ def get_registered_shards(self) -> list:
...
+        Raises:
+            StopIteration - if the experiment is finished and there is no more metrics to report
+        """
         aggregator = self.experiment_stash[caller][experiment_name].aggregator
+        while True:
+            if not aggregator.metric_queue.empty():
+                yield aggregator.metric_queue.get()
+                continue
+
+            if aggregator.all_quit_jobs_sent():
```
aleksandr-mokrov: We should also check that `aggregator.metric_queue.empty()` here, because after last check it could be a switch to the thread in which the aggregator was executed and new data got into the queue.

openfl/transport/grpc/director_server.py
igor-davidyuk: Code was altered. Please check out the changes.

openfl/component/director/director.py
igor-davidyuk: done!

aleksandr-mokrov (approved):

dmitryagapov (approved):

PR #136, opened 2021-08-03
move get experiment info in separate function

dmitryagapov:


openfl/transport/grpc/director_client.py
```diff
@@ -191,31 +190,35 @@ def set_new_experiment(self, name, col_names, arch_path,
                            initial_tensor_dict=None):
         """Send the new experiment to director to launch."""
         logger.info('SetNewExperiment')
-        model_proto = None
         if initial_tensor_dict:
```
igor-davidyuk, line 193: What happens if the function receives no tensor dict? 
It probably never happens.

igor-davidyuk (commented):
In the end, I am not sure if it is a good idea to fill the message and send it in different methods.

openfl/transport/grpc/director_client.py
dmitryagapov: Function return `None`.
It looks like that it is not part of current task.

aleksandr-mokrov (approved):

igor-davidyuk (approved):

PR #140, opened 2021-08-04
WIP: Python current work directory no longer changes when starting and stopping aggregator and collaborator

aleksandr-mokrov:
Change working dir structure (client_id was added to path).
An opportunity to pass working directory path to aggregator and collaborator was added.
Python current work directory no longer changes when starting and stopping aggregator and collaborator form director and envoy.
Collaborator working directory creation was moved from director_client to envoy.

### DIFF
--- a/openfl/component/director/director.py
+++ b/openfl/component/director/director.py
@@ -64,20 +64,25 @@ class Director:
         """Set new experiment."""
         # TODO: save to file
         self.experiment_data[experiment_name] = data
+        experiment_work_dir = Path(f'{os.getcwd()}/{sender_name}/{experiment_name}')
 
-        self.create_workspace(experiment_name, data)
+        self.create_workspace(experiment_name, experiment_work_dir, data)
         asyncio.create_task(self._run_aggregator(
-            sender_name,
-            tensor_dict,
-            experiment_name,
-            collaborator_names
+            experiment_sender=sender_name,
+            initial_tensor_dict=tensor_dict,
+            experiment_name=experiment_name,
+            collaborator_names=collaborator_names,
+            experiment_work_dir=experiment_work_dir
         ))
 
         logger.info(f'New experiment {experiment_name} for '
                     f'collaborators {collaborator_names}')
         for col_name in collaborator_names:
             queue = self.col_exp_queues[col_name]
-            await queue.put(experiment_name)
+            await queue.put({
+                'experiment_name': experiment_name,
+                'client_id': sender_name
+            })
 
         return True
 
@@ -102,9 +107,9 @@ class Director:
     async def wait_experiment(self, collaborator_name: str) -> str:
         """Wait an experiment."""
         queue = self.col_exp_queues[collaborator_name]
-        experiment_name = await queue.get()
+        experiment_data = await queue.get()
 
-        return experiment_name
+        return experiment_data
 
     def get_dataset_info(self):
         """Get dataset info."""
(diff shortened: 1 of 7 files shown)


openfl/component/director/director.py
```diff
@@ -64,20 +64,25 @@ async def set_new_experiment(self, *, experiment_name: str, sender_name: str,
...
         ))
 
         logger.info(f'New experiment {experiment_name} for '
                     f'collaborators {collaborator_names}')
         for col_name in collaborator_names:
             queue = self.col_exp_queues[col_name]
-            await queue.put(experiment_name)
+            await queue.put({
+                'experiment_name': experiment_name,
+                'client_id': sender_name
```
igor-davidyuk, line 84: Don't we disclose the experiment owner's identity to envoys here? Is it necessary?

openfl/component/director/director.py
```diff
@@ -64,20 +64,25 @@ async def set_new_experiment(self, *, experiment_name: str, sender_name: str,
         """Set new experiment."""
         # TODO: save to file
         self.experiment_data[experiment_name] = data
+        experiment_work_dir = Path(f'{os.getcwd()}/{sender_name}/{experiment_name}')
```
igor-davidyuk, line 67: We discussed making timestamps-names for experiment working directories, is it a better idea?

openfl/component/envoy/envoy.py
```diff
@@ -46,20 +54,21 @@ def run(self):
         """Run of the envoy working cycle."""
         while True:
             try:
-                # Workspace import should not be done by gRPC client!
-                experiment_name = self.director_client.get_experiment_data()
+                experiment_name, client_id = self.director_client.wait_experiment()
```
igor-davidyuk, line 57: experiment owner's id on envvoys side, is it really needed?

openfl/component/envoy/envoy.py
```diff
@@ -46,20 +54,21 @@ def run(self):
...
+                experiment_work_dir = Path(f'{os.getcwd()}/{client_id}/{experiment_name}')
+                data_stream = self.director_client.get_experiment_data(experiment_name)
+                self.create_workspace(experiment_name, experiment_work_dir, data_stream)
```
igor-davidyuk, line 60: This is a strange signature for the `create_workspace` method. Why it accepts data_stream? Its only responsibility should be to create the workspace, not reading the stream.

openfl/component/envoy/envoy.py
```diff
@@ -73,26 +82,74 @@ def send_health_check(self):
...
+            if experiment_work_dir in sys.path:
+                sys.path.remove(str(experiment_work_dir))
+
+    @staticmethod
+    def remove_workspace(experiment_name):
+        """Remove the workspace."""
+        shutil.rmtree(experiment_name, ignore_errors=True)
+
+    @staticmethod
+    def create_workspace(experiment_name, experiment_work_dir, response_iter):
```
igor-davidyuk, line 118: Again, we have this functionality in three places: Director, Envoy, and the old CLI's `workspace import`.
We may think of creating one helper function and use it everywhere.

openfl/component/envoy/envoy.py
```diff
@@ -73,26 +82,74 @@ def send_health_check(self):
...
+                else:
+                    raise Exception('Broken archive')
+
+        shutil.unpack_archive(arch_name, experiment_work_dir)
+        os.remove(arch_name)
+
+        requirements_filename = f'./{experiment_work_dir}/requirements.txt'
+
+        if os.path.isfile(requirements_filename):
+            attempts = 3
```
igor-davidyuk, line 140: a hardcoded variable.

openfl/federated/plan/plan.py
```diff
@@ -64,7 +64,8 @@ def ignore_aliases(self, data):
 
     @staticmethod
     def parse(plan_config_path: Path, cols_config_path: Path = None,
-              data_config_path: Path = None, resolve=True):
+              data_config_path: Path = None, resolve=True,
+              working_dir: Path = None):
```
igor-davidyuk, line 68: It feels like we are polluting the Plan's code with folder management.
We removed two `os.chdir()` lines and then wrote like 150 more.

igor-davidyuk (changes requested):

openfl/component/director/director.py
dmitryagapov: Maybe we can create constant path to save experiments data, e.g. `~/.openfl/experiments/`?

PR #141, opened 2021-08-05
Save experiment archive to file instead store it in memory.

aleksandr-mokrov:


openfl/component/director/director.py
```diff
@@ -230,6 +230,7 @@ async def _run_aggregator(
         finally:
             os.chdir(cwd)
             shutil.rmtree(experiment_name)
+            os.remove(data_file_name)
```
igor-davidyuk, line 239: Do we need `data_file_name` here just to remove the archive? We could do it after line 69 `create_workspace`

aleksandr-mokrov: We use it to send to envoys. Inside `create_workspace` we create experiment workspace for aggregator, but the director has to send experiment archive to envoy and it is hardly a good idea get it form aggregator workspace. We can remove it earlier, after sending to all envoys, but it makes sense to keep it until the end of the experiment like director experiment context.

openfl/component/director/director.py
```diff
@@ -60,17 +60,19 @@ def acknowledge_shard(self, shard_info: director_pb2.ShardInfo) -> bool:
 
     async def set_new_experiment(self, *, experiment_name: str, sender_name: str,
                                  tensor_dict: dict,
-                                 collaborator_names: typing.Iterable[str], data: bytes) -> bool:
+                                 collaborator_names: typing.Iterable[str],
+                                 data_file_path: str) -> bool:
```
dmitryagapov: Will be better if all params use one style
```python
async def set_new_experiment(self, *, 
                             experiment_name: str, 
                             sender_name: str,
                             tensor_dict: dict,
                             collaborator_names: typing.Iterable[str],
                             data_file_path: str) -> bool:
```

dmitryagapov (changes requested):

openfl/component/director/director.py
aleksandr-mokrov: reformatted

dmitryagapov (approved):

PR #143, opened 2021-08-11
replace validate_caller -> get_sender

dmitryagapov:


openfl/component/director/director.py
```diff
@@ -88,13 +88,13 @@ async def set_new_experiment(
 
         return True
 
-    def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
+    def get_trained_model(self, experiment_name: str, sender: str, model_type: str):
```
igor-davidyuk, line 91: In my opinion, it is a confusing renaming. `get_trained_model` is a call, so we passed the caller's name as a parameter.
Now we pass the sender's name, and it doesn't make sense, this method will send the model back to the caller, so the sender is the Director in this case

openfl/protocols/director.proto
```diff
@@ -7,142 +7,132 @@ import "google/protobuf/duration.proto";
...
 
 message NodeInfo {
-    string name = 1;
-    string adress = 2;
-    bool cuda_available = 3;
-    uint32 memory_size = 4;
+  string name = 1;
+  string adress = 2;
+  bool cuda_available = 3;
+  uint32 memory_size = 4;
```
igor-davidyuk, line 18: Doesn't it reduce interface readability?

openfl/transport/grpc/director_server.py
```diff
@@ -54,26 +54,21 @@ def _fill_certs(self, root_ca, key, cert):
...
 
         Args:
-            request: The gRPC message request
             context: The gRPC context
 
         Returns:
-            True - if the caller has valid cert
-            False - if the callers cert name is invalid
+            If TLS is used return sender name from certificate
+            If TLS is unused return default name 'unauthorized_sender'
```
igor-davidyuk, line 66: This should be discussed at a separate architectural meeting.

igor-davidyuk (commented):
I vote for returning the `caller` back.
Also `unauthorized_sender` part is misleading. Any user that can connect to a Director is authorized to access their userspace. They may be not authenticated without certificates, though

openfl/protocols/director.proto
dmitryagapov: https://developers.google.com/protocol-buffers/docs/style?hl=ru

itrushkin:
Please merge the latest changes from develop branch in order to pass new CI tests.

dmitryagapov:
will be fixed in https://github.com/intel/openfl/pull/166

PR #148, opened 2021-08-12
Workspace context manager

aleksandr-mokrov:
Use context managers to manage experiment workspaces
Fix plan.config['task_runner'] filling.


openfl/component/director/workspace.py
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Workspace module."""
+
+import os
+import shutil
+from pathlib import Path
+
+
+class AggregatorWorkspace:
```
igor-davidyuk: I suggest moving this away from the components package, to the utilities probably

openfl/component/envoy/workspace.py
```diff
@@ -0,0 +1,79 @@
...
+import shutil
+import sys
+import time
+from subprocess import check_call
+from sys import executable
+
+logger = logging.getLogger(__name__)
+
+
+class CollaboratorWorkspace:
```
igor-davidyuk: I suggest moving this away from the components package too.

igor-davidyuk (commented):
Could we combine `CollaboratorWorkspace` and `AggregatorWorkspace`, maybe introduce an additional parameter.
At least make a base `ActorWorkspace` class with common functionality?

Also could you please remove the old version in this PR?
https://github.com/aleksandr-mokrov/openfl/blob/e04c2e00b3bd072ff23e06b4c5a36e4de768020d/openfl/utilities/utils.py#L56

openfl/component/director/workspace.py
aleksandr-mokrov: Moved to utilities/workspace.py

openfl/component/envoy/workspace.py
aleksandr-mokrov: Moved to utilities/workspace.py

aleksandr-mokrov:
> Could we combine `CollaboratorWorkspace` and `AggregatorWorkspace`, maybe introduce an additional parameter.
> At least make a base `ActorWorkspace` class with common functionality?
> 
> Also could you please remove the old version in this PR?
> https://github.com/aleksandr-mokrov/openfl/blob/e04c2e00b3bd072ff23e06b4c5a36e4de768020d/openfl/utilities/utils.py#L56
The old version was removed. `CollaboratorWorkspace` and `AggregatorWorkspace` were merged to `ExperimentWorkspace` context manager.

openfl/utilities/workspace.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class ExperimentWorkspace:
+    """Experiment workspace context manager."""
+
+    def __init__(
+            self,
+            experiment_name: str,
+            data_file_path: Path,
+            is_install_requirements: bool = False
```
igor-davidyuk, line 25: is_install_requirements -> install_requirements
probably this will look better

igor-davidyuk (approved):
Nice!

openfl/utilities/workspace.py
aleksandr-mokrov: It's boolean flag. It will be strange:
```python
if self.install_requirements:
            self._install_requirements()
```
Probably `is_install_requirements_required` would be better, but too long.

openfl/component/envoy/envoy.py
```diff
@@ -47,21 +47,36 @@ def run(self):
...
                 logger.error(f'Failed to get experiment: {exc}')
                 time.sleep(DEFAULT_RETRY_TIMEOUT_IN_SECONDS)
                 continue
+            data_file_path = self._save_data_stream_to_file(data_stream)
             self.is_experiment_running = True
             try:
-                self._run_collaborator(experiment_name)
+                with ExperimentWorkspace(
+                        experiment_name, data_file_path, is_install_requirements=True
+                ):
```
dmitryagapov, line 63: Maybe rename `is_install_requirements` => `requirements` in order to avoid so long line.

openfl/transport/grpc/director_client.py
```diff
@@ -69,67 +64,28 @@ def report_shard_info(self, shard_descriptor) -> bool:
         acknowledgement = self.stub.AcknowledgeShard(shard_info)
         return acknowledgement.accepted
 
-    def get_experiment_data(self):
-        """Get an experiment data from the director."""
+    def wait_experiment(self):
+        """Wait an experiment data from the director."""
```
dmitryagapov, line 68: Maybe `wait_experiment` should be `get_experiment_name`?

openfl/utilities/workspace.py
dmitryagapov: I also think that `requirements` will be better. You still can rename it inside `__init__`.

openfl/utilities/workspace.py
```diff
@@ -0,0 +1,77 @@
...
+        """Initialize workspace context manager."""
+        self.experiment_name = experiment_name
+        self.data_file_path = data_file_path
+        self.cwd = os.getcwd()
+        self.experiment_work_dir = f'{self.cwd}/{self.experiment_name}'
+        self.is_install_requirements = is_install_requirements
+
+    def _install_requirements(self):
+        """Install experiment requirements."""
+        requirements_filename = f'{self.experiment_work_dir}/requirements.txt'
```
dmitryagapov, line 36: Maybe better to use `pathlib.Path` here

dmitryagapov (changes requested):

openfl/utilities/workspace.py
aleksandr-mokrov: This variable is used in 3 places in this method, some of them are converted to a string, and some must be converted to a string manually. None of them use type Path.

dmitryagapov (approved):

openfl/utilities/workspace.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if os.path.isfile(requirements_filename):
+            attempts = 10
+            for _ in range(attempts):
```
itrushkin, line 40: @aleksandr-mokrov Shouldn't we raise an exception when all attempts to install the requirements failed?

PR #150, opened 2021-08-16
Create global variables for CA files/paths

dmitryagapov:


aleksandr-mokrov (approved):

openfl/interface/aggregator.py
```diff
@@ -62,7 +62,7 @@ def generate_cert_request(fqdn):
     from openfl.cryptography.participant import generate_csr
     from openfl.cryptography.io import write_crt
     from openfl.cryptography.io import write_key
-    from openfl.interface.cli_helper import PKI_DIR
+    from openfl.component.ca.ca import CA_PKI_DIR
```
igor-davidyuk: May be better to leave `PKI_DIR` instead of `CA_PKI_DIR`?
Or even make it `CERT_DIR`?

dmitryagapov: Agree

igor-davidyuk (approved):

PR #151, opened 2021-08-16
Tests for interactive API

itrushkin:


itrushkin:
Jenkins please retry a build

igor-davidyuk (approved):

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,134 @@
...
+        """Initialize KvasirShardDescriptor."""
+        super().__init__()
+
+        self.data_folder = Path.cwd() / data_folder
+        self.download_data(self.data_folder)
+
+        # Settings for resizing data
+        self.enforce_image_hw = None
+        if enforce_image_hw is not None:
+            self.enforce_image_hw = tuple(int(size) for size in enforce_image_hw.split(','))
```
dmitryagapov: `self.enforve_image_hw` is it should be `tuple`?
`enforce_image_hw` could we use two different variable with type `int` in order to avoid this transformation?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,134 @@
...
+        # Settings for sharding the dataset
+        self.rank_worldsize = tuple(int(num) for num in rank_worldsize.split(','))
```
dmitryagapov: `self.rank_worldsize` is it should be tuple?
`rank_worldsize` could we use two different variable with type int in order to avoid this transformation?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,134 @@
...
+
+        self.images_path = self.data_folder / 'segmented-images' / 'images'
+        self.masks_path = self.data_folder / 'segmented-images' / 'masks'
+
+        self.images_names = [
+            img_name
+            for img_name in sorted(os.listdir(self.images_path))
+            if len(img_name) > 3 and img_name[-3:] == 'jpg'
```
dmitryagapov: To check file extension could be used 
```python
os.path.splitext(img_name)[1] == ".jpg"
```
or
```python 
Path(img_name).suffix == '.mp3'
```

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,134 @@
...
+        ]
+        # Sharding
+        self.images_names = self.images_names[self.rank_worldsize[0] - 1::self.rank_worldsize[1]]
+
+        # Calculating data and target shapes
+        sample, target = self[0]
+        self._sample_shape = [str(dim) for dim in sample.shape]
+        self._target_shape = [str(dim) for dim in target.shape]
```
dmitryagapov, line 52: Why we should store numbers as strings?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/shard_config.yaml
```diff
@@ -0,0 +1,5 @@
+template: kvasir_shard_descriptor.KvasirShardDescriptor
+params:
+  data_folder: kvasir_data
+  rank_worldsize: 1,90
+  enforce_image_hw: '300,400'
```
dmitryagapov: could this values divided?
could its have the same type?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
```diff
@@ -4,56 +4,139 @@
...
 from openfl.interface.interactive_api.federation import Federation
-from tests.github.interactive_api.experiments.pytorch_kvasir_unet.tasks import validate, task_interface
+from tests.github.interactive_api_director.experiments.pytorch_kvasir_unet.tasks import validate, task_interface
 # from tests.github.interactive_api.experiment_runner import run_experiment
 
 from copy import deepcopy
 
 
-logger = logging.getLogger(__name__)
+federation = Federation(client_id='frontend', director_node_fqdn='localhost', director_port='50051', tls=False)
```
dmitryagapov: could `director_port` be number?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
```diff
@@ -4,56 +4,139 @@
...
+dummy_shard_desc = federation.get_dummy_shard_descriptor(size=10)
+sample, target = dummy_shard_desc[0]
+
+from openfl.interface.interactive_api.experiment import TaskInterface, DataInterface, ModelInterface, FLExperiment
+
+import os
+import PIL
+import numpy as np
+from torch.utils.data import Dataset, DataLoader, SubsetRandomSampler
+from torchvision import transforms as tsf
```
dmitryagapov: is it really make sense to keep some imports not in the file header?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
```diff
@@ -4,56 +4,139 @@
...
+        img, mask = self.shard_descriptor[index]
+        img = self.img_trans(img).numpy()
+        mask = self.mask_trans(mask).numpy()
+        return img, mask
+
+    def __len__(self):
+        return len(self.shard_descriptor)
+    
+    
+    def get_train_loader(self, **kwargs):
```
dmitryagapov: `kwargs` unused

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
```diff
@@ -4,56 +4,139 @@
...
+        """
+        Output of this method will be provided to tasks with optimizer in contract
+        """
+        train_sampler = SubsetRandomSampler(self.train_indeces)
+        return DataLoader(
+            self, num_workers=8, batch_size=self.kwargs['train_bs'], sampler=train_sampler
```
dmitryagapov: Could we define `train_bs` as param in `__init__`?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
```diff
@@ -4,56 +4,139 @@
...
+            )
+
+    def get_valid_loader(self, **kwargs):
```
dmitryagapov: `kwargs` unused

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,134 @@
...
+    print(kvasir_sd.sample_shape, kvasir_sd.target_shape)
+
+    from openfl.component.envoy.envoy import Envoy
+
+    shard_name = 'one'
+    director_uri = 'localhost:50051'
+
+    keeper = Envoy(
+        shard_name=shard_name,
+        director_uri=director_uri,
```
dmitryagapov: `director_uri` was replaced with `director_host` and `director_port`

tests/github/interactive_api_director/experiments/tensorflow_mnist/director/config.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright (C) 2020-2021 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+# Director's config.
+# Parameters:
+#   1. sample_shape - sample shape interface unified across the Federation
+#   1. target_shape - target shape interface unified across the Federation
+
+settings:
+  sample_shape: ['784']
```
dmitryagapov, line 10: could `'784'` be number `784`

tests/github/interactive_api_director/experiments/tensorflow_mnist/envoy/shard_descriptor.py
```diff
@@ -0,0 +1,92 @@
...
+        self._sample_shape = [str(dim) for dim in sample.shape]
+        self._target_shape = [str(dim) for dim in target.shape]
+
+    
+    def __getitem__(self, index):
+        """Return a item by the index."""
+        if index < len(self.X_train):
+            return self.X_train[index], self.y_train[index]
+        index -= len(self.X_train) + 1
+        return self.X_test[index], self.y_test[index]
```
dmitryagapov, line 41: It would be greate if we have the same names for similar variables `X_train` or `train_indices` and `index` or `indice`

tests/github/interactive_api_director/experiments/tensorflow_mnist/experiment.py
```diff
@@ -1,143 +1,127 @@
...
+MI = ModelInterface(model=model, optimizer=optimizer, framework_plugin=framework_adapter)
+
+
+def function_defined_in_notebook(some_parameter):
+    print(f'Also I accept a parameter and it is {some_parameter}')
+
+
+TI = TaskInterface()
+# Task interface currently supports only standalone functions.
+@TI.register_fl_task(model='model', data_loader='train_dataset', \
```
dmitryagapov: remove slash

dmitryagapov (changes requested):

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/shard_config.yaml
itrushkin: This object reflects 'size' type that by design consists of fixed width and height. For example, the same approach is used for `Resize*` transforms in [`torchvision.transforms`](https://pytorch.org/vision/stable/transforms.html), where `size` variable is passed as a tuple or single integer, which means that `h=w`. In my opinion, these values should be stored together since they are always used together.

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,134 @@
...
+        shard_descriptor=kvasir_sd,
+        tls=False,
+        root_ca='./cert/root_ca.crt',
```
aleksandr-mokrov: `root_ca`, `key` and `cert` were renamed to `root_certificate`, `private_key` and `certificate`

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/start_envoy.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+fx envoy start -n env_one --disable-tls --shard-config-path shard_config.yaml -d localhost:50051
```
aleksandr-mokrov: fx envoy start -n env_one --disable-tls --shard-config-path shard_config.yaml -dh localhost -dp 50051

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
aleksandr-mokrov: Could be.  It uses to get director address string, like `f'{director_node_fqdn}:{director_port}'`

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/shard_config.yaml
dmitryagapov: `rank_worldsize` and `enforce_image_hw` have diferent types `list` and `string`. It looks like it should be more consistent

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
itrushkin: Because `ShardDirectorClient.report_shard_info` expects strings for `sample_shape` and `target_shape`. We use string type here to allow users to specify shape ranges, e. g. if images have different heights or widths we would use `['100-200','300-400']`.

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,142 @@
...
+        """Initialize KvasirShardDescriptor."""
+        super().__init__()
+        
+        self.data_folder = Path.cwd() / data_folder
+        self.download_data(self.data_folder)
+
+        # Settings for resizing data
+        self.enforce_image_hw = None
+        if enforce_image_hw is not None:
+            self.enforce_image_hw = tuple(enforce_image_hw)
```
dmitryagapov, line 33: Could we specify `enforce_image_hw: Tuple[int] = None` to avoid this check?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
dmitryagapov: I created separate issue for it
https://github.com/intel/openfl/issues/159

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
```diff
@@ -0,0 +1,142 @@
...
+
+    keeper = Envoy(
+        shard_name=shard_name,
+        director_host=director_host,
+        director_port=director_port,
+        shard_descriptor=kvasir_sd,
+        tls=False,
+        root_certificate='./cert/root_ca.crt',
+        private_key='./cert/one.key',
+        certificate='./cert/one.crt',
```
dmitryagapov: `root_certificate`, `private_key`, `certificate` are not required then `tls=False`

dmitryagapov (changes requested):

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
itrushkin: To allow using `Tuple` here we need to parse YAML collections as tuples. But it requires a specific `!!python/tuple` tag that is not supported by `yaml.safe_load` function. All collections are parsed as lists, and they need to be converted to tuples inside `ShardDescriptor` class since the usage of `yaml.safe_load` is strictly required.

aleksandr-mokrov (approved):

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
dmitryagapov: Maybe we can convert it to Tuple before we are passing the arguments to constructor?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/experiment.py
igor-davidyuk: That is Collaborator's inner signature for `get_train_loader` method. Better not to mess with it without full understanding

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/kvasir_shard_descriptor.py
itrushkin: We probably could if we created `ShardDescriptor` instance manually only. But when we spawn envoy via `fx envoy start`, it reads Shard Descriptor YAML file, calls the constructor of `template` class, and passes `settings` as keyword arguments. In this case, the constructor will always receive collections as lists. We can not access intermediate steps between parsing the config file and calling the constructor because it happens inside the library code.

dmitryagapov (approved):

PR #157, opened 2021-08-18
Full traceback call stack on broad exception handling.

### LINKED ISSUE #142: Poor error traceback
When I have some errors in my code running federation, I get only the error without full traceback with call stack. For example: `[17:46:46] ERROR Collaborator failed: list index out of range` . And I have no way of knowing where specifically error is. I only have a message above and link to https://github.com/intel/openfl/blob/c2796b6c3a425436d38c3b3b7f8867e2ea4f9918/openfl/component/envoy/envoy.py#L59

aleksandr-mokrov:
Fixes #142


dmitryagapov (approved):

igor-davidyuk (approved):
magnificent

PR #186, opened 2021-09-20
Pin versions to avoid a long pip version resolving and provide tested environment

### LINKED ISSUE #182: Pin versions of packages
We need to ping all unpinned versions of packages in setup.py & workspace environments.

aleksandr-mokrov:
resolves #182


psfoley (changes requested):
As mentioned in the linked issue, pinning package versions in OpenFL can lead to unnecessary version conflicts when used in conjunction with other frameworks. This PR should not be merged until https://github.com/intel/openfl/issues/182 is resolved.

igor-davidyuk:
I would argue that pinning versions is not an optimal solution for this issue. 
On the contrary, we could relax requirements versioning in develop branch.

PR #188, opened 2021-09-21
Feature/experiment redesign

dmitryagapov:
Create ExperimentRegistry class to manage experiments execution order. 
Create Experiment class in order to move all logic connected to run aggregator to this class.


openfl/component/director/director.py
```diff
@@ -6,19 +6,245 @@
...
+        if self.__aggregator_grpc_server:
+            return self.__aggregator_grpc_server.aggregator
+
+
+class ExperimentsList:
+    """ExperimentsList class."""
+
+    def __init__(self, experiments: List[Experiment] = None, **kwargs) -> None:
+        """Initialize an experiments list object."""
+        super().__init__(**kwargs)
```
aleksandr-mokrov: Why is it here?

openfl/component/director/director.py
```diff
@@ -6,19 +6,245 @@
...
+        self.collaborators = collaborators
+        self.sender = sender
+        self.init_tensor_dict = init_tensor_dict
+        if isinstance(plan_path, str):
+            plan_path = Path(plan_path)
+        self.plan_path = plan_path
+        self.__aggregator_grpc_server = aggregator_grpc_server
+        self.users = set() if users is None else set(users)
+        self.status = Status.PENDING
+        self.col_exp_queues = defaultdict(asyncio.Queue)
```
aleksandr-mokrov: Looks like it doesn't use.

openfl/component/director/director.py
```diff
@@ -6,19 +6,245 @@
...
 
 ENVOY_HEALTH_CHECK_PERIOD = 60  # in seconds
 
 
+class Status:
+    """Experiment's statuses."""
+
+    PENDING = 'pending'
+    FINISHED = 'finished'
+    IN_PROGRESS = 'inProgress'
```
aleksandr-mokrov: Probably `in_progress`?

openfl/component/director/director.py
```diff
@@ -6,19 +6,245 @@
...
+        if self.__active_experiment is None:
+            return None
+        return self.__dict[self.__active_experiment]
+
+    @property
+    def queue(self) -> List[str]:
+        """Get queue of not started experiments."""
+        return self.__experiments_queue
+
+    def add(self, experiment: Experiment) -> None:
```
aleksandr-mokrov: Why not `append` if class was named `ExperimentList`?

openfl/component/director/director.py
```diff
@@ -118,13 +341,13 @@ def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
...
-    def get_experiment_data(self, experiment_name: str) -> bytes:
+    def get_experiment_data(self, experiment_name: str) -> Path:
         """Get experiment data."""
-        return self.experiment_data.get(experiment_name, b'')
+        return self.experiments[experiment_name].data_path
 
     async def wait_experiment(self, collaborator_name: str) -> str:
         """Wait an experiment."""
-        queue = self.col_exp_queues[collaborator_name]
+        queue = self.experiments.col_exp_queues[collaborator_name]
```
aleksandr-mokrov: Now experiments adding into `col_exp_queues`  incapsulates in `ExperimentList`, but getting is here. Can we operate with this queue on the same level (Director level or ExperimentList level)?

openfl/component/director/director.py
```diff
@@ -34,24 +260,29 @@ def __init__(
...
-        # {API name : {experiment name : aggregator}}
+        self.experiments = ExperimentsList()
 
         self.tls = tls
         self.root_certificate = root_certificate
         self.private_key = private_key
         self.certificate = certificate
         self.settings = settings or {}
 
+    async def start(self):
```
aleksandr-mokrov: Looks like it is not `director.start`. Director will start without this method. It's a background task, that director executes.

openfl/component/director/director.py
```diff
@@ -6,50 +6,279 @@
...
+
+            await self._run_aggregator(
+                tls=tls,
+                root_certificate=root_certificate,
+                certificate=certificate,
+                private_key=private_key,
+            )
+            self.status = Status.FINISHED
+        except Exception:
+            self.status = Status.FAILED
```
aleksandr-mokrov: It would be nice to add error log

openfl/component/director/director.py
```diff
@@ -6,50 +6,279 @@
...
+
+    def __init__(
+            self, *,
+            name: str,
+            data_path: Union[Path, str],
+            collaborators: List[str],
+            sender: str,
+            init_tensor_dict: dict,
+            plan_path: Union[Path, str] = 'plan/plan.yaml',
+            aggregator_grpc_server: AggregatorGRPCServer = None,
```
aleksandr-mokrov: It is created inside and looks like we shouldn't pass it.

openfl/component/director/director.py
```diff
@@ -6,50 +6,279 @@
...
+
+    async def _run_aggregator(
+            self, *,
+            tls: bool = False,
+            root_certificate: Union[Path, str] = None,
+            certificate: Union[Path, str] = None,
+            private_key: Union[Path, str] = None,
+    ) -> None:
+        with ExperimentWorkspace(self.name, self.data_path):
+            self.__aggregator_grpc_server = self._create_aggregator_grpc_server(
```
aleksandr-mokrov: No reason to return aggregator_grpc_server, it's internal method and it can define self.__aggregator_grpc_server by itself. Or we should pass it into  _run_aggregator_server,  because _run_aggregator_server will be failed without running self._create_aggregator_grpc_server.

openfl/transport/grpc/director_server.py
```diff
@@ -72,7 +72,8 @@ def get_caller(self, context):
 
     def start(self):
         """Launch the director GRPC server."""
-        loop = asyncio.get_event_loop()   # TODO: refactor after end of support for python3.6
+        loop = asyncio.get_event_loop()  # TODO: refactor after end of support for python3.6
+        loop.create_task(self.director.run_background_tasks())
```
aleksandr-mokrov: It really should be a task?

openfl/component/director/director.py
```diff
@@ -205,51 +432,12 @@ def get_envoys(self) -> list:
...
-                await asyncio.sleep(10)
-        except KeyboardInterrupt:
-            pass
-        finally:
-            grpc_server.stop(0)
-            # Temporary solution to free RAM used by TensorDB
-            aggregator_server.aggregator.tensor_db.clean_up(0)
+    async def _monitor_experiment_task(self):
+        while True:
+            await self.experiments_registry.set_next_experiment()
```
aleksandr-mokrov: Maybe we replace these to methods with another one? `run_next_experiment` for example.

openfl/component/director/director.py
dmitryagapov: I thought it would be useful when we start restoring active experiment after director restart. but I agree that it could be deleted for now

openfl/component/director/director.py
dmitryagapov: Done

openfl/component/director/director.py
dmitryagapov: Done

openfl/transport/grpc/director_server.py
dmitryagapov: Please give me your proposal

openfl/component/director/director.py
dmitryagapov: `run_next_experiment` looks like it should stop current experiment and start next.

openfl/component/director/director.py
dmitryagapov: Done

openfl/transport/grpc/director_server.py
aleksandr-mokrov:
```
self.director.run_background_tasks()
```
`run_background_tasks` can be static, it only creates tasks in event loop for each background_task.

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+class Status:
+    """Experiment's statuses."""
+
+    PENDING = 'PENDING'
+    FINISHED = 'FINISHED'
+    IN_PROGRESS = 'IN_PROGRESS'
+    FAILED = 'FAILED'
+
+
+class Experiment:
```
igor-davidyuk: May rename this class so it may be distinguished from the `Experiment` from the interactive API?
I propose something like `ExperimentRepresentation`

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
 from openfl.protocols import director_pb2
+from openfl.transport import AggregatorGRPCServer
 from openfl.utilities.workspace import ExperimentWorkspace
 
 logger = logging.getLogger(__name__)
 
 ENVOY_HEALTH_CHECK_PERIOD = 60  # in seconds
 
 
+class Status:
```
igor-davidyuk: Status for an experiment is the 'one of' object. Can we use Enum here?

openfl/component/director/director.py
igor-davidyuk: This class should be put in a separate module probably.
My vision is we put it, along with `EnvoyRepresentation` and probably other abstractions under `representations` folder/package. The alternative name may be `common_structures` or something in similar fashion

igor-davidyuk: Did you also consider using this object on the Envoy's side? Would it be convenient?

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+    FAILED = 'FAILED'
+
+
+class Experiment:
+    """Experiment class."""
+
+    def __init__(
+            self, *,
+            name: str,
+            data_path: Union[Path, str],
```
igor-davidyuk:
```suggestion
            experiment_archive_path: Union[Path, str],
```

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+            collaborators: List[str],
+            sender: str,
+            init_tensor_dict: dict,
+            plan_path: Union[Path, str] = 'plan/plan.yaml',
+            users: Iterable[str] = None,
```
igor-davidyuk:
```suggestion
            authorized_users: Iterable[str] = None,
```

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+        self.init_tensor_dict = init_tensor_dict
+        if isinstance(plan_path, str):
+            plan_path = Path(plan_path)
+        self.plan_path = plan_path
+        self.users = set() if users is None else set(users)
+        self.status = Status.PENDING
+        self.__aggregator_grpc_server = None
+
+    @property
+    def aggregator(self) -> Union[Aggregator, None]:
```
igor-davidyuk: We should discuss does it make sense to move this functional part somewhere else, for example to the Experiment registry, so this representation could only contain data

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+    def get(self, key: str, default=None) -> Experiment:
+        """Get experiment by name."""
+        return self.__dict.get(key, default)
+
+    def __contains__(self, key: str) -> bool:
+        """Check if experiment exists."""
+        return key in self.__dict
+
+    def _finish_active(self) -> None:
+        self.__dict[self.__active_experiment].__aggregator_grpc_server = None
```
igor-davidyuk: Should we explicitly do `del self.__dict[self.__active_experiment].__aggregator_grpc_server` here?

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+            self.aggregator.tensor_db.clean_up(0)
+
+
+class ExperimentsRegistry:
+    """ExperimentsList class."""
+
+    def __init__(
+            self,
+            tls: bool,
+            root_certificate: Union[Path, str],
```
igor-davidyuk: I would argue that we should not keep certificates here

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+            certificate: Union[Path, str],
+            private_key: Union[Path, str],
+            experiments: List[Experiment] = None,
```
igor-davidyuk: looks like an excess variable

openfl/component/director/director.py
igor-davidyuk: This will be convenient if we will persist experiments

openfl/component/director/director.py
```diff
@@ -6,50 +6,282 @@
...
+        if experiments is None:
+            self.__experiments_queue = []
+            self.__archived_experiments = []
+            self.__dict = {}
+        else:
+            self.__dict = {
+                exp.name: exp
+                for exp in experiments
+            }
+            self.__experiments_queue = list(self.__dict.keys())
```
igor-davidyuk: could we use a real queue here?

openfl/transport/grpc/director_server.py
dmitryagapov: `run_background_tasks` was moved to `director_server.py`

openfl/component/director/director.py
dmitryagapov: `set_next_experiment` and `start_active` was combined to `run_next_experiment`

openfl/component/director/director.py
dmitryagapov: It is internal class and it should not be used from interactive API.
In interactive API we have `FLExperiment` class.

openfl/component/director/director.py
dmitryagapov: For what porpose we should use enum here?

openfl/component/director/director.py
dmitryagapov: Was moved to `openfl/components/expeiment/experiment.py`

dmitryagapov: I didn't think about it. It looks like separate task

openfl/component/director/director.py
dmitryagapov: changed to `archive_path`

openfl/component/director/director.py
dmitryagapov: There are could be unauthorized_users? If no, I think we can call them `users`.

openfl/component/director/director.py
dmitryagapov: If you mean `del __aggregator_grpc_server` it sets to `None`.
If you mean `del self.__dict[self.__active_experiment]` Here we store all information about experiment in `dict[experiment_name] = Experiment()`. If we want to have access to this experiment when it was finished we shouldn't delete this data.

openfl/component/director/director.py
dmitryagapov: It would be used when we start recreate experiments from disk

openfl/component/director/director.py
dmitryagapov: We couldn't because we couldn't see all queue, we only could get element from it (once).

aleksandr-mokrov (approved):

igor-davidyuk (approved):

openfl/component/director/experiment.py
```diff
@@ -0,0 +1,199 @@
...
+                )
+                self.aggregator = aggregator_grpc_server.aggregator
+                await self._run_aggregator_grpc_server(
+                    aggregator_grpc_server=aggregator_grpc_server,
+                )
+            self.status = Status.FINISHED
+            logger.info(f'Experiment "{self.name}" was finished successfully.')
+        except Exception as e:
+            self.status = Status.FAILED
+            logger.error(f'Experiment "{self.name}" was failed with error: {e}.')
```
itrushkin, line 86: This prints only the error message. It would be much more informative if traceback is included, and [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) can handle it.

PR #192, opened 2021-09-27
CLI validation update

alexey-khorkin:
Update validation for CLI input

### DIFF
--- a/openfl/utilities/click_types.py
+++ b/openfl/utilities/click_types.py
@@ -64,5 +64,5 @@ class IpAddressParamType(click.ParamType):
         return value
 
 
-DomainType = DomainParamType()
-IpAdressType = IpAddressParamType()
+Domain = DomainParamType()
+IpAdress = IpAddressParamType()
--- a/openfl/utilities/click_types.py
+++ b/openfl/utilities/click_types.py
@@ -1,0 +1,68 @@
+# Copyright (C) 2020-2021 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+"""Click types module."""
+
+import ipaddress
+import re
+
+
+import click
+
+
+class DomainParamType(click.ParamType):
+    """Domain Type for click arguments."""
+
+    name = 'DomainType'
+
+    @staticmethod
+    def is_fqdn(hostname: str) -> bool:
+        """https://en.m.wikipedia.org/wiki/Fully_qualified_domain_name."""
+        if not 1 < len(hostname) < 253:
+            return False
+
+        # Remove trailing dot
+        if hostname[-1] == '.':
+            hostname = hostname[:-1]
+
+        #  Split hostname into list of DNS labels
+        labels = hostname.split('.')
+
+        #  Define pattern of DNS label
+        #  Can begin and end with a number or letter only
+        #  Can contain hyphens, a-z, A-Z, 0-9
+        #  1 - 63 chars allowed
+        fqdn = re.compile(r'^[a-z0-9]([a-z-0-9-]{0,61}[a-z0-9])?$', re.IGNORECASE) # noqa 
+
+        # Check that all labels match that pattern.
+        return all(fqdn.match(label) for label in labels)
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_fqdn(value):
+            self.fail(f'{value} is not a valid domain name', param, ctx)
+        return value
+
+
+class IpAddress(click.ParamType):
+    """IpAddress Type for click arguments."""
+
+    name = 'IpAddress type'
+
+    @staticmethod
+    def is_api_adress(address: str) -> bool:
+        """Validate ip-adress value."""
+        try:
+            _ = ipaddress.ip_address(address)
+            return True
+        except ValueError:
+            return False
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_api_adress(value):
+            self.fail(f'{value} is not a valid ip adress name', param, ctx)
+        return value
+
+
+DOMAIN = DomainParamType()
+IPADDRESS = IpAddress()
--- a/openfl/utilities/click_types.py
+++ b/openfl/utilities/click_types.py
@@ -31,5 +31,5 @@ class IpAddressParamType(click.ParamType):
         return value
 
 
-Domain = FqdnParamType()
-IpAdress = IpAddressParamType()
+FQDN = FqdnParamType()
+IP_ADDRESS = IpAddressParamType()
--- a/openfl/utilities/click_types.py
+++ b/openfl/utilities/click_types.py
@@ -2,43 +2,19 @@
 # SPDX-License-Identifier: Apache-2.0
 """Click types module."""
 
-import ipaddress
-import re
-
-
 import click
 
+from openfl.utilities import utils
 
-class DomainParamType(click.ParamType):
+
+class FqdnParamType(click.ParamType):
     """Domain Type for click arguments."""
 
-    name = 'DomainType'
-
-    @staticmethod
-    def is_fqdn(hostname: str) -> bool:
-        """https://en.m.wikipedia.org/wiki/Fully_qualified_domain_name."""
-        if not 1 < len(hostname) < 253:
-            return False
-
-        # Remove trailing dot
-        if hostname[-1] == '.':
-            hostname = hostname[:-1]
-
-        #  Split hostname into list of DNS labels
-        labels = hostname.split('.')
-
-        #  Define pattern of DNS label
-        #  Can begin and end with a number or letter only
-        #  Can contain hyphens, a-z, A-Z, 0-9
-        #  1 - 63 chars allowed
-        fqdn = re.compile(r'^[a-z0-9]([a-z-0-9-]{0,61}[a-z0-9])?$', re.IGNORECASE) # noqa 
-
-        # Check that all labels match that pattern.
-        return all(fqdn.match(label) for label in labels)
+    name = 'fqdn'
 
     def convert(self, value, param, ctx):
         """Validate value, if value is valid, return it."""
-        if not self.is_fqdn(value):
+        if not utils.is_fqdn(value):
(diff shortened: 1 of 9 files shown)


openfl/utilities/utils.py
```diff
@@ -3,13 +3,71 @@
...
+        return all(fqdn.match(label) for label in labels)
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_fqdn(value):
+            self.fail(f'{value!r} is not a valid domain name', param, ctx)
+        return value
+
+
+class IpAdress(click.ParamType):
```
aleksandr-mokrov: Address

openfl/utilities/utils.py
```diff
@@ -3,13 +3,71 @@
...
 import logging
 import os
+import re
 from socket import getfqdn
 
+import click
 import numpy as np
 
 
+class DomainParamType(click.ParamType):
```
aleksandr-mokrov: It would be better to put click type in a separated file, `click_types.py` for example.

openfl/utilities/utils.py
```diff
@@ -3,13 +3,71 @@
...
+        #  1 - 63 chars allowed
+        fqdn = re.compile(r'^[a-z0-9]([a-z-0-9-]{0,61}[a-z0-9])?$', re.IGNORECASE) # noqa 
+
+        # Check that all labels match that pattern.
+        return all(fqdn.match(label) for label in labels)
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_fqdn(value):
+            self.fail(f'{value!r} is not a valid domain name', param, ctx)
```
aleksandr-mokrov: !r is not f-string way, it supported only for str.format compatibility. Use `repr(value)`. But in that case value is string, what is the reason to use !r?

openfl/utilities/utils.py
```diff
@@ -3,13 +3,71 @@
...
+        try:
+            _ = ipaddress.ip_address(address)
+            return True
+        except ValueError:
+            return False
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_api_adress(value):
+            self.fail(f'{value!r} is not a valid ip adress name', param, ctx)
```
aleksandr-mokrov: The same as in `DomainParamType`

aleksandr-mokrov (changes requested):

alexey-gruzdev:
ok to test

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+            return False
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_api_adress(value):
+            self.fail(f'{value} is not a valid ip adress name', param, ctx)
+        return value
+
+
+DOMAIN = DomainParamType()
```
aleksandr-mokrov: I suggest remove it and use like:
```
from openfl.utilities import click_types

...
@option('-ip', '--ip', required=False, type=click_types.IpAddress,
```

alexey-khorkin: We have to use instance of `IpAddress` class in type parameter, instead of self class
Because of:
```
    raise AssertionError(
AssertionError: Attempted to use an uninstantiated parameter type (<class '__main__.DomainParamType'>).
```
and on official [documentation](https://click.palletsprojects.com/en/8.0.x/parameters/) the same workflow

aleksandr-mokrov (approved):

itrushkin:
Jenkins please retry a build

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+"""Click types module."""
+
+import ipaddress
+import re
+
+
+import click
+
+
+class DomainParamType(click.ParamType):
```
dmitryagapov: FqdnParamType

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+    """Domain Type for click arguments."""
+
+    name = 'DomainType'
```
dmitryagapov: fqdn

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @staticmethod
+    def is_fqdn(hostname: str) -> bool:
```
dmitryagapov: Looks like standalon function in utils.
```python 
def is_fqdn(fqdn: str) -> bool:
    ...
```

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+        """https://en.m.wikipedia.org/wiki/Fully_qualified_domain_name."""
+        if not 1 < len(hostname) < 253:
+            return False
+
+        # Remove trailing dot
+        if hostname[-1] == '.':
+            hostname = hostname[:-1]
```
dmitryagapov:
```python 
hostname.rstrip('.')
```

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+class IpAddressParamType(click.ParamType):
+    """IpAddress Type for click arguments."""
+
+    name = 'IpAddress type'
+
+    @staticmethod
+    def is_api_adress(address: str) -> bool:
+        """Validate ip-adress value."""
+        try:
+            _ = ipaddress.ip_address(address)
```
dmitryagapov: Looks like `_ =` is nor required

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+        return value
+
+
+class IpAddressParamType(click.ParamType):
+    """IpAddress Type for click arguments."""
+
+    name = 'IpAddress type'
+
+    @staticmethod
+    def is_api_adress(address: str) -> bool:
```
dmitryagapov: Looks like standalone function in utils
```python
def is_ip_address(address: str) -> bool:
    ...
```

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not self.is_api_adress(value):
+            self.fail(f'{value} is not a valid ip adress name', param, ctx)
+        return value
+
+
+Domain = DomainParamType()
+IpAdress = IpAddressParamType()
```
dmitryagapov: Variables should be capitalized. Like it used in `click`
`click.STRING`
https://click.palletsprojects.com/en/8.0.x/api/#click.STRING

alexey-khorkin: @dmitryagapov , yes, but upper case uses only for standard python types
As I understood, for more specific types they use C++ google code style ([Pascal_case](https://en.wiktionary.org/wiki/Pascal_case) for types naming) https://google.github.io/styleguide/cppguide.html#Type_Names
For example: `FloatRange`, `click.DateTime` and `click.Path`
Actually it was upper case before 2 last commits - https://github.com/intel/openfl/pull/192/commits/23973e1961a9f5c549d99caa4da7dde932c195bd , 
but @aleksandr-mokrov didn't prefer upper case here =)

dmitryagapov: There is no any contradictions. `FloatRange` and `IntRange` are classes that's why we are using PascalCase. `click.INT` and `click.FLOAT` are class instances that's why we are using capital letters. Here you create class instances so you should use capital letters. In case if you import `DomainParamType` and use it like `type=DomainParamType()` it will be also correct. But not mixing it.

openfl/utilities/click_types.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def convert(self, value, param, ctx):
+        """Validate value, if value is valid, return it."""
+        if not utils.is_api_adress(value):
+            self.fail(f'{value} is not a valid ip adress name', param, ctx)
+        return value
+
+
+Domain = FqdnParamType()
+IpAdress = IpAddressParamType()
```
dmitryagapov: adDress

openfl/utilities/click_types.py
alexey-khorkin: @aleksandr-mokrov , I agree with Dmitry, because it looks like global constant instances, can we use upper case?

openfl/utilities/utils.py
```diff
@@ -26,6 +28,36 @@ def getfqdn_env(name: str = '') -> str:
...
+    hostname.rstrip('.')
+
+    #  Split hostname into list of DNS labels
+    labels = hostname.split('.')
+
+    #  Define pattern of DNS label
+    #  Can begin and end with a number or letter only
+    #  Can contain hyphens, a-z, A-Z, 0-9
+    #  1 - 63 chars allowed
+    fqdn = re.compile(r'^[a-z0-9]([a-z-0-9-]{0,61}[a-z0-9])?$', re.IGNORECASE) # noqa
```
dmitryagapov: Could you please specify error code in NOQA?

openfl/utilities/utils.py
```diff
@@ -26,6 +28,36 @@ def getfqdn_env(name: str = '') -> str:
...
+    #  Can contain hyphens, a-z, A-Z, 0-9
+    #  1 - 63 chars allowed
+    fqdn = re.compile(r'^[a-z0-9]([a-z-0-9-]{0,61}[a-z0-9])?$', re.IGNORECASE) # noqa 
+
+    # Check that all labels match that pattern.
+    return all(fqdn.match(label) for label in labels)
+
+
+def is_api_adress(address: str) -> bool:
+    """Validate ip-adress value."""
```
dmitryagapov: ip address

openfl/utilities/click_types.py
aleksandr-mokrov: Yes, I agree with this approach too.

openfl/utilities/utils.py
alexey-khorkin: @dmitryagapov https://github.com/MichaelKim0407/flake8-use-fstring/issues/7

dmitryagapov (approved):

PR #205, opened 2021-10-08
Remove sh tutorial scripts

alexey-khorkin:
This pull request is to remove all `.sh` files and replace them with readme files with detailed instructions on how to start the experiment with or without TLS.


openfl-tutorials/interactive_api/PyTorch_Kvasir_UNet/README.md
```diff
@@ -0,0 +1,32 @@
...
+fx director start --disable-tls -c director_config.yaml
+```
+
+2. Run envoy:
+```sh
+cd envoy_folder
+fx envoy start -n env_one --disable-tls --shard-config-path shard_config_one.yaml -dh localhost -dp 50051
+```
+
+Optional: start second envoy:
```
dmitryagapov: This string looks strange

openfl-tutorials/interactive_api/PyTorch_Market_Re-ID/README.md
```diff
@@ -0,0 +1,34 @@
+# Person re-identification Tutorial (Market-1501)
+
+<img src="https://production-media.paperswithcode.com/datasets/Market-1501-0000000097-a728ab2d_gyNBlrI.jpg" alt="alt text" width="400" height="whatever">
```
dmitryagapov: `alt="alt text"` and `height="whatever"` could be deleted

openfl-tutorials/interactive_api/PyTorch_TinyImageNet/README.md
```diff
@@ -0,0 +1,34 @@
+# TinyImagenet Classification Tutorial
+
+<img src="https://production-media.paperswithcode.com/datasets/Tiny_ImageNet-0000001404-a53923c3_XCrVSGm.jpg" alt="alt text" width="500" height="whatever">
```
dmitryagapov: `alt="alt text"` and `height="whatever"` could be deleted

openfl-tutorials/interactive_api/Tensorflow_MNIST/README.md
```diff
@@ -1,10 +1,11 @@
 # MNIST Classification Tutorial
 
-![mnist digits](http://i.ytimg.com/vi/0QI3xgXuB-Q/hqdefault.jpg "MNIST Digits")
+<img src="http://i.ytimg.com/vi/0QI3xgXuB-Q/hqdefault.jpg" alt="alt text" width="whatever" height="whatever">
```
dmitryagapov: `alt="alt text"`, `width="whatever"` and `height="whatever"` could be deleted

dmitryagapov (changes requested):

dmitryagapov:
It would be great if we stopped using `jupyter notebook` and start using `jupyter lab` everywhere.

dmitryagapov:
Should we add to `README.md` instruction for `tls` case?

alexey-gruzdev:
> It would be great if we stopped using `jupyter notebook` and start using `jupyter lab` everywhere.

In that case, I believe we need to correct https://github.com/intel/openfl/blob/develop/setup.py#L66 to install lab with installation of OpenFL.

openfl-tutorials/interactive_api/PyTorch_Kvasir_UNet/README.md
alexey-gruzdev: Agree, let's add second  '.yaml' config and use second envoy as well. Not optional step

openfl-tutorials/interactive_api/PyTorch_Kvasir_UNet/README.md
```diff
@@ -0,0 +1,32 @@
+# HyperKvasir Tutorial
+
+
+### 1. About dataset
+The data is collected during real gastro- and colonoscopy examinations at a Hospital in Norway and partly labeled by experienced gastrointestinal endoscopists. For more information, please visit [this](https://datasets.simula.no/hyper-kvasir/) web site.
+
+
+### 3. How to run this tutorial (without TLC and locally as a simulation):
```
alexey-gruzdev: Why we have missing step 2.?

alexey-gruzdev:
> Should we add to `README.md` instruction for `tls` case?

I would agree here. I believe it would be great to have a `tls` case, with steps for certificate generation too.

openfl-tutorials/interactive_api/PyTorch_Market_Re-ID/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+
+### 1. About dataset
+Market-1501 is a large-scale public benchmark dataset for person re-identification. It contains 1501 identities which are captured by six different cameras, and 32,668 pedestrian image bounding-boxes obtained using the Deformable Part Models pedestrian detector. Each person has 3.6 images on average at each viewpoint. For more information, please visit [this](https://paperswithcode.com/dataset/market-1501) web site.
+
+
+### 2. How to run this tutorial (without TLC and locally as a simulation):
```
alexey-gruzdev: I would suggest to reformulate this, because we could simply replace `localhost` with real machine name and run in distributed manner, so it's not a big deal between local & distributed running. With or Without TLS - differs because of cert generation step.

alexey-khorkin:
@alexey-gruzdev , @dmitryagapov,  Should I create another one README file inside `interactive_api` folder with shared instruction about how to run one of examples?
Because now we have the same instruction inside all readme files (i mean - "How to run this tutorial" step)

katerina-merkulova:
@alexey-khorkin, Could you add `pip install -r requirements.txt` before starting each envoy?

alexey-gruzdev:
It was decided to not go with strategy to remove all `.sh` scripts and move completely to `fx` commands for user convenience (at least for right now).

PR #208, opened 2021-10-12
Fix aggregation functions serialization

itrushkin:
# Problem
When deserializing an aggregation function on the director side in another Python environment, serialization error occurs. Since aggregation functions are included in the `TaskInterface` pickle file, deserializing it requires having all frameworks used not only by aggregation functions but by all tasks registered for the federation.
# Solution
This PR adds a separate pickle file for the aggregation function interface, so the aggregator is not required to have frameworks used in tasks functions definitions.
**NOTE**: But there is still an error possible if aggregation functions use third-party libraries not supplied with OpenFL.


openfl/component/aggregation_functions/interface.py
```diff
@@ -13,7 +13,7 @@
 from openfl.utilities import SingletonABCMeta
 
 
-class AggregationFunctionInterface(metaclass=SingletonABCMeta):
+class AggregationFunction(metaclass=SingletonABCMeta):
```
igor-davidyuk, line 16: Why do we need a singleton here?

itrushkin: The intention is to prevent creating multiple instances of default `WeightedAverage` class per task.
Related commit: https://github.com/intel/openfl/pull/79/commits/58bfe34ce9893e50c6f16d4de4a61b3879185b29
Possible concern for it is that it makes all tasks with the same type of aggregation use the same class instance. If `AggregationFunction` has its own state, then this state will be shared across tasks.

openfl/interface/interactive_api/experiment.py
```diff
@@ -361,8 +366,8 @@ def __init__(self) -> None:
         self.task_contract = {}
         # Mapping 'task name' -> arguments
         self.task_settings = defaultdict(dict)
-        # Mapping task name -> callable
-        self.aggregation_functions = {}
+        # Mapping 'task name' -> callable
+        self.aggregation_functions = defaultdict(WeightedAverage)
```
igor-davidyuk, line 370: The default value for the Aggregation function is the Weighted average. May we avoid serializing it? Just leave the slot empty and judge on the Aggregator side that no aggregation function = default aggregation function = the Weighted average

igor-davidyuk (approved):

openfl/component/aggregation_functions/interface.py
igor-davidyuk: Sounds reasonable. We should add this consideration to the corresponding docs section

openfl/interface/interactive_api/experiment.py
itrushkin: It works just like you described. We serialize `defaultdict`, and it contains only custom aggregation functions that were defined and set manually from the API layer. If a task doesn't have an entry there, then the aggregator will try to get the value from the deserialized dictionary and since it is not there, will create `WeightedAverage` instance.

alexey-gruzdev:
ok to test

alexey-gruzdev (approved):

aleksandr-mokrov (approved):

PR #212, opened 2021-10-14
Exclude path traversal from cli for internal paths

aleksandr-mokrov:
Checking that the internal paths obtained through the cli are not outside the openfl directory.
It does not apply to paths with certificate or external configuration.

### DIFF
--- a/openfl/interface/aggregator.py
+++ b/openfl/interface/aggregator.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Aggregator module."""
 
+import sys
 from logging import getLogger
 
 from click import echo
@@ -12,6 +13,7 @@ from click import Path as ClickPath
 from click import style
 
 from openfl.utilities import click_types
+from openfl.utilities.path_check import is_directory_traversal
 from openfl.utilities.utils import getfqdn_env
 
 logger = getLogger(__name__)
@@ -39,6 +41,13 @@ def start_(plan, authorized_cols, secure):
     from pathlib import Path
 
     from openfl.federated import Plan
+
+    if is_directory_traversal(plan):
+        echo('Federated learning plan path is out of the openfl workspace scope.')
+        sys.exit(1)
+    if is_directory_traversal(authorized_cols):
+        echo('Authorized collaborator list file path is out of the openfl workspace scope.')
+        sys.exit(1)
 
     plan = Plan.parse(plan_config_path=Path(plan).absolute(),
                       cols_config_path=Path(authorized_cols).absolute())
--- a/openfl/utilities/path_check.py
+++ b/openfl/utilities/path_check.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import Union
 
 
-def is_directory_traversal(directory: Union[str, Path]):
+def is_directory_traversal(directory: Union[str, Path]) -> bool:
     """Check if the directory is traversal."""
     cwd = os.path.abspath(os.getcwd())
     requested_path = os.path.relpath(directory, start=cwd)
--- a/openfl/utilities/path_check.py
+++ b/openfl/utilities/path_check.py
@@ -9,7 +9,7 @@ from typing import Union
 
 
 def is_directory_traversal(directory: Union[str, Path]) -> bool:
-    """Check if the directory is traversal."""
+    """Check for directory traversal."""
     cwd = os.path.abspath(os.getcwd())
     requested_path = os.path.relpath(directory, start=cwd)
     requested_path = os.path.abspath(requested_path)
--- a/openfl/utilities/path_check.py
+++ b/openfl/utilities/path_check.py
@@ -1,0 +1,17 @@
+# Copyright (C) 2020-2021 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""openfl path checks."""
+
+import os
+from pathlib import Path
+from typing import Union
+
+
+def is_directory_traversal(directory: Union[str, Path]):
+    """Check if the directory is traversal."""
+    cwd = os.path.abspath(os.getcwd())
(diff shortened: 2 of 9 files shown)


openfl/interface/aggregator.py
```diff
@@ -40,6 +42,13 @@ def start_(plan, authorized_cols, secure):
 
     from openfl.federated import Plan
 
+    if is_directory_traversal(plan):
+        echo('Federated learning plan path is out of the openfl workspace scope.')
+        sys.exit(1)
```
itrushkin, line 47: Can we raise an exception instead of calling `sys.exit(1)`?

openfl/utilities/path_check.py
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""openfl path checks."""
+
+import os
+from pathlib import Path
+from typing import Union
+
+
+def is_directory_traversal(directory: Union[str, Path]) -> bool:
```
itrushkin, line 11: We can move exception throwing here and call it something like `check_directory_traversal`.

openfl/utilities/path_check.py
```diff
@@ -0,0 +1,17 @@
...
+    """Check if the directory is traversal."""
```
itrushkin:
```suggestion
    """Check for directory traversal."""

File: openfl/interface/aggregator.py (hunk 1 above)
Comment (aleksandr-mokrov) [reply openfl/interface/aggregator.py:47]: Exceptions are expected to be handled or to help get additional information about bug to fix it. But in that case we have already provided information about reason of error and want to exit with error code, that could be handled by system. It would be great if these were different error codes and could be automatically handled, but it looks like another topic for discussion.

File: openfl/utilities/path_check.py (hunk 3 above)
Comment (aleksandr-mokrov) [reply openfl/utilities/path_check.py]: fixed

File: tests/openfl/utilities/path_check.py [hunk 4]
```diff
@@ -0,0 +1,25 @@
...
+import os
+from pathlib import Path
+
+import pytest
+
+from openfl.utilities.path_check import is_directory_traversal
+
+
+@pytest.mark.parametrize(
+    'directory,expected_result', [
```
alexey-khorkin, line 14:
```suggestion
    'directory, expected_result', [
```

openfl/utilities/path_check.py
aleksandr-mokrov: If we raise exception we have to handle it above. And as for me
```python
if is_directory_traversal(some_path):
    print('error')
    do_something()
```
is preferable than
```python
try:
    check_directory_traversal(some_path)
except DirectoryTraversalException as exc:
    print('error')
    do_something()
```
in that case, because we use that checking only in cli.

itrushkin: We do not need to handle this exception anyhow. All we need to do is to pass text message to it and raise it. Shouldn't the execution be interrupted with non-zero exit code and printed exception message?

tests/openfl/utilities/path_check.py
aleksandr-mokrov: In other test we don't use whitespaces too, and they are not in the pytest documentation: https://docs.pytest.org/en/6.2.x/parametrize.html. I checked, now it works with whitespaces on python3.8, but I don't sure about earliest version.

openfl/utilities/path_check.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import Union
+
+
+def is_directory_traversal(directory: Union[str, Path]) -> bool:
+    """Check for directory traversal."""
+    cwd = os.path.abspath(os.getcwd())
+    requested_path = os.path.relpath(directory, start=cwd)
+    requested_path = os.path.abspath(requested_path)
+    common_prefix = os.path.commonprefix([requested_path, cwd])
+    return common_prefix != cwd
```
alexey-khorkin, line 17:
```suggestion
    if common_prefix != cwd:
        raise RuntimeError(f"{requested_path} is out of the openfl workspace scope.")
```
Just to keep some information about what happened, if we rename it to `check_directory_traversal(some_path)`

tests/openfl/utilities/path_check.py
alexey-khorkin: @aleksandr-mokrov , i checked python 3.6.9, it works, but let's use it the same way as in docs)

tests/openfl/utilities/path_check.py
```diff
@@ -0,0 +1,25 @@
...
+        ('first_level', False),
+        ('first_level/second_level', False),
+        (os.getcwd(), False),
+        (Path(os.getcwd(), 'first_level'), False),
+        (Path(os.getcwd(), 'first_level/second_level'), False),
```
alexey-khorkin, line 19:
```suggestion
        (Path(os.getcwd(), 'first_level/second_level'), False),
        ('first_level/second_level/..', False),
        ('first_level/../first_level', False),
        ('..', True),
        ('../../file', True),
        ('/home/naive_hacker', True),
        ('first_level/second_level/../../..', True),
```
Just to add more corner cases

openfl/utilities/path_check.py
aleksandr-mokrov: It's a user interface. It's a not the best approach to throw exceptions into a user interface. In addition the operating system doesn't understand exceptions. There are no reasons to raise exceptions in CLI.

openfl/utilities/path_check.py
aleksandr-mokrov: And then we get a generic exception into the user interface, if we don't catch it. It's expected error. It should be a human readable error description for humans, and error code for an operation system. Python exceptions are for python interpreters and developers.

dmitryagapov (approved):

alexey-khorkin (approved):

itrushkin (approved):

PR #214, opened 2021-10-20
Build docker test errors

aleksandr-mokrov:
Added filtering of packages without version when exporting workspace (local packages are excluded).
Added docker build logging in case of a build error

### DIFF
--- a/openfl/interface/workspace.py
+++ b/openfl/interface/workspace.py
@@ -13,8 +13,8 @@ from click import option
 from click import pass_context
 from click import Path as ClickPath
 
-from openfl.utilities.utils import is_package_has_version
 from openfl.utilities.path_check import is_directory_traversal
+from openfl.utilities.utils import is_package_has_version
 
 
 @group()
--- a/openfl/interface/workspace.py
+++ b/openfl/interface/workspace.py
@@ -14,7 +14,7 @@ from click import pass_context
 from click import Path as ClickPath
 
 from openfl.utilities.path_check import is_directory_traversal
-from openfl.utilities.utils import is_package_has_version
+from openfl.utilities.utils import is_package_versioned
 
 
 @group()
@@ -135,7 +135,7 @@ def export_():
     requirements_generator = freeze.freeze()
     with open('./requirements.txt', 'w') as f:
         for package in requirements_generator:
-            if is_package_has_version(package):
+            if is_package_versioned(package):
                 f.write(package + '\n')
 
     archive_type = 'zip'
@@ -170,7 +170,7 @@ def export_():
     # Create Zip archive of directory
     make_archive(archive_name, archive_type, tmp_dir)
 
-    echo(f'!Workspace exported to archive: {archive_file_name}')
+    echo(f'Workspace exported to archive: {archive_file_name}')
 
 
 @workspace.command(name='import')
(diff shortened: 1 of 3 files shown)


openfl/interface/workspace.py
```diff
@@ -397,10 +396,16 @@ def dockerize_(context, base_image, save):
...
-
-    except Exception as e:
+    except docker.errors.BuildError as e:
+        for log in e.build_log:
+            msg = log.get('stream')
+            if msg:
+                echo(msg)
         echo('Failed to build the image\n' + str(e) + '\n')
         sys.exit(1)
+    except Exception as e:
```
igor-davidyuk: could you please remove this last exception so we see the error traceback (or also print the traceback from here)

igor-davidyuk (approved):

openfl/interface/workspace.py
```diff
@@ -397,10 +396,16 @@ def dockerize_(context, base_image, save):
...
+        echo('Failed to build the image with unexpected error\n' + str(e) + '\n')
+        sys.exit(1)
     else:
         echo('The workspace image has been built successfully!')
```
dmitryagapov: Maybe we should delete this `else` block and put it on the end of `try` block?

openfl/interface/workspace.py
```diff
@@ -171,7 +170,7 @@ def export_():
     # Create Zip archive of directory
     make_archive(archive_name, archive_type, tmp_dir)
 
-    echo(f'Workspace exported to archive: {archive_file_name}')
+    echo(f'!Workspace exported to archive: {archive_file_name}')
```
itrushkin: This change seems occasional.

openfl/utilities/utils.py
```diff
@@ -196,3 +196,11 @@ def validate_file_hash(file_path, expected_hash, chunk_size=8192):
 
     if h.hexdigest() != expected_hash:
         raise SystemError('ZIP File hash doesn\'t match expected file hash.')
+
+
+def is_package_has_version(package: str) -> bool:
```
itrushkin: Function name has a grammatical mistake: in an interrogative sentence, the verb "have" must be used with the auxiliary verbs "do" or "does", not "is". 
Some correct versions of naming:
 - `is_package_versioned` - this one I like the most
 - `package_has_version`
 - `does_package_have_version`

itrushkin (changes requested):

openfl/utilities/utils.py
dmitryagapov: @itrushkin, I'm not argee with you. Generally all functions that checks smth and used in `if` statement started with `is_`.

openfl/interface/workspace.py
aleksandr-mokrov: removed

openfl/interface/workspace.py
aleksandr-mokrov: I put it after

openfl/interface/workspace.py
aleksandr-mokrov: fixed

openfl/utilities/utils.py
aleksandr-mokrov: renamed to is_package_versioned

itrushkin (approved):

dmitryagapov (approved):

alexey-gruzdev:
ok to test

PR #237, opened 2021-11-11
Fix Market downloading

katerina-merkulova:
Change download link to fixed Google Drive


alexey-gruzdev:
I would only propose to remove `Using low-level Python API`, all other things seem to be good! Tested with distributed setup, so LGTM.

alexey-gruzdev (approved):

igor-davidyuk (approved):

alexey-khorkin (approved):

PR #238, opened 2021-11-11
Remove Python 2 compatible object inheritance

alexey-gruzdev:
Replace structures like `class ABC(object):` to simple `class ABC:` since they seem to be useless because we support only Python 3.x (no Python2 compatibility)


alexey-khorkin (approved):

igor-davidyuk (approved):
nicely done 😁

PR #242, opened 2021-11-17
Update tutorials configs

katerina-merkulova:
Update configs and scripts with the current API


openfl-tutorials/interactive_api/PyTorch_Kvasir_UNet/director/director_config.yaml
```diff
@@ -1,5 +1,5 @@
 settings:
-  listen_host: localhost
+  listen_ip: localhost
```
alexey-gruzdev, line 2: Are we sure that `listen_host` was changed to `listen_ip`?

katerina-merkulova: yes, otherwise, in federated mode I got `RuntimeError: Failed to bind to address <address>:50050`

PR #252, opened 2021-12-01
Basic collaborator failure handling

aleksandr-mokrov:
The goal of this PR is to avoid federation getting stuck due to failures on the collaborator's side. The following behaviour has been implemented:
1. Envoy catches an error and call the Director new RPC `SetExperimentFailed` containing next information: collaborator name, experiment name, error id and error description with traceback call stack.
2. Director logs information about this error and call aggregator's new method `stop`, that exclude that collaborator from experiment and send signals to other collaborators to shutdown.
3. Experiment stopped and director starts new experiment from queue.


openfl/protocols/director.proto
```diff
@@ -185,13 +185,25 @@ message GetEnvoysResponse {
     repeated EnvoyInfo envoy_infos = 1;
 }
 
+message SetExperimentFailedRequest {
+    string experiment_name = 1;
+    string collaborator_name = 2;
+    uint32 error_code = 3;
```
dmitryagapov, line 191: Do we really need error code for now? It looks like it always `1`

openfl/component/director/director.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         self.settings = settings or {}
         self.col_exp_queues = defaultdict(asyncio.Queue)
         self.col_exp = {}
+        self.run_aggregator_future = None
```
dmitryagapov: We can delete this

openfl/component/director/director.py
```diff
@@ -264,7 +271,7 @@ async def start_experiment_execution_loop(self):
         while True:
             async with self.experiments_registry.get_next_experiment() as experiment:
                 loop = asyncio.get_event_loop()
-                run_aggregator_future = loop.create_task(experiment.start(
+                self.run_aggregator_future = loop.create_task(experiment.start(
```
dmitryagapov: `run_aggregator_future` - used only in this function

openfl/component/aggregator/aggregator.py
```diff
@@ -881,6 +881,14 @@ def _log_big_warning(self):
             f' WARNED!!!'
         )
 
+    def stop(self, failed_collaborator: str = None) -> None:
+        """Stop aggregator execution."""
+        self.logger.info('Force stopping the aggregator execution.')
+        self.authorized_cols.remove(failed_collaborator)
+        for collaborator_name in self.authorized_cols:
```
dmitryagapov: Maybe we shouldn't delete failed collaborator from `authorized_cols`?
alternative:
```python
for collaborator_name in filter(lambda c: c != failed_collaborator, self.authorized_cols):
    ...
```

openfl/protocols/director_pb2_grpc.py
```diff
@@ -5,7 +5,7 @@
 from . import director_pb2 as director__pb2
 
 
-class FederationDirectorStub:
+class FederationDirectorStub(object):
```
igor-davidyuk, line 8: should we continue python2 support? 😁

openfl/transport/grpc/director_client.py
```diff
@@ -95,6 +95,21 @@ def get_experiment_data(self, experiment_name):
 
         return data_stream
 
+    def set_experiment_failed(
```
igor-davidyuk, line 98: should user call this RPC?

openfl/component/aggregator/aggregator.py
aleksandr-mokrov: Make sense. Done.

openfl/component/director/director.py
aleksandr-mokrov: Deleted

openfl/component/director/director.py
aleksandr-mokrov: Changed

openfl/protocols/director.proto
aleksandr-mokrov: yes, you are right, it always 1, but we already have one error type and its description, and code to handle them (write specific log message at least). And view, that there are another error type, and we should handle them in the future. It's the groundwork for the future.

openfl/protocols/director_pb2_grpc.py
aleksandr-mokrov: It's an autogenerated code from proto files.

openfl/transport/grpc/director_client.py
aleksandr-mokrov: It's the internal director client used by the Envoy. User shouldn't use this RPC. Probably stop experiment, pause, delete, but not set failed.

alexey-gruzdev:
Jenkins please retry a build

openfl/transport/grpc/director_client.py
igor-davidyuk: is user notified somehow that experiment failed?

aleksandr-mokrov: No, we haven't a notification mechanism yet. But user can indirectly understand it by stopping StreamMetric and we write the corresponding error message to the director log.
The purpose of this PR is to avoid federation freezes after collaborator failures, not to implement a complete monitor system.

igor-davidyuk (approved):

dmitryagapov (approved):

PR #254, opened 2021-12-03
Fix 'gRPC is running on insecure channel with TLS' warnings spam

ViktoriiaRomanova:
I've noticed that we have multiple warning messages  "gRPC is running on insecure channel with TLS". I suggest to fix it.

### DIFF
--- a/openfl/transport/grpc/client.py
+++ b/openfl/transport/grpc/client.py
@@ -164,7 +164,6 @@ class CollaboratorGRPCClient:
             An insecure gRPC channel object
 
         """
-
         return grpc.insecure_channel(uri, options=self.channel_options)
 
     def create_tls_channel(self, uri, root_certificate, disable_client_auth,
--- a/openfl/transport/grpc/client.py
+++ b/openfl/transport/grpc/client.py
@@ -110,7 +110,6 @@ class CollaboratorGRPCClient:
         self.root_certificate = root_certificate
         self.certificate = certificate
         self.private_key = private_key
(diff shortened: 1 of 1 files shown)


openfl/transport/grpc/client.py
```diff
@@ -162,8 +163,10 @@ def create_insecure_channel(self, uri):
             An insecure gRPC channel object
 
         """
-        self.logger.warn(
-            'gRPC is running on insecure channel with TLS disabled.')
+        if not self.notified_about_insecure:
```
alexey-khorkin: I think we can do it without `self.notified_about_insecure` variable,  we can replace 
```Python3
self.logger.warn(
                'gRPC is running on insecure channel with TLS disabled.')
            self.notified_about_insecure = True
```
code inside initialization, before the following line https://github.com/intel/openfl/blob/fd919f449119224290f1237a5700b156b565ba42/openfl/transport/grpc/client.py#L123
Correct me if I am wrong =)

dmitryagapov:
Maybe simpler use `debug` log level?

openfl/transport/grpc/client.py
aleksandr-mokrov: I agree, multiple warnings are spammed by reconnect.

aleksandr-mokrov (approved):

alexey-khorkin (approved):

ViktoriiaRomanova:
Jenkins please retry a build

dmitryagapov (approved):

PR #279, opened 2021-12-15
Fixes delta updates = True

### LINKED ISSUE #269: An error occurs when 'collaborator.settings.delta_updates' set to 'True'
**Describe the bug**
Hi OpenFL team,
I try to train my model based on the 'default' workspace. I override the value of 'collaborator.settings.delta_updates' to 'True' in the plan.  An error occurs at the beginning of the second round: "AttributeError: 'tuple' object has no attribute 'remove'".

I've also tried to change this value in "Federated_Keras_MNIST_Tutorial" and got the same error. But everything is OK when I change this value back to 'False'.

Could you help me out to resolve the issue?

**Screenshots**
bug

**Desktop (please
 complete the following information):**
 - OS: ubuntu 20.04.2.0
 - Browser Firefox

psfoley:
Resolves https://github.com/intel/openfl/issues/269

- Small bug fix to removing the 'compressed' or 'lossy_compressed' tag from the TensorKey
- Changes the default rounds to save in the TensorDB to 2 for the aggregator, because the delta is generated between the prior and current round.


openfl/component/aggregator/aggregator.py
```diff
@@ -341,9 +341,9 @@ def get_aggregated_tensor(self, collaborator_name, tensor_name,
         #  The returned tensor will
         # be recompressed anyway.
         if 'compressed' in tags:
-            tags.remove('compressed')
+            list(tags).remove('compressed')
```
aleksandr-mokrov: `list(tags)` returns new list and `'compressed'` will be removed from that list, not form original `tags`.

psfoley: Thanks @aleksandr-mokrov. Fixed.

openfl/component/aggregator/aggregator.py
```diff
@@ -341,8 +341,10 @@ def get_aggregated_tensor(self, collaborator_name, tensor_name,
         #  The returned tensor will
         # be recompressed anyway.
         if 'compressed' in tags:
+            tags = list(tags)
```
dmitryagapov: It looks like it should be always `list` according to openfl/protocols/federation.proto:41.
If convert is really needed maybe we can do it in the start of the function.

psfoley: The cast is necessary (an error is thrown about modifying a tuple otherwise) - but I've moved this out of the if statements

dmitryagapov (approved):

aleksandr-mokrov (approved):

alexey-gruzdev (approved):

PR #288, opened 2022-01-11
Default random seed for non-deterministic DataSplitters

### LINKED ISSUE #282: Same data shard on multiple envoys when using shard descriptor with random data splitter
**Describe the bug**
When using shard descriptor with data splitter functionality (kvasir_shard_descriptor_with_data_splitter.py), the same data shard is appearing across multiple envoys. This is because the split function to shard the dataset (RandomNumpyDataSplitter) is being called on every envoy, causing the split to occur on every envoy. https://github.com/intel/openfl/blob/b60cbfbdad595e653c94cee23fd35add993b94b0/openfl-tutorials/interactive_api/PyTorch_Kvasir_UNet/envoy/kvasir_shard_descriptor_with_data_splitter.py#L38 This leads to duplicate shards on envoys. Ideally, the split function needs to be called once and then sharded dataset should be distributed according to the rank of the envoys.

**To Reproduce**
Run openfl-tutorials/interactive_api/PyTorch_Kvasir_Unet tutorial with 2 envoys having rank,worldsize of 1,3 and 2,3 and print the shard index for each of the envoys.

**Expected behavior**
The same data shard index should not be given to multiple envoys.

**Screenshots**
shard indices on envoy 1:
image

shard indices on envoy 2:
image

Index numbers like 912, 638, 651, 653 etc are present in both images.

**Desktop (please complete the following information):**
 - OS: Ubuntu 16.04.6 LTS
 - Browser chrome

itrushkin:
Fixes the random behavior of data splitters that use non-deterministic logic.

If shard descriptors of Envoys use the same data splitter, they should share the same parts of data.
By default, NumPy random seed of all splitters will be equal to zero. It means if Envoys use the same shard descriptor code (where data splitter is created) then data split is identical.

Closes #282.


itrushkin:
Jenkins please retry a build

aleksandr-mokrov (approved):

alexey-khorkin (approved):

ViktoriiaRomanova:
Shouldn't we add in `class EqualNumPyDataSplitter` 'seed' too? To make the same shuffle on each envoy

openfl/utilities/data_splitters/numpy.py
```diff
@@ -31,16 +31,21 @@ def split(self, data: np.ndarray, num_collaborators: int) -> List[List[int]]:
...
+            seed(int): Random numbers generator seed.
+                For different splits on envoys, try setting different values for this parameter
+                on each shard descriptor.
         """
         self.shuffle = shuffle
+        self.seed = seed
 
     def split(self, data, num_collaborators):
         """Split the data."""
+        np.random.seed(0)
```
ViktoriiaRomanova: Looks like a typo here. Should be `np.random.seed(self.seed)`

igor-davidyuk:
Will [this line](https://github.com/intel/openfl/blob/5e5ff192612d4e66658c75fc6bf5e69eeaef7736/openfl/utilities/data_splitters/numpy.py#L52) split classes unevenly in case of no shuffle? 
if the original dataset is ordered by target, like [0,0,0,1,1,1,1], the first collaborator will get only 0th class.
If it is true, may we use `idx[rank : len(idx) : cols]` instead of `np.split_array`?

igor-davidyuk:
May we consider including seed in the interface class' signature instead of all the child classes?

itrushkin:
> May we consider including seed in the interface class' signature instead of all the child classes

@igor-davidyuk Interface should include only required parameters. Fully deterministic behaviour does not require random seed.

itrushkin:
> Will [this line](https://github.com/intel/openfl/blob/5e5ff192612d4e66658c75fc6bf5e69eeaef7736/openfl/utilities/data_splitters/numpy.py#L52) split classes unevenly in case of no shuffle? if the original dataset is ordered by target, like [0,0,0,1,1,1,1], the first collaborator will get only 0th class. If it is true, may we use `idx[rank : len(idx) : cols]` instead of `np.split_array`?

It's true. In this case the first collaborator will receive the first part of data, and the second the remains. Isn't it better to use shuffling here instead of guessing the ordering?

ViktoriiaRomanova (approved):

igor-davidyuk (approved):

PR #296, opened 2022-01-19
Redesign protocols. Now we have two services with their protos: Aggre…

aleksandr-mokrov:
This PR is prerequisite to parallel experiment launching, that means that several aggregators will be launch separately from the director, and the director will communicate with aggregators by gRPC. Also https://github.com/intel/openfl/issues/195 was covered.
Changes:
1. Aggregator service protocols moved to aggregator.proto and rewrite to proto3 syntax. (also additional method were added, but without implementation - this PR is only protocol's part)
2. The common part with base entities separated into base.proto.
3. Protocols are brought to general [style guide](https://developers.google.com/protocol-buffers/docs/style). 
4. Fixed compile script and moved to root (in purpose to keep import hierarchy).
5. RPC methods renamed in style `rpc <RpcName> (<RpcName>Request) returns (<RpcName>Response) {}`, expect stream methods. The following recommendation were followed: https://cloud.google.com/apis/design/naming_convention#method_names
6. CollaboratorGRPCClient is renamed to AggregatorGRPCClient (it's client to Aggregator and it will be used by collaborator and the director)
7. client.py is renamed to aggregator_client.py, server.py to aggregator_server.py to avoid confusion with other servers and clients.


igor-davidyuk:
Do we consider not using TCP for Director-Aggregator communication?
https://docs.microsoft.com/en-us/aspnet/core/grpc/interprocess?view=aspnetcore-6.0

igor-davidyuk:
Also do we consider splitting the Director service in several parts? For the aggregator and the frontend API?

compile_proto.sh
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
psfoley: In the interest of not having too many top level files, this file should either stay where it is (`/openfl/protocols`), or be moved to the `/scripts` folder with other bash scripts used for development / packaging purposes.

Also, it's not needed for this PR, but as a general comment it would be good to have documentation explaining when developers need to run this file (i.e. when submitting PR's that modify existing protocols)

dmitryagapov (approved):

compile_proto.sh
aleksandr-mokrov: complie_proto.sh was returned to openfl/protocols directory with correct build paths. Also I added README file to that directory.

igor-davidyuk (approved):

alexey-gruzdev (approved):

PR #297, opened 2022-01-21
Add dispatching by system and architecture while downloading step CLI

Alexsandruss:
While working with certificates, openfl downloads only linux-x86_64 version of step CLI. This PR targets to add more supported systems and architectures for downloading.

Changes were tested on linux-x86_64 and linux-arm64.

### DIFF
--- a/openfl/component/ca/ca.py
+++ b/openfl/component/ca/ca.py
@@ -73,6 +73,9 @@ def download_step_bin(url, grep_name, architecture, prefix='.', confirmation=Tru
         if (grep_name in a['name'] and architecture in a['name']
             and 'application/gzip' in a['content_type'])
     ]
+    if len(archive_urls) == 0:
+        raise Exception('Applicable CA binaries from github were not found '
+                        f'(name: {grep_name}, architecture: {architecture})')
     archive_url = archive_urls[-1]
     archive_url = archive_url.replace('https', 'http')
     name = archive_url.split('/')[-1]
--- a/openfl/component/ca/ca.py
+++ b/openfl/component/ca/ca.py
@@ -34,15 +34,15 @@ def get_system_and_architecture():
     system = uname_res.system.lower()
 
     architecture_aliases = {
-        ('x86_64', ): 'amd64',
-        ('armv6l', ): 'armv6',
-        ('armv7l', ): 'armv7',
-        ('aarch64', ): 'arm64'
+        'x86_64': 'amd64',
+        'armv6l': 'armv6',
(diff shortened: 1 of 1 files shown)


Alexsandruss:

openfl/component/ca/ca.py
```diff
@@ -27,6 +28,26 @@
...
+    """Get system and architecture of machine."""
+    uname_res = platform.uname()
+    system = uname_res.system.lower()
+
+    architecture_aliases = {
+        ('x86_64', ): 'amd64',
+        ('armv6l', ): 'armv6',
+        ('armv7l', ): 'armv7',
+        ('aarch64', ): 'arm64'
+    }
```
dmitryagapov: Maybe better to use Dict[str, str] instead of Dict[tuple, str]

openfl/component/ca/ca.py
```diff
@@ -27,6 +28,26 @@
...
+    architecture = uname_res.machine
+    for aliases in architecture_aliases:
+        if architecture in aliases:
```
dmitryagapov: If change `architecture_aliases` typing to Dict[str, str], we can compare them using `architecture == aliases`

openfl/component/ca/ca.py
```diff
@@ -27,6 +28,26 @@
...
+            architecture = architecture_aliases[aliases]
+            break
+
+    return system, architecture
```
dmitryagapov, line 48: I think that we need to have some default architecture here (e.g. `amd64`) because in case if we didn't match architecture with alias wrong value will be returned. Or maybe we need to raise exception in this case.

dmitryagapov (changes requested):

openfl/component/ca/ca.py
Alexsandruss: Dict[tuple, str] was used due to suggestion that arch has multiple aliases in 'platform' module, but it was not confirmed for now.

openfl/component/ca/ca.py
Alexsandruss: Changed to Dict[str, str] and `architecture == alias`

igor-davidyuk (approved):

openfl/component/ca/ca.py
Alexsandruss: Arch might be correctly detected without alias, for example, on arm variants. Exception for case with no applicable binaries was added

aleksandr-mokrov (approved):

dmitryagapov (approved):

PR #306, opened 2022-01-31
Implementation of director-aggregator gRPC communication

aleksandr-mokrov:
This is implementation of Director - Aggregator communication by gRPC in order to provide possibility of a separate launch. 
1 . There was implemented the next APIs: GetMetricStream, GetTrainedModel, GetExperimentDescription.
2. All request from user sent to the Director service. The Director service checks a caller, an experiment status and etc, and if it possible sends request to Aggregator service. User -> Federation -> director client -> Director service -> Director -> aggregator client -> Aggregator service -> Aggregator.
3. It was implemented an asynchronous aggregator client (AsyncAggregatorGRPCClient) for asynchronous Director based on the synchronous client.  
4. The async aggregator client makes several attempts to connect to the server in case of unavailability (each second during 30 seconds by default; synchronous aggregator client, that used by collaborators makes it last forever). It implemented using Interceptors. 
5. Aggregator's IP address and ports are taken from the plan (a port is generated by using the plan hash).
6. pytorch_kvasir_unet test was fixed.

There is a restriction: it is possible get an information from an aggregator only when it is alive. It is acceptable for GetMetricStream, but should be resolved for GetTrainedModel in the future (there several way to resolve, but they should be discussed).

### DIFF
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -5,6 +5,7 @@
 import time
 import queue
 from logging import getLogger
+from typing import List
 
 from openfl.interface.aggregation_functions import WeightedAverage
 from openfl.component.straggler_handling_functions import CutoffTimeBasedStragglerHandling
@@ -572,6 +573,55 @@ class Aggregator:
         self.collaborator_tasks_results[task_key] = task_results
 
         self._end_of_task_check(task_name)
+
+    def get_experiment_description(self) -> dict:
+        """Get a experiment information by name for specific user."""
+        progress = self.round_number / self.rounds_to_train
+        model_statuses = self.model_download_statuses
+        tasks = [{
+            'name': task['function'],
+            'description': 'Task description Mock',
+        } for task in self.assigner.tasks.values()]
+        collaborators = [{
+            'name': name,
+            'status': 'pending_mock',
+            'progress': 0.0,
+            'round': 0,
+            'current_task': 'Current Task Mock',
+            'next_task': 'Next Task Mock'
+        } for name in self.authorized_cols]
+        result = {
+            'current_round': self.round_number,
+            'total_rounds': self.rounds_to_train,
+            'download_statuses': {
+                'models': model_statuses,
+                'logs': [{
+                    'name': 'aggregator',
+                    'status': 'ready'
+                }],
+            },
+            'collaborators': collaborators,
+            'tasks': tasks,
+            'progress': progress
+        }
+        return result
+
+    @property
+    def model_download_statuses(self) -> List[dict]:
+        """Return model download statuses representation."""
+        best_model_status = 'ready' if self.best_tensor_dict else 'pending'
+        last_model_status = 'ready' if self.last_tensor_dict else 'pending'
+        model_statuses = [{
+            'name': 'best',
+            'status': best_model_status,
+        }, {
+            'name': 'last',
+            'status': last_model_status,
+        }, {
+            'name': 'init',
+            'status': 'ready'
+        }]
+        return model_statuses
 
     def _process_named_tensor(self, named_tensor, collaborator_name):
         """
--- a/openfl/component/director/director.py
+++ b/openfl/component/director/director.py
@@ -10,9 +10,11 @@ from collections import defaultdict
 from pathlib import Path
 from typing import Callable
 from typing import Iterable
-from typing import List
 from typing import Union
+from typing import ValuesView
 
+from openfl.protocols import base_pb2
+from openfl.transport import AsyncAggregatorGRPCClient
 from openfl.transport.grpc.exceptions import ShardNotFoundError
 
 from .experiment import Experiment
@@ -89,6 +91,26 @@ class Director:
         self.experiments_registry.add(experiment)
         return True
 
+    async def get_aggregator_client(self, experiment_name):
+        """Return an aggregator client for the experiment."""
+        exp = self.experiments_registry[experiment_name]
+        while exp.status != Status.IN_PROGRESS:
+            await asyncio.sleep(1)
+        agg_port = exp.plan.agg_port
+        agg_addr = exp.plan.agg_addr
+        logger.info(f'Aggregator uri: {agg_addr}:{agg_port}')
+
+        aggregator_client = AsyncAggregatorGRPCClient(
+            agg_addr,
+            agg_port,
+            tls=self.tls,
+            disable_client_auth=not self.tls,
+            root_certificate=self.root_certificate,
+            certificate=self.certificate,
+            private_key=self.private_key
+        )
+        return aggregator_client
+
     async def get_experiment_status(
             self,
             experiment_name: str,
@@ -100,26 +122,23 @@ class Director:
             return None
         return self.experiments_registry[experiment_name].status
 
-    def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
+    async def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
         """Get trained model."""
         if (experiment_name not in self.experiments_registry
                 or caller not in self.experiments_registry[experiment_name].users):
             logger.error('No experiment data in the stash')
             return None
-
-        aggregator = self.experiments_registry[experiment_name].aggregator
-
-        if aggregator.last_tensor_dict is None:
-            logger.error('Aggregator have no aggregated model to return')
+        exp = self.experiments_registry[experiment_name]
+        if exp.status != Status.IN_PROGRESS:
             return None
 
-        if model_type == 'best':
-            return aggregator.best_tensor_dict
-        elif model_type == 'last':
-            return aggregator.last_tensor_dict
-        else:
-            logger.error('Unknown model type required.')
-            return None
+        aggregator_client = await self.get_aggregator_client(experiment_name)
+        trained_model = await aggregator_client.get_trained_model(
+            experiment_name,
+            model_type
+        )
+
+        return trained_model
 
     def get_experiment_data(self, experiment_name: str) -> Path:
         """Get experiment data."""
@@ -176,19 +195,9 @@ class Director:
                 f' does not have access to this experiment'
             )
 
-        while not self.experiments_registry[experiment_name].aggregator:
-            await asyncio.sleep(1)
-        aggregator = self.experiments_registry[experiment_name].aggregator
-
-        while True:
-            if not aggregator.metric_queue.empty():
-                yield aggregator.metric_queue.get()
-                continue
-
-            if aggregator.all_quit_jobs_sent() and aggregator.metric_queue.empty():
-                return
-
-            yield None
+        aggregator_client = await self.get_aggregator_client(experiment_name)
+        async for metric_dict in aggregator_client.get_metric_stream():
+            yield metric_dict
 
     def remove_experiment_data(self, experiment_name: str, caller: str):
         """Remove experiment data from stash."""
@@ -241,7 +250,7 @@ class Director:
 
         return self.envoy_health_check_period
 
-    def get_envoys(self) -> list:
+    def get_envoys(self) -> ValuesView:
         """Get a status information about envoys."""
         logger.info(f'Shard registry: {self._shard_registry}')
         for envoy_info in self._shard_registry.values():
@@ -250,11 +259,11 @@ class Director:
                 + envoy_info.get('valid_duration', 0)
             )
             envoy_name = envoy_info['shard_info']['node_info']['name']
-            envoy_info['experiment_name'] = self.col_exp[envoy_name]
+            envoy_info['experiment_name'] = self.col_exp.get(envoy_name)
 
         return self._shard_registry.values()
 
-    def get_experiments_list(self, caller: str) -> list:
+    async def get_experiments_list(self, caller: str) -> list:
         """Get experiments list for specific user."""
         experiments = self.experiments_registry.get_user_experiments(caller)
         result = []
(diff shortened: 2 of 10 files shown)


openfl/transport/grpc/aggregator_server.py
```diff
@@ -206,9 +209,102 @@ def SendLocalTaskResults(self, request, context):  # NOQA:N802
...
+                download_statuses=download_statuses,
+                collaborators=collaborators,
+                tasks=tasks,
+            ),
+        )
+
     def get_server(self):
         """Return gRPC server."""
-        self.server = server(ThreadPoolExecutor(max_workers=cpu_count()),
+        self.server = server(ThreadPoolExecutor(max_workers=cpu_count() + 1),
```
dmitryagapov, line 345: Why should we add extra thread?

openfl/component/director/director.py
```diff
@@ -85,26 +87,41 @@ async def set_new_experiment(
         self.experiments_registry.add(experiment)
         return True
 
-    def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
+    def get_aggregator_client(self, experiment_name):
+        """Return an aggregator client for the experiment."""
+        exp = self.experiments_registry[experiment_name]
+        agg_port = exp.plan.agg_port
+        agg_addr = exp.plan.agg_addr
```
dmitryagapov, line 100: Maybe it should be `agg_host`?

openfl/component/director/director.py
```diff
@@ -85,26 +87,41 @@ async def set_new_experiment(
...
             logger.error('No experiment data in the stash')
             return None
-
-        aggregator = self.experiments_registry[experiment_name].aggregator
-
-        if aggregator.last_tensor_dict is None:
-            logger.error('Aggregator have no aggregated model to return')
+        exp = self.experiments_registry[experiment_name]
+        if exp.status != Status.IN_PROGRESS:
             return None
```
dmitryagapov, line 133: It looks like the final model could not be got because experiment will have `Status.FINISHED`.
Maybe we can create special tasks `wait` amd `stop`. `wait` task could give director time to get final model. `stop` task will finish the aggregator process from aggregator side.

openfl/component/director/director.py
```diff
@@ -155,17 +172,10 @@ async def stream_metrics(self, experiment_name: str, caller: str):
 
         while not self.experiments_registry[experiment_name].aggregator:
             await asyncio.sleep(1)
```
dmitryagapov: Maybe we shouldn't have `aggregator` inside of experiment if we are going to have only grpc communication between argerator and director?

openfl/component/director/director.py
```diff
@@ -238,32 +251,20 @@ def get_experiments_list(self, caller: str) -> list:
...
-                    'status': 'ready'
-                }],
-            },
-            'collaborators': collaborators,
-            'tasks': tasks,
-            'progress': progress
-        }
-        return result
+        if exp.status != Status.IN_PROGRESS:
+            return base_pb2.ExperimentDescription(status=exp.status)
```
dmitryagapov: I think that we need to return full `ExperimentDescription` in any case.

openfl/component/director/experiment.py
```diff
@@ -92,11 +95,11 @@ def _create_aggregator_grpc_server(
             private_key: Union[Path, str] = None,
             certificate: Union[Path, str] = None,
     ) -> AggregatorGRPCServer:
-        plan = Plan.parse(plan_config_path=Path(self.plan_path))
-        plan.authorized_cols = list(self.collaborators)
+        self.plan = Plan.parse(plan_config_path=Path(self.plan_path))
```
dmitryagapov: Maybe we don't need to have plan as experiment attribute because it is using only in this function

openfl/component/director/experiment.py
```diff
@@ -76,6 +78,7 @@ async def start(
                     certificate=certificate,
                 )
                 self.aggregator = aggregator_grpc_server.aggregator
+                self.aggregator_server = aggregator_grpc_server
```
dmitryagapov: Maybe we don't need to have `aggregator_server` as experiment attribute because it's never used?

openfl/transport/grpc/aggregator_client.py
```diff
@@ -164,27 +232,26 @@ def create_insecure_channel(self, uri):
         """
         return grpc.insecure_channel(uri, options=self.channel_options)
 
-    def create_tls_channel(self, uri, root_certificate, disable_client_auth,
-                           certificate, private_key):
+    def get_credentials(self, root_certificate, disable_client_auth,
```
dmitryagapov: may be static

openfl/transport/grpc/aggregator_client.py
```diff
@@ -251,8 +343,24 @@ def reconnect(self):
...
+        """Create a server stub."""
+        # Adding an interceptor for RPC Errors
+        self.interceptors = (
+            RetryOnRpcErrorClientInterceptor(
+                sleeping_policy=ConstantBackoff(
+                    reconnect_interval=int(kwargs.get('client_reconnect_interval', 1)),
+                    uri=self.uri),
+                status_for_retry=(grpc.StatusCode.UNAVAILABLE,),
+            ),
+        )
```
dmitryagapov: I don't think that we shoud have this variable, at least it shouldn't be `self.`

openfl/transport/grpc/aggregator_client.py
```diff
@@ -301,21 +409,109 @@ def send_local_task_results(self, collaborator_name, round_number,
...
+            AsyncRetryOnRpcErrorUnaryUnaryClientInterceptor(
+                reconnect_interval=int(self.kwargs.get('client_reconnect_interval', 1)),
+                connect_timeout=int(self.kwargs.get('client_connect_timeout', 30)),
+                status_for_retry=(grpc.StatusCode.UNAVAILABLE,),
+            ),
+            AsyncRetryOnRpcErrorUnaryStreamClientInterceptor(
+                reconnect_interval=int(self.kwargs.get('client_reconnect_interval', 1)),
+                connect_timeout=int(self.kwargs.get('client_connect_timeout', 30)),
+                status_for_retry=(grpc.StatusCode.UNAVAILABLE,),
+            ),
```
dmitryagapov: Maybe we need to create global variables for default `client_reconnect_interval` and `client_connect_timeout`

dmitryagapov (changes requested):

openfl/transport/grpc/aggregator_server.py
```diff
@@ -206,9 +209,102 @@ def SendLocalTaskResults(self, request, context):  # NOQA:N802
...
+                logger.debug('Try to get metric dict')
+                metric_dict = self.aggregator.metric_queue.get(timeout=1)
+            except queue.Empty:
+                if self.aggregator.all_quit_jobs_sent() and self.aggregator.metric_queue.empty():
+                    return
+                continue
+            logger.debug('Metric dict has got. Yield it')
+            yield aggregator_pb2.GetMetricStreamResponse(**metric_dict)
+
+    def GetTrainedModel(self, request, context):  # NOQA:N802
```
psfoley, line 260: If you are going to use the same gRPC server for director and collaborators, there should to be some way to filter between them to limit access only to intended APIs for each role. The three prior endpoints all used the helper function `validate_collaborator`, so a helper function like `validate_director` would be of use here

openfl/component/director/director.py
aleksandr-mokrov: We got it (parameter named agg_addr) from different old configs and pass as kwargs further.  Renaming only here will only confuse.

openfl/component/director/director.py
aleksandr-mokrov: I suggest discuss how we can handle it out of this PR and implement in an additional PR.

openfl/component/director/director.py
aleksandr-mokrov: I removed it from this and whenever it possible, but keep `aggregator` inside of experiment only for set_experiment_failed. We should decide how we will stop experiment and aggregator when we will merge PR with dockerization. In the simplest way we can add an extra RPC.

openfl/component/director/director.py
aleksandr-mokrov: `name` parameter was added, but we haven't more info about experiment in that case yet.

openfl/component/director/experiment.py
aleksandr-mokrov: Removed

openfl/component/director/experiment.py
aleksandr-mokrov: It used by director to get port and address from plan.

openfl/transport/grpc/aggregator_server.py
aleksandr-mokrov: To avoid case when we have only CPU and only thread in the thread pool. Moreover, it looks strange, that the number of thread depends from CPU count.

openfl/transport/grpc/aggregator_server.py
aleksandr-mokrov: Method `validate_director_request`. It checks that a subject name from client certificate matches an aggregator FQDN (Director and Aggregator FQDNs are same and we set director FQDN as a subject name for certificate https://openfl.readthedocs.io/en/latest/source/utilities/pki.html#semi-automatic-certification). Collaborator's and user's request will be rejected.

openfl/transport/grpc/aggregator_client.py
aleksandr-mokrov: done

openfl/transport/grpc/aggregator_client.py
aleksandr-mokrov: done

openfl/transport/grpc/aggregator_client.py
aleksandr-mokrov: done

dmitryagapov (approved):

openfl/component/aggregator/aggregator.py
```diff
@@ -532,6 +533,60 @@ def send_local_task_results(self, collaborator_name, round_number, task_name,
...
+            'collaborators': collaborators,
+            'tasks': tasks,
+            'progress': progress
+        }
+        return result
+
+    @property
+    def experiment_progress(self):
+        """Return experiment progress."""
+        return self.round_number / self.rounds_to_train
```
igor-davidyuk: could we just go with the current round number? 😁 without floats

openfl/component/director/director.py
```diff
@@ -85,26 +87,43 @@ async def set_new_experiment(
         self.experiments_registry.add(experiment)
         return True
 
-    def get_trained_model(self, experiment_name: str, caller: str, model_type: str):
+    async def get_aggregator_client(self, experiment_name):
+        """Return an aggregator client for the experiment."""
+        exp = self.experiments_registry[experiment_name]
+        while exp.status != Status.IN_PROGRESS:
+            await asyncio.sleep(1)
```
igor-davidyuk, line 98: could we keep all the async code in transport components?

openfl/component/director/director.py
igor-davidyuk: Could we get the best model 'so far' while the training is still continuing?

openfl/component/director/director.py
```diff
@@ -85,26 +87,43 @@ async def set_new_experiment(
...
 
-        if model_type == 'best':
-            return aggregator.best_tensor_dict
-        elif model_type == 'last':
-            return aggregator.last_tensor_dict
-        else:
-            logger.error('Unknown model type required.')
-            return None
+        aggregator_client = await self.get_aggregator_client(experiment_name)
+        trained_model = await aggregator_client.get_trained_model(
```
igor-davidyuk, line 136: again, async code, can we isolate it in transport components?

openfl/federated/plan/plan.py
```diff
@@ -234,6 +236,12 @@ def hash(self):  # NOQA
 
         return self.hash_.hexdigest()
 
+    def get_agg_port(self):
```
igor-davidyuk:
```suggestion
    def generate_agg_port(self):
```

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/envoy_config.yaml
```diff
@@ -8,4 +8,4 @@ shard_descriptor:
   params:
     data_folder: kvasir_data
     rank_worldsize: 1,90
-    enforce_image_hw: [300, 400]
\ No newline at end of file
+    enforce_image_hw: '300, 400'
```
igor-davidyuk: did you turn a list into one string? What is the reason?

openfl/component/aggregator/aggregator.py
aleksandr-mokrov: Removed

openfl/component/director/director.py
aleksandr-mokrov: In that implementation, where we get it information from an aggregator, it is impossible. But in the next step an aggregator will send to director, and it reduces async code in the director.

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/envoy_config.yaml
aleksandr-mokrov: It was a bug, we use string in the current implementation.

theakshaypant:
@aleksandr-mokrov can you please resolve the conflicts and request for review again?

PR #314, opened 2022-02-08
Passing pip install options

igor-davidyuk:
Workspace export routine is now aware of pip install options in requirements files.
Users may now pass options for remote pip install execution.

### DIFF
--- a/openfl/interface/interactive_api/experiment.py
+++ b/openfl/interface/interactive_api/experiment.py
@@ -8,6 +8,7 @@ from collections import defaultdict
 from copy import deepcopy
 from logging import getLogger
 from pathlib import Path
+from typing import Tuple
 
 from tensorboardX import SummaryWriter
 
@@ -17,7 +18,7 @@ from openfl.federated import Plan
 from openfl.interface.cli import setup_logging
 from openfl.interface.cli_helper import WORKSPACE
 from openfl.utilities import split_tensor_dict_for_holdouts
-from openfl.utilities.utils import is_package_versioned
+from openfl.utilities.workspace import dump_requirements_file
 
 
 class ModelStatus:
@@ -140,7 +141,8 @@ class FLExperiment:
 
         self.logger.info(log_message)
 
-    def prepare_workspace_distribution(self, model_provider, task_keeper, data_loader):
+    def prepare_workspace_distribution(self, model_provider, task_keeper, data_loader,
+                                       pip_install_options: Tuple[str] = ()):
         """Prepare an archive from a user workspace."""
         # Save serialized python objects to disc
         self._serialize_interface_objects(model_provider, task_keeper, data_loader)
@@ -149,14 +151,16 @@ class FLExperiment:
 
         # PACK the WORKSPACE!
         # Prepare requirements file to restore python env
-        self._export_python_env()
+        dump_requirements_file(keep_original_prefixes=True,
+                               prefixes=pip_install_options)
 
         # Compress te workspace to restore it on collaborator
         self.arch_path = self._pack_the_workspace()
 
     def start(self, *, model_provider, task_keeper, data_loader,
               rounds_to_train, delta_updates=False, opt_treatment='RESET',
-              device_assignment_policy='CPU_ONLY'):
+              device_assignment_policy='CPU_ONLY',
+              pip_install_options: Tuple[str] = ()):
         """Prepare experiment and run."""
         self._prepare_plan(model_provider, task_keeper, data_loader,
                            rounds_to_train,
@@ -167,7 +171,8 @@ class FLExperiment:
                            dataloader_interface_file='loader_obj.pkl')
 
         self.prepare_workspace_distribution(
-            model_provider, task_keeper, data_loader
+            model_provider, task_keeper, data_loader,
+            pip_install_options
         )
 
         self.logger.info('Starting experiment!')
@@ -194,17 +199,6 @@ class FLExperiment:
         self.task_runner_stub = self.plan.get_core_task_runner(model_provider=model_provider)
         self.current_model_status = ModelStatus.RESTORED
         self.experiment_accepted = True
-
-    @staticmethod
-    def _export_python_env():
-        """Prepare requirements.txt."""
-        from pip._internal.operations import freeze
-        requirements_generator = freeze.freeze()
-
-        with open('./requirements.txt', 'w') as f:
-            for pack in requirements_generator:
-                if is_package_versioned(pack):
-                    f.write(pack + '\n')
 
     @staticmethod
     def _pack_the_workspace():
--- a/openfl/interface/interactive_api/experiment.py
+++ b/openfl/interface/interactive_api/experiment.py
@@ -184,7 +184,7 @@ class FLExperiment:
             will always be 'cpu',
             'CUDA_PREFERRED' - enable passing CUDA device identifiers to tasks
             by collaborators, works with cuda-device-monitor plugin equipped Envoys.
-        pip_install_options - list of options for the remote `pip install` calls,
+        pip_install_options - tuple of options for the remote `pip install` calls,
             example: ('-f some.website', '--no-index')
         """
         self._prepare_plan(model_provider, task_keeper, data_loader,
--- a/openfl/interface/interactive_api/experiment.py
+++ b/openfl/interface/interactive_api/experiment.py
@@ -158,10 +158,35 @@ class FLExperiment:
         self.arch_path = self._pack_the_workspace()
 
     def start(self, *, model_provider, task_keeper, data_loader,
-              rounds_to_train, delta_updates=False, opt_treatment='RESET',
-              device_assignment_policy='CPU_ONLY',
-              pip_install_options: Tuple[str] = ()):
-        """Prepare experiment and run."""
+              rounds_to_train: int, delta_updates: bool = False,
+              opt_treatment: str = 'RESET',
+              device_assignment_policy: str = 'CPU_ONLY',
+              pip_install_options: Tuple[str] = ()) -> None:
+        """
+        Prepare workspace distribution and send to Director.
+
+        A successful call of this function will result in sending the experiment workspace
+        to the Director service and experiment start.
+
+        Parameters:
+        model_provider - Model Interface instance.
+        task_keeper - Task Interface instance.
+        data_loader - Data Interface instance.
+        rounds_to_train - required number of training rounds for the experiment.
+        delta_updates - [bool] Tells if collaborators should send delta updates
+            for the locally tuned models. If set to False, whole checkpoints will be sent.
+        opt_treatment - Optimizer state treatment policy.
+            Valid options: 'RESET' - reinitialize optimizer for every round,
+            'CONTINUE_LOCAL' - keep local optimizer state,
+            'CONTINUE_GLOBAL' - aggregate optimizer state.
+        device_assignment_policy - device assignment policy.
+            Valid options: 'CPU_ONLY' - device parameter passed to tasks
+            will always be 'cpu',
+            'CUDA_PREFERRED' - enable passing CUDA device identifiers to tasks
+            by collaborators, works with cuda-device-monitor plugin equipped Envoys.
+        pip_install_options - list of options for the remote `pip install` calls,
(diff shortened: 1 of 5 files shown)


openfl/interface/workspace.py
```diff
@@ -109,7 +110,12 @@ def create(prefix, template):
 
 
 @workspace.command(name='export')
-def export_():
+@option('-o', '--pip-install-options', required=False,
+        type=str, multiple=True, default=tuple,
+        help='Options for remote pip install. '
+        'You may pass several options in quotation marks alongside with arguments, '
```
aleksandr-mokrov: Can you align text to text (Second and third row of help text under first row, not under argument `help`)?

openfl/interface/workspace.py
```diff
@@ -124,19 +130,14 @@ def export_():
...
         echo(f'Plan file "{plan_file}" not found. No freeze performed.')
 
-    from pip._internal.operations import freeze
-    requirements_generator = freeze.freeze()
-    with open('./requirements.txt', 'w') as f:
-        for package in requirements_generator:
-            if is_package_versioned(package):
-                f.write(package + '\n')
+    # Dump requirements.txt
+    dump_requirements_file(prefixes=pip_install_options, keep_original_prefixes=True)
```
aleksandr-mokrov, line 140: dump_requirements_file reads requirements.txt file to get prefixes form it and then rewrite that file by merged prefixes (pip_install_options with prefixes from requirements.txt) and installed packages (from `pip freeze`). Probably we should explicitly use two different file, one as a source of prefixes  and write to other?

openfl/interface/interactive_api/experiment.py
```diff
@@ -149,15 +151,42 @@ def prepare_workspace_distribution(self, model_provider, task_keeper, data_loade
...
+        opt_treatment - Optimizer state treatment policy.
+            Valid options: 'RESET' - reinitialize optimizer for every round,
+            'CONTINUE_LOCAL' - keep local optimizer state,
+            'CONTINUE_GLOBAL' - aggregate optimizer state.
+        device_assignment_policy - device assignment policy.
+            Valid options: 'CPU_ONLY' - device parameter passed to tasks
+            will always be 'cpu',
+            'CUDA_PREFERRED' - enable passing CUDA device identifiers to tasks
+            by collaborators, works with cuda-device-monitor plugin equipped Envoys.
+        pip_install_options - list of options for the remote `pip install` calls,
```
aleksandr-mokrov: It's a tuple

igor-davidyuk:
```suggestion
        pip_install_options - tuple of options for the remote `pip install` calls,
```

openfl/interface/workspace.py
```diff
@@ -109,7 +110,12 @@ def create(prefix, template):
 
 
 @workspace.command(name='export')
-def export_():
+@option('-o', '--pip-install-options', required=False,
+        type=str, multiple=True, default=tuple,
+        help='Options for remote pip install. '
+        'You may pass several options in quotation marks alongside with arguments, '
+        'e.g. -o "--find-links source.site"')
```
igor-davidyuk:
```suggestion
        help='Options for remote pip install. '
          'You may pass several options in quotation marks alongside with arguments, '
          'e.g. -o "--find-links source.site"')
```

openfl/interface/workspace.py
igor-davidyuk: flake8:
./openfl/interface/workspace.py:116:11: E127 continuation line over-indented for visual indent

openfl/interface/interactive_api/experiment.py
igor-davidyuk: fixed

openfl/interface/workspace.py
igor-davidyuk: we do not manage different versions of requirements files

openfl/interface/workspace.py
```diff
@@ -109,7 +110,12 @@ def create(prefix, template):
 
 
 @workspace.command(name='export')
-def export_():
+@option('-o', '--pip-install-options', required=False,
+        type=str, multiple=True, default=tuple,
+        help='Options for remote pip install. '
+          'You may pass several options in quotation marks alongside with arguments, '
+          'e.g. -o "--find-links source.site"')
```
igor-davidyuk:
```suggestion
        'You may pass several options in quotation marks alongside with arguments, '
        'e.g. -o "--find-links source.site"')
```

openfl/utilities/workspace.py
```diff
@@ -75,3 +78,52 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    if keep_original_prefixes and path.is_file():
+        with open(path) as f:
+            while True:
+                line = f.readline()
+                if line == '':
+                    continue
+                if line[0] == '-':
+                    prefixes |= {line.replace('\n', '')}
+                else:
+                    break
```
dmitryagapov:
```suggestion
            for line in f:
                if line[0] == '-':
                    prefixes |= {line.replace('\n', '')}
                else:
                    break
```

tests/openfl/utilities/test_dump_requirements.py
```diff
@@ -0,0 +1,69 @@
...
+    expected_options = []
+    for li in expected_lines:
+        if li[0] == '-':
+            expected_options.append(li)
+        else:
+            break
+    expected_packages = expected_lines[len(expected_options):]
+
+    assert len(read_options) == len(expected_options)
+    assert len(read_packages) == len(expected_packages)
```
dmitryagapov, line 70: This is not required.

openfl/interface/workspace.py
aleksandr-mokrov: If you add only tab, then it will be that error, but if you align text to text ('You ...' strictly under 'Options...' ). Like this https://github.com/intel/openfl/blob/2d66cd816a78670d6a99ef04669f5fa47bcf45f7/openfl/interface/collaborator.py#L255

tests/openfl/utilities/test_dump_requirements.py
```diff
@@ -0,0 +1,69 @@
...
+    monkeypatch.setattr(freeze, 'freeze', mock_pip_freeze)
+
+    dump_requirements_file(path=requirements_file,
+                           keep_original_prefixes=keep_original_prefixes,
+                           prefixes=prefixes)
+
+    with open(requirements_file) as f:
+        read_lines = f.readlines()
+
+    read_options = []
```
dmitryagapov, line 53: I don't think that `read_options` and `expected_options` are the good naming

tests/openfl/utilities/test_dump_requirements.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.mark.parametrize('keep_original_prefixes,prefixes,expected_lines', [
+    (False, None, LINES[1:]),
+    (True, None, LINES),
+    (False, PREFIXES, [p + '\n' for p in PREFIXES] + [li for li in LINES if li[0] != '-']),
+    (True, PREFIXES, [li for li in LINES if li[0] == '-'
+                      ] + [p + '\n' for p in PREFIXES] + [li for li in LINES if li[0] != '-']),
+    (True, PREFIXES_OVERLAP, [p + '\n' for p in PREFIXES_OVERLAP
+                              ] + [li for li in LINES if li[0] != '-']),
+])
```
dmitryagapov: To simplify it, maybe make sense:
@pytest.mark.parametrize('keep_original_prefixes,prefixes,expected_lines', [
    (False, None, REQ_WITHOUT_PREFIXES),
    (True, None, REQ_WITH_PREFIXES),
    (False, PREFIXES, PREFIXES + REQ_WITHOUT_PREFIXES,
    (True, PREFIXES, PRESIXES + REQ_WITH_PREFIXES,
    (True, PREFIXES_OVERLAP, PREFIXES_OVERLAP + REQ_WITH_PREFIXES),
])
Because multiple list comprehension hard to read.
Also maybe this test should be separated to increase readability.

dmitryagapov (changes requested):

dmitryagapov:
I think that we need to test such cases as requirements.txt doesn't have prefixes and empty requirements.txt

tests/openfl/utilities/test_dump_requirements.py
igor-davidyuk: why?

tests/openfl/utilities/test_dump_requirements.py
igor-davidyuk: Do you have any suggestions?

tests/openfl/utilities/test_dump_requirements.py
igor-davidyuk: These list comprehension will be hard to read but in another place

openfl/utilities/workspace.py
igor-davidyuk: replaced while with for

tests/openfl/utilities/test_dump_requirements.py
dmitryagapov: Because `set(read_options) == set(expected_options)` can't be equal if its have different length

igor-davidyuk: x = ['a', 'a', 'b']
y = ['a', 'b']
set(x) == set(y)  # True

tests/openfl/utilities/test_dump_requirements.py
igor-davidyuk: Replaced list comprehension with variables

igor-davidyuk:
> I think that we need to test such cases as requirements.txt doesn't have prefixes and empty requirements.txt

Added additional tests

tests/openfl/utilities/test_dump_requirements.py
```diff
@@ -0,0 +1,125 @@
...
+def test_dump_empty_original_list(
+        monkeypatch, touch_file,
+        keep_original_prefixes, prefixes, expected_lines):
+    """Test dump_requirements_file function with no file to start."""
+    def mock_pip_freeze():
+        return [line.replace('\n', '') for line in ORIGINAL_PACKAGE_LINES]
+    monkeypatch.setattr(freeze, 'freeze', mock_pip_freeze)
+
+    requirements_file = Path('./test_requirements_2.txt').absolute()
```
aleksandr-mokrov, line 95: You can use fixture like` requirements_file` but with args to pass `touch_file`, or tmp_path to be sure that file will be unlinked if test fails with exception before `requirements_file.unlink()`.

aleksandr-mokrov (approved):

openfl/utilities/workspace.py
```diff
@@ -75,3 +78,51 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        prefixes = set(prefixes)
+
+    # Merge prefixes:
+    # We expect that all the prefixes in a requirement file
+    # are placed at the top
+    if keep_original_prefixes and path.is_file():
+        with open(path) as f:
+            for line in f:
+                if line == '':
+                    continue
```
dmitryagapov: Looks like it is not needed. Or maybe you found corner case for it?

dmitryagapov (approved):

tests/openfl/utilities/test_dump_requirements.py
igor-davidyuk: I believe parametrizing the fixture will lead to overengineering this test module. The problem is we can not broadcast expected results to different values of the `touch_file` parameter.
Instead, I introduced `try - finally` construction to make sure the `test_requirements` file is always unlinked.

openfl/utilities/workspace.py
igor-davidyuk: This is a mistake. I fixed it to state
```
if line == '\n':
    continue
```
The idea is to skip empty lines between options in the original requirements.txt, for instance:
```
-o option1

-s option2
requirement1==0.0.1
```

tests/openfl/utilities/test_dump_requirements.py
dmitryagapov: I thought that `read` is verb)

alexey-gruzdev (approved):

PR #338, opened 2022-02-15
General configuration mechanism

### LINKED ISSUE #158: General configuration mechanism
I propose to create a general mechanism for receiving configurable data and exclude the default values for them from the code, and set them in default configurable file. For example, the director will take params from cli, if they are passed, otherwise from director.yaml in the director workspace, otherwise from `openfl-workspace/default/director.yaml`.

dmitryagapov:
Use configs and input CLI commands in the same way for different components


tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/envoy_config.yaml
```diff
@@ -1,11 +1,11 @@
...
+optional_plugin_components: { }
 
 shard_descriptor:
   template: kvasir_shard_descriptor.KvasirShardDescriptor
   params:
     data_folder: kvasir_data
     rank_worldsize: 1,90
-    enforce_image_hw: [300, 400]
\ No newline at end of file
+    enforce_image_hw: '300,400'
```
psfoley, line 11: Minor request. Can this be changed back to:

``` enforce_image_hw: [300,400] ``` 

and then make the director config consistent with this? It's a bit confusing to have ```[ str , str , str ]``` in one config and ```int,int``` in another

File: openfl/utilities/utils.py [hunk 2]
```diff
@@ -202,11 +207,32 @@ def validate_file_hash(file_path, expected_hash, chunk_size=8192):
...
+        data_preparation: Optional[List[Tuple[str, Callable]]] = None,
+        **kwargs,
+) -> Dynaconf:
+    """Create Dynaconf settings, merge its with `overwrite_dict` and validate result."""
+    settings = Dynaconf(**kwargs, YAML_LOADER='safe_load')
+    if overwrite_dict:
+        for key, value in overwrite_dict.items():
+            if value is not None or settings.get(key) is None:
+                settings.set(key, value, merge=True)
+    if data_preparation:
```
psfoley: This is interesting. Because the purpose is transforming the provided value in some way to match the expected format, I think this could be generalized beyond `data_preparation` (an example would be fixing a malformed URL provided by the user in the case that `merge_configs` was extended to other functions). Maybe change the name to `value_transform`?

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/envoy/envoy_config.yaml
dmitryagapov: It was changed because in original tutorial it looks like:
```yaml
enforce_image_hw: '300,400'
```
https://github.com/intel/openfl/blob/develop/openfl-tutorials/interactive_api/PyTorch_Kvasir_UNet/envoy/envoy_config.yaml

In any case we should deside how it should look and fix it in all places.

alexey-gruzdev:
Jenkins, please retry a build

alexey-gruzdev:
Jenkins please retry a build

alexey-gruzdev:
Jenkins please retry a build

dmitryagapov:
Jenkins please retry a build

aleksandr-mokrov (approved):

psfoley (approved):

igor-davidyuk (approved):

alexey-gruzdev (approved):

PR #339, opened 2022-02-15
Gramine + OpenFL

igor-davidyuk:
# OpenFL + Gramine
This manual will help you run OpenFL with Aggregator-based workflow inside SGX enclave with Gramine.

## Prerequisites
Building machine:
- OpenFL

- SGX enabled in BIOS
- Ubuntu with Linux kernel 5.11+
- ? SGX drivers, it is built in kernel: `/dev/sgx_enclave`
- [aesmd service](https://github.com/intel/linux-sgx) (`/var/run/aesmd/aesm.socket`)
This is a short list, see more in [Gramine docs](https://gramine.readthedocs.io/en/latest/devel/building.html).

## Workflow
The user will mainly interact with OpenFL CLI, docker CLI, and other command-line tools. But the user is also expected to modify plan.yaml file and Python code under workspace/src folder to set up an FL Experiment.
### On the building machine (Data Scientist's node):
1. As usual, **create a workspace**: 
```
export WORKSPACE_NAME=my_sgx_federation_workspace
export TEMPLATE_NAME=torch_cnn_histology

fx workspace create --prefix $WORKSPACE_NAME --template $TEMPLATE_NAME
cd $WORKSPACE_NAME
```
Modify the code and the plan.yaml, set up your training procedure. </br>
Pay attention to the following: 
- make sure the data loading code reads data from ./data folder inside the workspace
- if you download data (development scenario) make sure your code first checks if data exists, as connecting to the internet from an enclave may be problematic.
- make sure you do not use any CUDA driver-dependent packages

Default workspaces (templates) in OpenFL differ in their data downloading procedures. Workspaces with data loading flow that do not require changes to run with Gramine include:
- torch_unet_kvasir
- torch_cnn_histology
- keras_nlp

2. **Initialize the experiment plan** </br> 
Find out the FQDN of the aggregator machine and use it for plan initialization.
For example, on Unix-like OS try the following command:
```
hostname --all-fqdns | awk '{print $1}'
```
(In case this FQDN does not work for your federation, try putting the machine IP instead)
Then pass the result as `AGG_FQDN` parameter to:
```
fx plan initialize -a $AGG_FQDN
```

3. (Optional) **Generate a signing key** on the building machine if you do not have one.</br>
It will be used to calculate hashes of trusted files. If you plan to test the application without SGX (gramine-direct) you also do not need a signer key.
```
export KEY_LOCATION=.

openssl genrsa -3 -out $KEY_LOCATION/key.pem 3072
```
This key will not be packed into the final Docker image.

4. **Build the Experiment Docker image**

```
fx workspace graminize -s $KEY_LOCATION/key.pem
```
This command will build and save a Docker image with your Experiment. The saved image will contain all the required files to start a process in an enclave.</br>
If a signing key is not provided, the application will be built without SGX support, but it still can be started with gramine-direct executable.

### Image distribution:
Data scientist (builder) now must transfer the Docker image to the aggregator and collaborator machines. The Aggregator will also need initial model weights.

5. **Transfer files** to the aggregator and collaborator machines.
If there is a connection between machines, you may use `scp`. In other cases use the transfer channel that suits your situation.</br>
Send files to the aggregator machine:
```
scp BUILDING_MACHINE:WORKSPACE_PATH/WORKSPACE_NAME.tar.gz AGGREGATOR_MACHINE:SOME_PATH
scp BUILDING_MACHINE:WORKSPACE_PATH/save/WORKSPACE_NAME_init.pbuf AGGREGATOR_MACHINE:SOME_PATH
```

Send the image archive to collaborator machines:
```
scp BUILDING_MACHINE:WORKSPACE_PATH/WORKSPACE_NAME.tar.gz COLLABORATOR_MACHINE:SOME_PATH
```

Please, keep in mind, if you run a test Federation, with data downloaded from the internet, you should also transfer/download data to collaborator machines.

### On the running machines (Aggregator and Collaborator nodes):
6. **Load the image.**
Execute the following command on all running machines:
```
docker load < WORKSPACE_NAME.tar.gz
```

7. **Prepare certificates**
Certificates exchange is a big separate topic. To run an experiment following OpenFL Aggregator-based workflow, a user must follow the established procedure, please refer to [the docs](https://openfl.readthedocs.io/en/latest/running_the_federation.html#bare-metal-approach).
Following the above-mentioned procedure, running machines will acquire certificates. Moreover, as the result of this procedure, the aggregator machine will also obtain a `cols.yaml` file (required to start an experiment) with registered collaborators' names, and the collaborator machines will obtain `data.yaml` files.

We recommend replicating the OpenFL workspace folder structure on all the machines and following the usual certifying procedure. Finally, on the aggregator node you should have the following folder structure:
```
workspace/
--save/WORKSPACE_NAME_init.pbuf
--logs/
--plan/cols.yaml
--cert/
  --client/*col.crt
  --server/
    --agg_FQDN.crt
    --agg_FQDN.key
```

On collaborator nodes:
```
workspace/
--data/*dataset*
--plan/data.yaml
--cert/
  --client/
    --col_name.crt
    --col_name.key
```

To speed up the certification process for one-node test runs, it makes sense to utilize the OpenFL integration test script [make this a link after merge]  openfl/tests/github/test_graminize.sh, that will create required folders and certify an aggregator and two collaborators.

#### On the Aggregator machine run:
```
export WORKSPACE_NAME=your_workspace_name
export WORKSPACE_PATH=path_to_workspace
docker run -it --rm --device=/dev/sgx_enclave --volume=/var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket \
--network=host \
--volume=${WORKSPACE_PATH}/cert:/workspace/cert \
--volume=${WORKSPACE_PATH}/logs:/workspace/logs \
--volume=${WORKSPACE_PATH}/plan/cols.yaml:/workspace/plan/cols.yaml \
--mount type=bind,src=${WORKSPACE_PATH}/save,dst=/workspace/save,readonly=0 \
${WORKSPACE_NAME} aggregator start
```

#### On the Collaborator machines run:
```
export WORKSPACE_NAME=your_workspace_name
export WORKSPACE_PATH=path_to_workspace
export COL_NAME=col_name
docker run -it --rm --device=/dev/sgx_enclave --volume=/var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket \
--network=host \
--volume=${WORKSPACE_PATH}/cert:/workspace/cert \
--volume=${WORKSPACE_PATH}/plan/data.yaml:/workspace/plan/data.yaml \
--volume=${WORKSPACE_PATH}/data:/workspace/data \
${WORKSPACE_NAME} collaborator start -n ${COL_NAME}
```

### **No SGX run (`gramine-direct`)**:
The user may run an experiment under gramine without SGX. Note how we do not mount `sgx_enclave` device and pass a `--security-opt` instead that allows syscalls required by `gramine-direct`

#### On the Aggregator machine run:
```
export WORKSPACE_NAME=your_workspace_name
export WORKSPACE_PATH=path_to_workspace
docker run -it --rm --security-opt seccomp=unconfined -e GRAMINE_EXECUTABLE=gramine-direct \
--network=host \
--volume=${WORKSPACE_PATH}/cert:/workspace/cert \
--volume=${WORKSPACE_PATH}/logs:/workspace/logs \
--volume=${WORKSPACE_PATH}/plan/cols.yaml:/workspace/plan/cols.yaml \
--mount type=bind,src=${WORKSPACE_PATH}/save,dst=/workspace/save,readonly=0 \
${WORKSPACE_NAME} aggregator start
```

#### On the Collaborator machines run:

```
export WORKSPACE_NAME=your_workspace_name
export WORKSPACE_PATH=path_to_workspace
export COL_NAME=col_name
docker run -it --rm --security-opt seccomp=unconfined -e GRAMINE_EXECUTABLE=gramine-direct \
--network=host \
--volume=${WORKSPACE_PATH}/cert:/workspace/cert \
--volume=${WORKSPACE_PATH}/plan/data.yaml:/workspace/plan/data.yaml \
--volume=${WORKSPACE_PATH}/data:/workspace/data \
${WORKSPACE_NAME} collaborator start -n ${COL_NAME}
```

## The Routine
Gramine+OpenFL PR brings in `openfl-gramine` folder, that contains the following files:
 - MANUAL.md - this manual
 - Dockerfile.gamine - the base image Dockerfile for all experiments, it starts from Python3.8 image and installs OpenFL and Gramine packages.
 - Dockerfile.graminized.workspace - this one is for building the final experiment image. It starts from the previous image and imports the experiment archive (executes 'fx workspace import') inside an image. At this stage, we have an experiment workspace and all the requirements installed inside the image. Then it runs a unified Makefile that uses the openfl.manifest.template to prepare required files to run OpenFL under gramine inside an SGX enclave.
 - Makefile - follows regular gramine workflow, please refer to (gramine docs)[] for more info
 - openfl.manifest.template - gramine manifest template, it is the same for all the experiments
 - start_process.sh - bash script used to start an OpenFL actor in a container.

There is a files access peculiarity that should be kept in mind during debugging and development.
Both Dockerfiles are read from the bare-metal OpenFL installation, i.e. from an OpenFL package on a building machine.
While the gramine manifest template and the Makefile are read in image build time from the local (in-image) OpenFL package. </br>
Thus, if one wants to make changes to the gramine manifest template or the Makefile, they should change the OpenFL installation procedure in Dockerfile.gramine, so their changes may be pulled to the base image. One option is to push the changes to a GitHub fork and install OpenFL from this fork. 
```
*Dockerfile.gramine:*

RUN git clone https://github.com/your-username/openfl.git --branch some_branch
WORKDIR /openfl
RUN --mount=type=cache,target=/root/.cache/ \
    pip install --upgrade pip && \
    pip install .
WORKDIR /
```
In this case, to rebuild the image, use `fx workspace dockerize --rebuild` with `--rebuild` flag that will pass '--no-cache' to docker build command.

Another option is to copy OpenFL source files from an on-disk cloned repo, but it would mean that the user must build the graminized image from the repo directory using Docker CLI.

## Known issues:
- Kvasir experiment: aggregation takes really long, debug log-level does not show the reason
- We need workspace zip to import it and create certs. We need to know the number of collaborators prior to zipping the workspace. SOLUTION: mount cols.yaml and data.yaml
- During plan initialization we need data to initialize the model. so at least one collaborator should be in data.yaml and its data should be available. cols.yaml may be empty at first
During cert sign request generation cols.yaml on collaborators remain empty, data.yaml is extended if needed. On aggregator, cols.yaml are updated during signing procedure, data.yaml remains unmodified
- `error: Disallowing access to file '/usr/local/lib/python3.8/__pycache__/signal.cpython-38.pyc.3423950304'; file is not protected, trusted or allowed.`

 ## TO-DO:
- [X] import manifest and makefile from OpenFL dist-package 
- [X] pass wheel repository to pip (for CPU versions of PyTorch for example)
- [ ] get rid of command line args (insecure)
- [ ] introduce `fx workspace create --prefix WORKSPACE_NAME` command without --template option to the OpenFL CLI, which will create just an empty workspace with the right folder structure.
- [ ] introduce `fx *actor* start --from image`

### DIFF
--- a/openfl-gramine/Dockerfile.gramine
+++ b/openfl-gramine/Dockerfile.gramine
@@ -7,7 +7,7 @@ FROM ${BASE_IMAGE}
 #     pip install openfl
 
 # for now install from source (branch with logs_writer turned off)
-RUN git clone https://github.com/igor-davidyuk/openfl.git --branch gramine
+RUN git clone https://github.com/igor-davidyuk/openfl.git --branch gramine-rebased
 WORKDIR /openfl
 RUN --mount=type=cache,target=/root/.cache/ \
     pip install --upgrade pip && \
--- a/openfl-gramine/Dockerfile.gramine
+++ b/openfl-gramine/Dockerfile.gramine
@@ -1,0 +1,38 @@
+ARG BASE_IMAGE=python:3.8
+FROM ${BASE_IMAGE}
+
+# install openfl
+# RUN --mount=type=cache,target=/root/.cache/ \
+#     pip install --upgrade pip && \
+#     pip install openfl
+
+# for now install from source (branch with logs_writer turned off)
+RUN git clone https://github.com/igor-davidyuk/openfl.git --branch gramine
+WORKDIR /openfl
+RUN --mount=type=cache,target=/root/.cache/ \
+    pip install --upgrade pip && \
+    pip install .
+WORKDIR /
+
+# install gramine
+RUN curl -fsSLo /usr/share/keyrings/gramine-keyring.gpg https://packages.gramineproject.io/gramine-keyring.gpg && \
+    echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/gramine-keyring.gpg] https://packages.gramineproject.io/ stable main' | \
+    tee /etc/apt/sources.list.d/gramine.list
+RUN --mount=type=cache,id=apt-dev,target=/var/cache/apt \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+    gramine libprotobuf-c-dev
+    # there is an issue for libprotobuf-c in gramine repo, install from apt for now
+
+# graminelibos is under this dir
+ENV PYTHONPATH=/usr/local/lib/python3.8/site-packages/:/usr/lib/python3/dist-packages/:
+
+# install linux headers
+# WORKDIR /tmp/
+# RUN wget -c https://kernel.ubuntu.com/~kernel-ppa/mainline/v5.11/amd64/linux-headers-5.11.0-051100_5.11.0-051100.202102142330_all.deb
+# RUN dpkg -i *.deb
+# RUN mv /usr/src/linux-headers-5.11.0-051100/ /usr/src/linux-headers-5.11.0-051100rc5-generic/
+# WORKDIR /
+
+# ENV LC_ALL=C.UTF-8
+# ENV LANG=C.UTF-8
--- a/openfl-gramine/MANUAL.md
+++ b/openfl-gramine/MANUAL.md
@@ -1,19 +1,19 @@
 # OpenFL + Gramine
 This manual will help you run OpenFL with Aggregator-based workflow inside SGX enclave with Gramine.
 ## TO-DO:
-- [X] import manifest and makefile from dist-package openfl 
-- [X] pass wheel repository to pip (for cpu versions of pytorch for example)
+- [X] import manifest and makefile from OpenFL dist-package 
+- [X] pass wheel repository to pip (for CPU versions of PyTorch for example)
 - get rid of command line args (insecure)
 ## Known issues:
 - Kvasir experiment: aggregation takes really long, debug log-level does not show the reason
-- We need workspace zip to import it and create certs. We need to know number of collaborators prior to zipping the workspace. SOLUTOIN: mount cols.yaml and data.yaml
+- We need workspace zip to import it and create certs. We need to know the number of collaborators prior to zipping the workspace. SOLUTOIN: mount cols.yaml and data.yaml
 - During plan initialization we need data to initialize the model. so at least one collaborator should be in data.yaml and its data should be available. cols.yaml may be empty at first
 During cert sign request generation cols.yaml on collaborators remain empty, data.yaml is extended if needed. On aggregator, cols.yaml are updated during signing procedure, data.yaml remains unmodified
 - `error: Disallowing access to file '/usr/local/lib/python3.8/__pycache__/signal.cpython-38.pyc.3423950304'; file is not protected, trusted or allowed.`
 ## Prerquisites
 Building machine:
 - OpenFL
-- Docker should be installed, user included into Docker group
+- Docker should be installed, user included in Docker group
 
 Machines that will run an Aggregator and Collaborator containers should have the following:
 - SGX enebled in BIOS
@@ -23,7 +23,7 @@ Machines that will run an Aggregator and Collaborator containers should have the
 This is a short list, see more in Gramine docs.
 
 ## Workflow
-The user will mainly interact with OpenFL CLI, docker CLI and other commandline tools. But the user is also expected to modify plan.yaml file and Python code under workspace/src folder to set up an FL Experiment.
+The user will mainly interact with OpenFL CLI, docker CLI, and other command-line tools. But the user is also expected to modify plan.yaml file and Python code under workspace/src folder to set up an FL Experiment.
 ### On the building machine (Data Scientist's node):
 1. As usual, **create a workspace**: 
 ```
@@ -32,38 +32,38 @@ cd WORKSPACE_NAME
 ```
 Modify the code and the plan.yaml, set up your training procedure. </br>
 Pay attention to the following: 
-- make sure data loading code reads data from ./data folder inside the workspace
-- if you download data (developement scenario) make sure your code first checks if data exists, as connecting to the internet from an enclave may be problematic.
+- make sure the data loading code reads data from ./data folder inside the workspace
+- if you download data (development scenario) make sure your code first checks if data exists, as connecting to the internet from an enclave may be problematic.
 - make sure you do not use any CUDA driver-dependent packages
 
 2. **Initialize the experiment plan** </br> 
-Find out the FQDN of the aggregator machine and use for plan initialization.
+Find out the FQDN of the aggregator machine and use it for plan initialization.
 ```
 fx plan initialize -a ${FQDN}
 ```
 To find out FQDN (Unix-like OS) try `hostname --all-fqdns | awk '{print $1}'` command.
 
-3. (Optional) **Generate a signing key** on building machine if you do not have one.</br>
+3. (Optional) **Generate a signing key** on the building machine if you do not have one.</br>
 It will be used to calculate hashes of trusted files. If you plan to test the application without SGX (gramine-direct) you also do not need a signer key.
 ```
 openssl genrsa -3 -out KEY_LOCATION/key.pem 3072
 ```
-This key will not be packed to the final Docker image.
+This key will not be packed into the final Docker image.
 
 4. **Build the Experiment Docker image**
 
 ```
 fx workspace graminize -s KEY_LOCATION/key.pem --sgx-target/--no-sgx-target
 ```
-This command will build and save a Docker image with you Experiment. The saved image will contain all the required files to start a process in an enclave.</br>
-If `--no-sgx-target` option passed to the command, the image will run processes under gramine-direct, in this case it is not necessary to pass signing key.
+This command will build and save a Docker image with your Experiment. The saved image will contain all the required files to start a process in an enclave.</br>
+If `--no-sgx-target` option is passed to the command, the image will run processes under gramine-direct, in this case, it is not necessary to pass the signing key.
 
 
 ### Image distribution:
-Data scientist now must transfer the Docker image to the aggregator and collaborator machines. Aggregator will also need initial model weights.
+Data scientist (builder) now must transfer the Docker image to the aggregator and collaborator machines. The Aggregator will also need initial model weights.
 
 5. **Transfer files** to the aggregator an collaborator machines.
-If there is a connaction between machines, you may use `scp`. In other case use the transfer channel that suits your situation.</br>
+If there is a connection between machines, you may use `scp`. In other cases use the transfer channel that suits your situation.</br>
 Send files to the aggregator machine:
 ```
 scp BUILDING_MACHINE:WORKSPACE_PATH/WORKSPACE_NAME.tar.gz AGGREGATOR_MACHINE:SOME_PATH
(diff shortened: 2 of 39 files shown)


mansishr:
In readme, step number 4 should be fx workspace graminize and not dockerize.

igor-davidyuk:
> In readme, step number 4 should be fx workspace graminize and not dockerize.

fixed

openfl-gramine/MANUAL.md
```diff
@@ -0,0 +1,189 @@
...
+```
+fx plan initialize -a $AGG_FQDN
+```
+
+3. (Optional) **Generate a signing key** on the building machine if you do not have one.</br>
+It will be used to calculate hashes of trusted files. If you plan to test the application without SGX (gramine-direct) you also do not need a signer key.
+```
+export KEY_LOCATION=.
+
+openssl genrsa -3 -out $KEY_LOCATION/key.pem 3072
```
psfoley, line 55: This week the gramine project added a function to generate the signing key with the cryptography package: https://github.com/gramineproject/gramine/commit/3def085a9689d6642d2ad9d86e962e227d8b0ca4

We should reuse/copy this functionality to avoid adding OpenSSL as another dependency

openfl-gramine/Dockerfile.gramine
```diff
@@ -0,0 +1,38 @@
...
+    echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/gramine-keyring.gpg] https://packages.gramineproject.io/ stable main' | \
+    tee /etc/apt/sources.list.d/gramine.list
+RUN --mount=type=cache,id=apt-dev,target=/var/cache/apt \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+    gramine libprotobuf-c-dev
+    # there is an issue for libprotobuf-c in gramine repo, install from apt for now
+
+# graminelibos is under this dir
+ENV PYTHONPATH=/usr/local/lib/python3.8/site-packages/:/usr/lib/python3/dist-packages/:
```
mansishr, line 28: This path might be different for different users. Maybe we could add a note for users to know that we are using this as PYTHONPATH.

openfl-gramine/openfl.manifest.template
```diff
@@ -0,0 +1,80 @@
...
+fs.mount.lib2.path = "{{ arch_libdir }}"
+fs.mount.lib2.uri = "file:{{ arch_libdir }}"
+
+fs.mount.usr.type = "chroot"
+fs.mount.usr.path = "/usr"
+fs.mount.usr.uri = "file:/usr"
+
+fs.mount.tmp.type = "chroot"
+fs.mount.tmp.path = "/tmp"
+fs.mount.tmp.uri = "file:/tmp"
```
mansishr: I think we could try removing /tmp directory from manifest to see if it works fine

openfl-gramine/MANUAL.md
igor-davidyuk: I've just checked and this command is not in the package available via apt yet

openfl-gramine/Dockerfile.gramine
igor-davidyuk: This path is actually fixed as it Is inside the container (or image 🤷‍♂️)

openfl-gramine/openfl.manifest.template
igor-davidyuk: I removed mounting `/tmp` and it worked, thank you

openfl-gramine/MANUAL.md
```diff
@@ -0,0 +1,191 @@
...
+- [ ] introduce `fx *actor* start --from image`
+
+## Known issues:
+- Kvasir experiment: aggregation takes really long, debug log-level does not show the reason
+- We need workspace zip to import it and create certs. We need to know the number of collaborators prior to zipping the workspace. SOLUTION: mount cols.yaml and data.yaml
+- During plan initialization we need data to initialize the model. so at least one collaborator should be in data.yaml and its data should be available. cols.yaml may be empty at first
+During cert sign request generation cols.yaml on collaborators remain empty, data.yaml is extended if needed. On aggregator, cols.yaml are updated during signing procedure, data.yaml remains unmodified
+- `error: Disallowing access to file '/usr/local/lib/python3.8/__pycache__/signal.cpython-38.pyc.3423950304'; file is not protected, trusted or allowed.`
+
+## Prerquisites
```
igor-davidyuk:
```suggestion
## Prerequisites
```

openfl/interface/workspace.py
```diff
@@ -422,6 +427,93 @@ def dockerize_(context, base_image, save):
...
+            echo(line)
+        _ = process.communicate()  # pipe is already empty, used to get `returncode`
+        if process.returncode != 0:
+            raise Exception('\n ❌ Execution failed\n')
+
+    from openfl.interface.cli_helper import SITEPACKS
+
+    # We can build for gramine-sgx and run with gramine-direct,
+    # but not vice versa.
+    sgx_build = 1 if signing_key.is_file() else 0
```
dmitryagapov: Use `True` / `False` here would be great. Also try to rename variable to start it with `is_`.

openfl/interface/workspace.py
```diff
@@ -422,6 +427,93 @@ def dockerize_(context, base_image, save):
...
+    """
+    def open_pipe(command: str):
+        echo(f'\n 📦 Executing command:\n{command}\n')
+        process = subprocess.Popen(
+            command,
+            shell=True, stderr=subprocess.STDOUT,
+            stdout=subprocess.PIPE)
+        for line in process.stdout:
+            echo(line)
+        _ = process.communicate()  # pipe is already empty, used to get `returncode`
```
dmitryagapov, line 471: Looks like better to use `process.wait()`

openfl-workspace/keras_nlp_gramine_ready/src/dataloader_utils.py
```diff
@@ -0,0 +1,220 @@
+"""Copyright (C) 2020-2021 Intel Corporation.
+
+Licensed subject to the terms of the separately executed evaluation
+license agreement between Intel Corporation and you.
+"""
+from logging import getLogger
+from os import getcwd
+from os import path
```
dmitryagapov, line 8: It would be great if `os.path` has been replaces with `pathlib`

dmitryagapov (changes requested):

openfl-workspace/keras_nlp_gramine_ready/src/dataloader_utils.py
igor-davidyuk: In another PR, sure.
This line is not introduced in this PR

aleksandr-mokrov (approved):

openfl/interface/workspace.py
igor-davidyuk: Unfortunately, this will not work
https://docs.python.org/3/library/subprocess.html

openfl/interface/workspace.py
igor-davidyuk: done

dmitryagapov (approved):

alexey-gruzdev (approved):

psfoley:
Approved

PR #351, opened 2022-02-25
Optimizer parameters need to be copied if different tasks have model on different devices

### LINKED ISSUE #298: RuntimeError while training and validating on different devices
**The Bug**
Training and validation tasks could not be executed on different devices.
The current device assigning policy does not allow specifying devices per task, but if the device is overridden inside a task code (e.g. for CPU inference), an error is raised (see below).

**Steps To Reproduce**
1. Open any tutorial folder in `openfl-tutorials/interactive_api`. I found the bug in `PyTorch_Kvasir_UNet`.
2. Create and activate `director` Python environment.
3. Execute:
    ```
    pip install /path/to/openfl
    cd director
    bash start_director.sh
    ``` 
4. In a separate terminal create and activate `envoy` Python environment.
5. Execute:
    ```
    pip install /path/to/openfl
    cd envoy
    pip install sd_requirements.txt
    bash start_envoy.sh
    ```
3. In a separate terminal create and activate `workspace` Python environment.
4. Execute:
    ```
    pip install /path/to/openfl
    pip install torchvision==0.8.1 # for Kvasir workspace
    jupyter notebook
    ```
5. Open the Jupyter UI and go to the main experiment notebook in the `workspace` folder.
6. In tasks definition, add a device overriding line so devices for different tasks differ. Since for Kvasir example `CUDA_PREFERRED` policy is set, all tasks get `cuda:*` devices in parameters. So we need to change it to `cpu`:
    ```python
    @TI.register_fl_task(model='unet_model', data_loader='val_loader', device='device')     
    def validate(unet_model, val_loader, device):
        device='cpu'  #  HERE WE OVERRIDE THE DEVICE
        print(f'\n\n TASK VALIDATE GOT DEVICE {device}\n\n')
    
        unet_model.eval()
        unet_model.to(device)
    
        val_loader = tqdm.tqdm(val_loader, desc="validate")

        val_score = 0
        total_samples = 0

        with torch.no_grad():
            for data, target in val_loader:
                samples = target.shape[0]
                total_samples += samples
                data, target = torch.tensor(data).to(device), \
                    torch.tensor(target).to(device, dtype=torch.int64)
                output = unet_model(data)
                val = soft_dice_coef(output, target)
                val_score += val.sum().cpu().numpy()
                
        return {'dice_coef': val_score / total_samples,}
    ```
7. Run all cells.
8. During `fl_experiment.stream_metrics()` an error occurs:
    ```
    [15:53:29] ERROR    Collaborator failed with error: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!:  envoy.py:75
                    Traceback (most recent call last):                                                                                                          
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/openfl/component/envoy/envoy.py", line 73, in run                    
                        self._run_collaborator()                                                                                                                
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/openfl/component/envoy/envoy.py", line 141, in                       
                    _run_collaborator                                                                                                                           
                        col.run()                                                                                                                               
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/openfl/component/collaborator/collaborator.py", line 145,            
                    in run                                                                                                                                      
                        self.do_task(task, round_number)                                                                                                        
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/openfl/component/collaborator/collaborator.py", line 241,            
                    in do_task                                                                                                                                  
                        global_output_tensor_dict, local_output_tensor_dict = func(                                                                             
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/openfl/federated/task/task_runner.py", line 129, in                  
                    collaborator_adapted_task                                                                                                                   
                        metric_dict = callable_task(**task_kwargs)                                                                                              
                      File                                                                                                                                      
                    "/home/itrushkin/.virtualenvs/openfl_research/lib/python3.8/site-packages/openfl/interface/interactive_api/experiment.py", line             
                    436, in wrapper_decorator                                                                                                                   
                        metric_dict = training_method(**task_keywords)                                                                                          
                      File "<ipython-input-13-b3468c8701fc>", line 50, in train                                                                                 
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 26, in                            
                    decorate_context                                                                                                                            
                        return func(*args, **kwargs)                                                                                                            
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/torch/optim/adam.py", line 108, in step                              
                        F.adam(params_with_grad,                                                                                                                
                      File "/home/itrushkin/.virtualenvs/envoy/lib/python3.8/site-packages/torch/optim/functional.py", line 86, in adam                         
                        exp_avg.mul_(beta1).add_(grad, alpha=1 - beta1)                                                                                         
                    RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
    ```
**Expected behavior**
Training is completed without errors.

**Desktop**
OS: Ubuntu 18.04
Python: 3.8
CUDA: 11.2
PyTorch: 1.7.0

mansishr:
Fixes #298 

The optimizer parameters need to be on the same device where model is. When a model's tensors are moved from one device to the other, the optimizer parameters will be different objects before and after it is moved. Solution is to transfer the optimizer's parameters after the model is moved. Usage is in train task:

```
from openfl.utilities.optimizers.torch import optimizer_to

def train(net_model, train_loader, optimizer, device):
      device = torch.device('cuda')

      net_model.train()
      net_model.to(device)
      optimizer_to(optimizer,device)
```

### DIFF
--- a/openfl/utilities/optimizers/torch/__init__.py
+++ b/openfl/utilities/optimizers/torch/__init__.py
@@ -4,3 +4,4 @@ import pkgutil
 if pkgutil.find_loader('torch'):
     from .fedprox import FedProxOptimizer # NOQA
     from .fedprox import FedProxAdam # NOQA
+    from .optimdevice import optimizer_to # NOQA
--- a/openfl/utilities/optimizers/torch/optimdevice.py
+++ b/openfl/utilities/optimizers/torch/optimdevice.py
@@ -1,0 +1,20 @@
+"""PyTorch optimizer utility to transfer the optimizer's parametes to current device."""
+
+"""
+This function is needed to transfer the optimizer parameters to the same device where the model is.
+This is because optimized parameters must live in consistent locations when optimizers are constructed and used.
+Resembles model.to(device) for optimizer.
+
+Usage (in train task after the model is transferred to device):
+    optimizer_to(optimizer,device)
+"""
+
+import torch
+
+def optimizer_to(optim, device):
+    for param in optim.state.values():
+        for subparam in param.values():
+            if isinstance(subparam, torch.Tensor):
+                subparam.data = subparam.data.to(device)
+                if subparam._grad is not None:
+                    subparam._grad.data = subparam._grad.data.to(device)
--- a/openfl/utilities/optimizers/torch/optimdevice.py
+++ b/openfl/utilities/optimizers/torch/optimdevice.py
@@ -4,14 +4,18 @@ import torch
 
 
 def optimizer_to(optim, device):
-    """This function is needed to transfer optimizer
-    parameters to the same device where the model is.
+    """Transfer optimizer to device.
+
+    Transfer optimizer's parameters to the same device where the model is.
     This is because optimized parameters must live in consistent locations
     when optimizers are constructed and used.
     Resembles model.to(device) for optimizer.
 
     Usage (in train task after the model is transferred to device):
             optimizer_to(optimizer,device)
+    Args:
+        optimizer   :  Torch optimizer
+        device      :  model's device
     """
     for param in optim.state.values():
         for subparam in param.values():
--- a/openfl/utilities/optimizers/torch/optimdevice.py
+++ b/openfl/utilities/optimizers/torch/optimdevice.py
@@ -1,17 +1,18 @@
 """PyTorch optimizer utility to transfer the optimizer's parametes to current device."""
 
-"""
-This function is needed to transfer the optimizer parameters to the same device where the model is.
-This is because optimized parameters must live in consistent locations when optimizers are constructed and used.
-Resembles model.to(device) for optimizer.
-
-Usage (in train task after the model is transferred to device):
-    optimizer_to(optimizer,device)
-"""
-
 import torch
 
+
 def optimizer_to(optim, device):
+    """This function is needed to transfer optimizer
+    parameters to the same device where the model is.
+    This is because optimized parameters must live in consistent locations
+    when optimizers are constructed and used.
+    Resembles model.to(device) for optimizer.
+
+    Usage (in train task after the model is transferred to device):
+            optimizer_to(optimizer,device)
+    """
     for param in optim.state.values():
         for subparam in param.values():
             if isinstance(subparam, torch.Tensor):


dmitryagapov (approved):

itrushkin:
@mansishr We always need optimizer parameters to be on the same device as model parameters. Should we move the responsibility of device synchronization from user to OpenFL?

mansishr:
> @mansishr We always need optimizer parameters to be on the same device as model parameters. Should we move the responsibility of device synchronization from user to OpenFL?

I had thought about that but the optimizer's parameters need to be moved after the model has been moved. In our interactive API examples, we move the model in the train task so the optimizer has to be moved after it in the train task. If we don't allow users to move model in train task, then we could move both the model and optimizer through OpenFL but I think we give flexibility to the user like that.

itrushkin (approved):

igor-davidyuk:
> @mansishr We always need optimizer parameters to be on the same device as model parameters. Should we move the responsibility of device synchronization from user to OpenFL?

I believe this is too much. If some user wants to use a specific device (should we even allow this?) it is the user's responsibility to move all the tensors to this device.

In my opinion, we should reuse this PR to check another scenario.
We know that a Collaborator decides which device will be assigned to some task prior to task execution. We should make sure that changing the assigned device works well for a task without user interference.
The complete scenario: task `train` is assigned 'cpu' device, the framework adapter gets the tensors to set to the model and the optimizer along with the device; it sets the tensors and changes the device. Then task `validation` gets the device 'cuda:0', tensors go through the same procedure. On the next round, the task `train` gets the device 'cuda:1' and the training should go smoothly.

mansishr:
> 

@igor-davidyuk I checked this scenario and it works fine. The framework adapter sets different devices correctly and changes the model and optimizer as needed.

igor-davidyuk:
@alexey-gruzdev please review this PR, in my opinion this util goes against our device assignment philosophy

alexey-gruzdev:
After internal discussion, we close this due to creation of more general problem statement in https://github.com/intel/openfl/issues/399.

PR #363, opened 2022-03-02
Custom task assigner unit tests.

aleksandr-mokrov:
Units tests covering custom_assigner and FLexperiment's define_task_assigner method.

### DIFF
--- a/openfl/interface/interactive_api/experiment.py
+++ b/openfl/interface/interactive_api/experiment.py
@@ -57,6 +57,8 @@ class FLExperiment:
 
         self.experiment_accepted = False
 
+        self.validation_task_exist = False
+
         self.logger = getLogger(__name__)
         setup_logging()
 
@@ -227,23 +229,22 @@ class FLExperiment:
         else:
             self.logger.info('Experiment was not accepted or failed.')
 
-    @staticmethod
-    def define_task_assigner(task_keeper, rounds_to_train):
+    def define_task_assigner(self, task_keeper, rounds_to_train):
         """Define task assigner by registered tasks."""
         tasks = task_keeper.get_registered_tasks()
         train_task_exist = False
-        validation_task_exist = False
+        self.validation_task_exist = False
         for task in tasks.values():
             if task.task_type == 'train':
                 train_task_exist = True
             if task.task_type == 'validate':
-                validation_task_exist = True
+                self.validation_task_exist = True
 
         if not train_task_exist and rounds_to_train != 1:
             # Since we have only validation tasks, we do not have to train it multiple times
             raise Exception('Variable rounds_to_train must be equal 1, '
                             'because only validation tasks were given')
-        if train_task_exist and validation_task_exist:
+        if train_task_exist and self.validation_task_exist:
             def assigner(collaborators, round_number, **kwargs):
                 tasks_by_collaborator = {}
                 for collaborator in collaborators:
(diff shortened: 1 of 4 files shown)


tests/openfl/component/assigner/test_custom_assigner.py
```diff
@@ -0,0 +1,98 @@
...
+from openfl.component.aggregation_functions import WeightedAverage
+from openfl.component.assigner.custom_assigner import Assigner
+from openfl.component.assigner.tasks import TrainTask
+from openfl.component.assigner.tasks import ValidateTask
+
+
+def assigner_function(collaborators, round_number, **kwargs):
+    """Return tasks by collaborator."""
+    tasks_by_collaborator = {}
+    for collaborator in collaborators:
```
igor-davidyuk, line 36: could we also introduce tests for the case when the collaborator `two` is assigned just the validation task?

aleksandr-mokrov: I thought about it, but looks like there're no difference for assigner and that tests will check only test assigner function, not assigner.

igor-davidyuk (approved):

openfl/interface/interactive_api/experiment.py
```diff
@@ -233,19 +232,19 @@ def start(self, *, model_provider, task_keeper, data_loader,
...
         tasks = task_keeper.get_registered_tasks()
-        self.train_task_exist = False
+        train_task_exist = False
         self.validation_task_exist = False
         for task in tasks.values():
             if task.task_type == 'train':
-                self.train_task_exist = True
+                train_task_exist = True
             if task.task_type == 'validate':
                 self.validation_task_exist = True
```
dmitryagapov: Maybe make sense to rewrite `train_task_exist` and `self.validate_task_exist` as properties `self.is_train_task_exist` and `self.is_validate_task_exist`.

dmitryagapov (approved):

openfl/interface/interactive_api/experiment.py
```diff
@@ -233,19 +232,19 @@ def start(self, *, model_provider, task_keeper, data_loader,
     def define_task_assigner(self, task_keeper, rounds_to_train):
         """Define task assigner by registered tasks."""
         tasks = task_keeper.get_registered_tasks()
-        self.train_task_exist = False
+        train_task_exist = False
```
mansishr: Is there a reason to not define `train_task_exist `as `self.train_task_exist`

tests/openfl/component/assigner/test_custom_assigner.py
```diff
@@ -0,0 +1,98 @@
...
+    collaborators = assigner.get_collaborators_for_task('train', 2)
+
+    assert collaborators == ['one', 'two']
+
+
+def test_get_all_tasks_for_round(assigner):
+    """Test `get_all_tasks_for_round` base working."""
+    all_tasks = assigner.get_all_tasks_for_round(2)
+
+    assert len(all_tasks) == 3
```
mansishr: Should we also not check the task names here as well?

aleksandr-mokrov: Added

openfl/interface/interactive_api/experiment.py
aleksandr-mokrov: This variable is used only inside this function. And this variables looks like crutches, I don't think that it makes sense to use them like class object attributes to provide them to user. But if need arises in the future, it's easy to fix.

openfl/interface/interactive_api/experiment.py
aleksandr-mokrov: They were renamed, but I keep they scope and didn't make them as properties. It will be easy to fix if needed.

mansishr:
Looks good to me

alexey-khorkin (approved):

alexey-gruzdev (approved):

PR #369, opened 2022-03-07
Unify tensor tags

### LINKED ISSUE #316: Get organised with tensor tags type.
Tensor tags, that are used by the aggregator, can be different types inside the aggregator in different methods. It can be list, tuple, str or <class 'google.protobuf.pyext._message.RepeatedScalarContainer'>. One methods cast it to list, others to tuple. We should unify it.

mansishr:
Fixes #316 

Tensor tags are now tuples, with the exception of converting them to lists for modification but then changed back to tuples.


alexey-khorkin (changes requested):
@mansishr I think more changes are needed here.
If we will use grep command:
```bash
~/Repositories/alexey-khorkin-repos/openfl$ grep -R "tags = " ./
./openfl/protocols/base.proto:  repeated string tags = 5;
./openfl/protocols/utils.py:    tensor_name, origin, round_number, report, tags = tensor_key
./openfl/protocols/aggregator.proto:  repeated string   tags = 5;
./openfl/pipelines/tensor_codec.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/pipelines/tensor_codec.py:            new_tags = tuple(list(tags) + ['compressed'])
./openfl/pipelines/tensor_codec.py:            new_tags = tuple(list(tags) + ['lossy_compressed'])
./openfl/pipelines/tensor_codec.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/pipelines/tensor_codec.py:            new_tags = list(tags)
./openfl/pipelines/tensor_codec.py:            new_tags = list(tags)
./openfl/pipelines/tensor_codec.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/pipelines/tensor_codec.py:            new_tags = tuple([tensor_key[3]] + ['delta'])
./openfl/pipelines/tensor_codec.py:            new_tags = tuple(list(tags) + ['delta'])
./openfl/pipelines/tensor_codec.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/pipelines/tensor_codec.py:            tags = list(tags)
./openfl/pipelines/tensor_codec.py:            new_tags = tuple(tags)
./openfl/pipelines/tensor_codec.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/pipelines/tensor_codec.py:                    new_tags = ('aggregated', 'delta', 'lossy_compressed')
./openfl/pipelines/tensor_codec.py:                    new_tags = ('aggregated', 'delta', 'compressed')
./openfl/transport/grpc/aggregator_server.py:        tags = request.tags
./openfl/federated/task/runner_keras.py:        tags = ('trained',)
./openfl/federated/task/runner_keras.py:        tags = ('metric', suffix)
./openfl/federated/task/runner_fe.py:        tags = ('trained',)
./openfl/federated/task/runner_fe.py:        tags = ('metric', suffix)
./openfl/federated/task/runner_pt.py:        tags = ('metric', suffix)
./openfl/federated/task/runner_pt.py:        tags = ('trained',)
./openfl/federated/task/runner_pt.py:    state_subkey_tags = []
./openfl/federated/task/runner_pt.py:    state_subkeys_and_tags = list(zip(state_subkeys, state_subkey_tags))
./openfl/federated/task/runner_pt.py:    state_subkeys_and_tags = []
./openfl/federated/task/task_runner.py:            tags = ('trained',)
./openfl/federated/task/task_runner.py:            tags = (suffix,)
./openfl/federated/task/task_runner.py:        tags = ('metric', *tags)
./openfl/federated/task/runner_tf.py:        tags = ('trained',)
./openfl/federated/task/runner_tf.py:        tags = ('metric', suffix)
./openfl/plugins/frameworks_adapters/pytorch_adapter.py:    state_subkey_tags = []
./openfl/plugins/frameworks_adapters/pytorch_adapter.py:    state_subkeys_and_tags = list(zip(state_subkeys, state_subkey_tags))
./openfl/plugins/frameworks_adapters/pytorch_adapter.py:    state_subkeys_and_tags = []
./openfl/databases/tensor_db.py:                tensor_name, origin, fl_round, report, tags = tensor_key
./openfl/databases/tensor_db.py:        tensor_name, origin, fl_round, report, tags = tensor_key
./openfl/databases/tensor_db.py:        tensor_name, origin, fl_round, report, tags = tensor_key
./openfl/databases/tensor_db.py:                new_tags = tuple([tags] + [col])
./openfl/databases/tensor_db.py:                new_tags = tuple(list(tags) + [col])
./openfl/component/aggregator/aggregator.py:        tags = list(tags)
./openfl/component/aggregator/aggregator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/aggregator/aggregator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/aggregator/aggregator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/aggregator/aggregator.py:            dec_name, dec_origin, dec_round_num, dec_report, dec_tags = dec_tk
./openfl/component/aggregator/aggregator.py:                new_tags = tuple([dec_tags] + [collaborator_name])
./openfl/component/aggregator/aggregator.py:                new_tags = tuple(list(dec_tags) + [collaborator_name])
./openfl/component/aggregator/aggregator.py:            dec_name, dec_origin, dec_round_num, dec_report, dec_tags = dec_tk
./openfl/component/aggregator/aggregator.py:                new_tags = tuple([dec_tags] + [collaborator_name])
./openfl/component/aggregator/aggregator.py:                new_tags = tuple(list(dec_tags) + [collaborator_name])
./openfl/component/aggregator/aggregator.py:            tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/aggregator/aggregator.py:            new_tags = tuple(tags[:-1])
./openfl/component/aggregator/aggregator.py:            agg_tensor_name, agg_origin, agg_round_number, agg_report, agg_tags = agg_tensor_key
./openfl/component/collaborator/collaborator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/collaborator/collaborator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/collaborator/collaborator.py:            tensor_name, origin, fl_round, report, tags = tensor
./openfl/component/collaborator/collaborator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl/component/collaborator/collaborator.py:        tensor_name, origin, round_number, report, tags = tensor_key
./openfl-tutorials/Federated_Pytorch_MNIST_custom_aggregation_Tutorial.ipynb:    "        self.tags = tags\n",
./openfl-tutorials/Federated_Pytorch_MNIST_custom_aggregation_Tutorial.ipynb:    "                tags = set(self.tags + [local_tensor.col_name])\n",
./openfl-tutorials/Federated_PyTorch_UNET_Tutorial.ipynb:    "        tags = (\"metric\", suffix)\n",
./openfl-workspace/torch_cnn_histology_gramine_ready/src/pt_cnn.py:        tags = ('metric', suffix)
./openfl-workspace/torch_cnn_histology/src/pt_cnn.py:        tags = ('metric', suffix)
./openfl-workspace/torch_unet_kvasir_gramine_ready/src/fed_unet_runner.py:        tags = ('metric', suffix)
./openfl-workspace/torch_cnn_mnist/src/pt_cnn.py:        tags = ('metric', suffix)
./openfl-workspace/torch_unet_kvasir/src/fed_unet_runner.py:        tags = ('metric', suffix)
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/pipelines/test_tensor_codec.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/databases/test_tensor_db.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/databases/test_tensor_db.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/databases/test_tensor_db.py:    tensor_name, origin, round_number, report, tags = tensor_key
./tests/openfl/component/aggregator/test_aggregator.py:    tags = ['compressed']
```

mansishr:
@alexey-khorkin we are sticking to tuples for all tags, unless they need to be changed,  in that case we convert to list and then back to tuple. The changes I made are to convert type: <class 'google.protobuf.pyext._message.RepeatedScalarContainer'> to tuple.

openfl/transport/grpc/aggregator_client.py
```diff
@@ -277,7 +277,7 @@ def get_aggregated_tensor(self, collaborator_name, tensor_name, round_number,
             tensor_name=tensor_name,
             round_number=round_number,
             report=report,
-            tags=tags,
+            tags=tuple(tags),
```
aleksandr-mokrov: It's redundant converting.

aleksandr-mokrov, line 280: It's redundant converting.

itrushkin:
I wonder if we could use `set` for tags. This could speed up search by tags up to O(1).
/cc @alexey-gruzdev

aleksandr-mokrov:
> I wonder if we could use set for tags. This could speed up search by tags up to O(1).
> /cc @alexey-gruzdev

Set is not hashable, and this is a problem (tags are used as a key in tensor key). And in our case when we have only few tags, or mostly only one, it couldn't really speed up. But it make sense to avoid duplicating and different order of the same tags ((1, 2) and (2, 1) for example). But more on that below.
Also, we discussed, that we can improve current implementation by adding specific function that takes in the tensor tags and update them (also it can reorder and deduplicate them). `new_tags = tuple(list(tags) + ['compressed']` becomes  `new_tags = change_tags(tags, add_field= 'compressed')` for example. It can help to avoid mess with tag types, but we can't make it mandatory to use this function.
Another idea is to add a specific type for tags (like list or set, that has _hash_ method, that returns hash of ordered list converted to tuple;)).

dmitryagapov:
For `tags` we can use class with overloading `__add__` and `__sub__` methods to allow to change it simpler:
```python
tensor_key.tags += 'compressed'
tensor_key.tags -= 'compressed'
'compressed' in tensor_key.tags
```

Also looks like better to use 
```python 
tensor_key.name
tensor_key.origin
tensor_key.round_number
tensor_key.report
tensor_key.tags
```
instead of:
```python
tensor_name, origin, round_number, report, tags = tensor_key
```

mansishr:
@dmitryagapov in my opinion, `collections.namedtuple` class is more pythonic than defining a regular class and the objects are immutable.

aleksandr-mokrov:
In addition to idea of a new Tags class. It can be 
```python
class Tags(set):
    def __hash__(self):
        return hash(tuple(self))
```
and then it will be possible to use it as dictionary key (as a part of TensorKey):
```python
tags = Tags()
tags.add('test1')
tags.update(['test2', 'test3'])

d = {
    tags: '123'
}
```

Also it work fine with TensorDB (pd.DataFrame) without converting (Tags as subclass of list - won't).  Only place where converting will be needed is  getting tags from protobuf message.

mansishr:
@dmitryagapov  @aleksandr-mokrov , the only issue that I see with the above suggestion is that if we convert TensorKey from namedtuple to class, then we won't be able to use them as dictionary keys ( eg: in the code below)

```
    def _load_initial_tensors_from_dict(self, tensor_dict):
       
        tensor_key_dict = {
            TensorKey(k, self.uuid, self.round_number, False, ('model',)):
                v for k, v in tensor_dict.items()
        }
```
Let me know if there is a way around this.

dmitryagapov:
> @dmitryagapov @aleksandr-mokrov , the only issue that I see with the above suggestion is that if we convert TensorKey from namedtuple to class, then we won't be able to use them as dictionary keys ( eg: in the code below)
> 
> ```
>     def _load_initial_tensors_from_dict(self, tensor_dict):
>        
>         tensor_key_dict = {
>             TensorKey(k, self.uuid, self.round_number, False, ('model',)):
>                 v for k, v in tensor_dict.items()
>         }
> ```
> 
> Let me know if there is a way around this.

@mansishr, The only proposal about `TensorKey` is getting attributes like this `tensor_key.name`, because `TensorKey` is already class.

openfl/component/aggregator/aggregator.py
```diff
@@ -782,7 +776,7 @@ def _compute_validation_related_task_metrics(self, task_name):
                 f'Tensor {tensor_key} in task {task_name} has not been processed correctly'
             )
             # Strip the collaborator label, and lookup aggregated tensor
-            new_tags = tuple(tags[:-1])
+            new_tags = change_tags(tags[:-1])
```
dmitryagapov: If we manipulating with tags through `change_tags` if it possible, maybe here we need to do `change_tags(tags, remove_field='collaborator_name')`

openfl/transport/grpc/aggregator_server.py
```diff
@@ -185,7 +186,7 @@ def GetAggregatedTensor(self, request, context):  # NOQA:N802
         require_lossless = request.require_lossless
         round_number = request.round_number
         report = request.report
-        tags = request.tags
+        tags = change_tags(request.tags)
```
dmitryagapov: @aleksandr-mokrov, I don't know is it correct to use `change_tags` without `add_field` and `remove_field`?

dmitryagapov: @aleksandr-mokrov, I don't know is it correct to use `change_tags` without `add_field` and `remove_field`?

openfl/utilities/utils.py
```diff
@@ -236,3 +236,24 @@ def merge_configs(
             settings.set(key, operation(value))
     settings.validators.validate()
     return settings
+
+
+def change_tags(tags, add_field=None, remove_field=None):
```
dmitryagapov: Looks like it will not work correct for `change_tags('model', remove_field='model')`

dmitryagapov (changes requested):

openfl/federated/task/runner_pt.py
```diff
@@ -117,7 +118,7 @@ def validate(self, col_name, round_num, input_tensor_dict,
             suffix += '_local'
         else:
             suffix += '_agg'
-        tags = ('metric', suffix)
+        tags = change_tags(suffix, add_field='metric')
```
aleksandr-mokrov: Lets avoid of tags as string:
```python
tags = ('metric',)
tags = change_tags(tags, add_field=suffix)
```

openfl/federated/task/runner_keras.py
```diff
@@ -230,7 +231,7 @@ def validate(self, col_name, round_num, input_tensor_dict, **kwargs):
             suffix += '_local'
         else:
             suffix += '_agg'
-        tags = ('metric', suffix)
+        tags = change_tags(suffix, add_field='metric')
```
aleksandr-mokrov: Lets avoid of tags as string:
```python
tags = ('metric',)
tags = change_tags(tags, add_field=suffix)
```

openfl/component/aggregator/aggregator.py
```diff
@@ -568,7 +567,7 @@ def _process_named_tensor(self, named_tensor, collaborator_name):
             self.uuid,
             named_tensor.round_number,
             named_tensor.report,
-            tuple(named_tensor.tags)
```
aleksandr-mokrov: I suggest to keep `tuple(named_tensor.tags)` for converting from protobuf.

openfl/component/collaborator/collaborator.py
```diff
@@ -494,7 +495,7 @@ def named_tensor_to_nparray(self, named_tensor):
             self.collaborator_name,
             named_tensor.round_number,
             named_tensor.report,
-            tuple(named_tensor.tags)
+            change_tags(named_tensor.tags)
```
aleksandr-mokrov: I suggest to keep `tuple(named_tensor.tags)` for converting from protobuf.

openfl/transport/grpc/aggregator_server.py
aleksandr-mokrov: I suggest to keep `tags = change_tags(request.tags)` for converting from protobuf.

openfl/utilities/utils.py
aleksandr-mokrov: Let's just leave tuple. We can avoid tags as string and list.

aleksandr-mokrov: @mansishr, can you write test for this function?

openfl/component/aggregator/aggregator.py
mansishr: done

openfl/transport/grpc/aggregator_server.py
mansishr: Changed back to tuple

openfl/utilities/utils.py
mansishr: done

mansishr: Done

openfl/utilities/utils.py
```diff
@@ -236,3 +236,27 @@ def merge_configs(
...
+            tags = list(tags) + [add_field]
+    if remove_field:
+        if type(tags) == str:
+            tags = [tags]
+        else:
+            tags = list(tags)
+        tags.remove(remove_field)
+    # remove duplicates while keeping the order
+    tags = tuple(dict.fromkeys(tags))
+    return tags
```
dmitryagapov:
```suggestion
    if type(tags) == str:
        tags = [tags]
    else:
        tags = list(tags)
    
    if add_field:
        tags.append('add_field')
    if remove_field:
        tags.remove(remove_field)
    # remove duplicates while keeping the order
    tags = tuple(dict.fromkeys(tags))
    return tags
```

openfl/utilities/utils.py
```diff
@@ -236,3 +236,27 @@ def merge_configs(
             settings.set(key, operation(value))
     settings.validators.validate()
     return settings
+
+
+def change_tags(tags, add_field=None, remove_field=None) -> Tuple[str]:
```
itrushkin: We might need to add/remove multiple tags in a single operation. Does replacing strings with lists make sense here?

itrushkin: I like the idea of `Tags` class. It would allow us to store this logic as 2 separate methods of it: `add` and `remove`.

openfl/utilities/utils.py
```diff
@@ -236,3 +236,27 @@ def merge_configs(
...
+    """Change tensor tags to add or remove fields and change tensor type.
+
+    Args:
+        tags(list or tuple): tensor tags.
+        add_field(str): add a new tensor tag field.
+        remove_field(str): remove a tensor tag field.
+    """
+    if add_field:
+        if type(tags) == str:
```
itrushkin: Why do we expect `str`? Could the conversion be executed outside this method, so we operate only with tuples/lists here?

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,26 @@
...
+        ('abc', 'def', None, ('abc', 'def')),
+        (['abc'], 'ghi', None, ('abc', 'ghi')),
+        (['abc', 'def'], 'ghi', None, ('abc', 'def', 'ghi')),
+        (['abc', 'def'], None, 'abc', ('def',)),
+        (['abc', 'def'], 'ghi', 'abc', ('def', 'ghi')),
+        (('abc', 'def'), 'ghi', 'abc', ('def', 'ghi')),
+        (('abc', 'def', 'def'), 'ghi', 'abc', ('def', 'ghi')),
+        (('abc', 'ghi', 'def'), 'ghi', 'abc', ('ghi', 'def')),
+    ])
+def test_equal(tags, add_field, remove_field, expected_result):
```
dmitryagapov: Generally, test function should have name like `test_<tested_function_name>_<test_case>`.

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,26 @@
...
+        (['abc'], None, 'abc', ()),
+        ('abc', 'def', None, ('abc', 'def')),
+        (['abc'], 'ghi', None, ('abc', 'ghi')),
+        (['abc', 'def'], 'ghi', None, ('abc', 'def', 'ghi')),
+        (['abc', 'def'], None, 'abc', ('def',)),
+        (['abc', 'def'], 'ghi', 'abc', ('def', 'ghi')),
+        (('abc', 'def'), 'ghi', 'abc', ('def', 'ghi')),
+        (('abc', 'def', 'def'), 'ghi', 'abc', ('def', 'ghi')),
+        (('abc', 'ghi', 'def'), 'ghi', 'abc', ('ghi', 'def')),
+    ])
```
dmitryagapov: To fully test this function the following cases should be checked:
1. tags: tuple: empty, one element, two elements
2. add_field: None, string that not in tags, string that in tags
3. remove_field: None, string that not in tags, string that in tags
4. add_field == remove_field: None, string that not in tags, string that in tags

Better to create different tests for different cases. Think how to better organize groups for parametrize.

dmitryagapov (changes requested):

openfl/utilities/utils.py
dmitryagapov: I think that `change_tags` should operate only with `tuple`

dmitryagapov (changes requested):

tests/openfl/utilities/test_change_tags.py
mansishr: @dmitryagapov in case remove_field value doesn't exist, we would throw an error instead. Let me know if we still want to add it to test.

openfl/utilities/utils.py
mansishr: Done

openfl/utilities/utils.py
mansishr: Yes but we did not find any good solution, maybe we could convert to class in future and reopen the issue.

tests/openfl/utilities/test_change_tags.py
dmitryagapov: @mansishr, I think we should add test for this any way.

mansishr: Done

openfl/utilities/utils.py
```diff
@@ -236,3 +236,22 @@ def merge_configs(
...
+    """Change tensor tags to add or remove fields.
+
+    Args:
+        tags(tuple): tensor tags.
+        add_field(str): add a new tensor tag field.
+        remove_field(str): remove a tensor tag field.
+    """
+    tags = list(tags)
+
+    if add_field:
```
dmitryagapov: Could we check `add_field is not None` because empty string also valid value for tuple

openfl/utilities/utils.py
```diff
@@ -236,3 +236,22 @@ def merge_configs(
...
+        tags.append(add_field)
+    if remove_field:
```
dmitryagapov: Could we check `remove_field is not None` because empty string also valid value for tuple

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,86 @@
...
+def test_change_tags_without_add_or_remove(tags, expected_result):
+    """Test change_tags without add or remove fields."""
+    result = change_tags(tags)
+    assert result == expected_result
+
+
+@pytest.mark.parametrize(
+    'tags,add_field,expected_result', [
+        (('abc',), None, ('abc',)),
+        (('abc',), '', ('abc',)),
```
dmitryagapov: looks like it should be `(('abc',), '', ('abc', ''))`

openfl/utilities/utils.py
```diff
@@ -236,3 +236,22 @@ def merge_configs(
...
+        tags.remove(remove_field)
```
dmitryagapov: Could we here raise custom expention because it's not obvious that when we delete field from `tuple` raises exception about `list`

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+expected_exception_message = 'list.remove(x): x not in list'
+
+
+@pytest.mark.parametrize(
+    'tags,remove_field,expected_result', [
+        (('abc', 'def'), 'ghi', expected_exception_message),
+    ])
+def test_change_tags_remove_not_in_tags(tags, remove_field, expected_result):
```
dmitryagapov: `expected_result` is not used

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,86 @@
...
+    """Test change_tags with remove field not in tags."""
+    with pytest.raises(ValueError) as excinfo:
+        change_tags(tags, remove_field=remove_field)
+    assert str(excinfo.value) == expected_exception_message
```
dmitryagapov: If we create custom exception assert can be deleted

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,86 @@
...
+        (('abc', 'ghi', 'def'), 'ghi', 'abc', ('ghi', 'def')),
+        (('abc', 'ghi', 'def'), 'ghi', 'ghi', ('abc', 'def', 'ghi')),
+    ])
+def test_change_tags_add_and_remove_both(tags, add_field, remove_field, expected_result):
+    """Test change tags with both add and remove fields."""
+    result = change_tags(tags, add_field=add_field, remove_field=remove_field)
+    assert result == expected_result
+
+
+expected_exception_message = 'list.remove(x): x not in list'
```
dmitryagapov: We can create custom exception and work without this variable

openfl/utilities/utils.py
```diff
@@ -236,3 +236,22 @@ def merge_configs(
...
+
+    Args:
+        tags(tuple): tensor tags.
+        add_field(str): add a new tensor tag field.
+        remove_field(str): remove a tensor tag field.
+    """
+    tags = list(tags)
+
+    if add_field:
+        tags.append(add_field)
```
dmitryagapov: May be we should remove duplicates after adding field too? 
Beacause I expect that 
`change_tags(('a', 'b'), 'c', 'c') == change_tags(('a', 'b', 'c'), 'c', 'c')` 
but 
```python
change_tags(('a', 'b', 'c'), 'c', 'c') == ('a', 'b', 'c')
change_tags(('a', 'b'), 'c', 'c')  == ('a', 'b')
```

tests/openfl/utilities/test_change_tags.py
dmitryagapov: If we have only one case here maybe we can delete parametrize

openfl/utilities/utils.py
```diff
@@ -236,3 +236,22 @@ def merge_configs(
             settings.set(key, operation(value))
     settings.validators.validate()
     return settings
+
+
+def change_tags(tags, add_field=None, remove_field=None) -> Tuple[str]:
```
dmitryagapov:
```suggestion
def change_tags(tags, *, add_field=None, remove_field=None) -> Tuple[str]:
```
Looks like we expect that we shouldn't use add_field and remove_field as positional arguments.

dmitryagapov (changes requested):

openfl/utilities/utils.py
```diff
@@ -236,3 +236,22 @@ def merge_configs(
...
+        remove_field(str): remove a tensor tag field.
+    """
+    tags = list(tags)
+
+    if add_field:
+        tags.append(add_field)
+    if remove_field:
+        tags.remove(remove_field)
+    # remove duplicates while keeping the order
+    tags = tuple(dict.fromkeys(tags))
```
aleksandr-mokrov: Should we keep order? Are tags ('a', 'b') and ('b', 'a') different? Probably we should use the same sorting algorithm instead keeping order, for example `tags = tuple(tags.sort()).

openfl/component/aggregator/aggregator.py
```diff
@@ -328,7 +329,7 @@ def get_aggregated_tensor(self, collaborator_name, tensor_name,
             require_lossless: bool
             round_number: int
             report: bool
-            tags: list[str]
+            tags: tuple[str]
```
dmitryagapov:
```suggestion
            tags: tuple[str, ...]
```

openfl/utilities/utils.py
```diff
@@ -236,3 +236,25 @@ def merge_configs(
             settings.set(key, operation(value))
     settings.validators.validate()
     return settings
+
+
+def change_tags(tags, *, add_field=None, remove_field=None) -> Tuple[str]:
```
dmitryagapov:
```suggestion
def change_tags(tags, *, add_field=None, remove_field=None) -> Tuple[str, ...]:
```

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,86 @@
...
+    result = change_tags(tags)
+    assert result == expected_result
+
+
+@pytest.mark.parametrize(
+    'tags,add_field,expected_result', [
+        (('abc',), None, ('abc',)),
+        (('abc',), '', ('', 'abc')),
+        (('abc', 'def'), 'ghi', ('abc', 'def', 'ghi')),
+        (('def', 'abc'), 'ghi', ('abc', 'def', 'ghi')),
```
dmitryagapov: Maybe this test is unnecessary.

tests/openfl/utilities/test_change_tags.py
```diff
@@ -0,0 +1,86 @@
...
+    """Test change_tags with add field."""
+    result = change_tags(tags, add_field=add_field)
+    assert result == expected_result
+
+
+@pytest.mark.parametrize(
+    'tags,remove_field,expected_result', [
+        (('abc',), None, ('abc',)),
+        (('abc', 'def'), 'def', ('abc',)),
+        (('abc', 'def', 'def'), 'def', ('abc',)),
```
dmitryagapov: Maybe case with duplicated keys should be in separate test

dmitryagapov (changes requested):

aleksandr-mokrov (approved):

mansishr:
Jenkins please retry a build

dmitryagapov (approved):

alexey-gruzdev (approved):

PR #389, opened 2022-03-18
Corrupt shard utils and experiments

igor-davidyuk:
This PR introduces Corrupting Shard Descriptor wrapper.

  Decorating a Shard Descriptor (SD) class with this function will result in the following:
  1. `corrupt` parameter is added to the SD init function.
      The Envoy manager may enable corruption by putting `corrupt: true` to SD params.
  2. `spoil_method` passed to the wrapper will be used to spoil Shard Datasets of this Envoy
      if corruption is enabled.


mansishr:
I think adding a separate folder with 'workspace_corruption_experiment' (which has a different format) may confuse the user and they might think that we can't use corrupt shard wrapper with regular Jupyter notebook (unless we show the accuracy difference in experiments between corrupt and non corrupt collaborators)

theakshaypant:

PR #400, opened 2022-03-25
Choose a port for aggregator server

### LINKED ISSUE #342: Introduce a deterministic algorithm for choosing a port for the aggregator server.
The workflow will probably be as follows:
Director config will feature a new parameter `open_ports` listing one or more ports.
This list will be tracked by the Director-service, which will provide one of the ports to the Aggregator at the start time.

mansishr:
Fixes #342

### DIFF
--- a/openfl/component/director/director.py
+++ b/openfl/component/director/director.py
@@ -6,6 +6,7 @@
 import asyncio
 import itertools
 import logging
+import socket
 import time
 from collections import defaultdict
 from pathlib import Path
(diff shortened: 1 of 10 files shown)


openfl/component/director/director.py
```diff
@@ -74,16 +91,21 @@ async def set_new_experiment(
...
+                archive_path=experiment_archive_path,
+                collaborators=list(collaborator_names),
+                users=[sender_name],
+                sender=sender_name,
+                init_tensor_dict=tensor_dict,
+                agg_port=open_port,
+            )
+            self.experiments_registry.add(experiment)
+            return True
+        logger.info(f'Could not find open ports out of the list: {open_ports_list}')
```
psfoley: This should be changed. In the case that all user specified ports are in use, this would result in the data scientist not being able to submit their experiment.

theakshaypant:
@mansishr can you please resolve the conflicts and request for review again?

Issue #375, opened 2022-03-11
Circular import issue


alexey-khorkin:
Circular import error when starting director. The problem occurs when we will annotate the `model_interface` parammeter inside `NumPyAdagrad` and then start director (any interactive-api example could be used)

```python3
from openfl.interface.interactive_api.experiment import ModelInterface
...
class NumPyAdagrad(Optimizer):
    """Adagrad optimizer implementation.

    Original paper: http://jmlr.org/papers/v12/duchi11a.html
    """

    def __init__(
        self,
        *,
        params: Optional[Dict[str, np.ndarray]] = None,
        model_interface: Optional[ModelInterface] = None,     # <---- NEW ANNOTATION HERE
        learning_rate: float = 0.01,
        initial_accumulator_value: float = 0.1,
        epsilon: float = 1e-10,
    ) -> None:
``` 

The point here is that this interface class must have the ability to be imported anywhere, as it is an interface and can be used for annotations. 

**To Reproduce**
1. Сheckout on `circular_import_issue` branch
2. Install `openfl`:
```bash
pip install .
```
4. Go to MNIST example to start director:
```bash
cd openfl-tutorials/interactive_api/Tensorflow_MNIST/director/
./start_director.sh
```
7. See error:
```bash
./start_director.sh 
Traceback (most recent call last):
  File "/home/akhorkin/Repositories/alexey-khorkin-repos/test/openfl/venv/bin/fx", line 11, in <module>
    load_entry_point('openfl==1.2.1', 'console_scripts', 'fx')()
  File "/home/akhorkin/Repositories/alexey-khorkin-repos/test/openfl/venv/lib/python3.8/site-packages/openfl/interface/cli.py", line 207, in entry
    command_group = import_module(module, package)
  File "/usr/local/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
[...]

PR #404, opened 2022-04-04
Adaptive Optimizers Fix

alexey-khorkin:
Fixes for #375 
Rewrote optimizers to not contain parameters, but only their statistics. 
Fixed documentation, Tensorflow MNIST example, AdaptiveAggregation `call` method as well.


theakshaypant:
@alexey-khorkin  can you please resolve the conflicts and request for review again?

PR #426, opened 2022-05-05
add CI test for Tensorflow example

mansishr:


openfl-workspace/keras_cnn_mnist_new/plan/cols.yaml
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
shreyasnc1, line 1: 2022 here and other files as well?

psfoley:
@mansishr closing this for now - we can reevaluate if this test is needed in 1.6

PR #433, opened 2022-05-18
Fix flatten function for nested objects

mansishr:
`flatten_json.flatten_preserve_lists()` isn't smashing hierarchies beyond third nested level. The fix is to use `flatten_json.flatten()` instead and preserve lists while unflatting using `flatten_json.unflatten_list()`

### DIFF
--- a/openfl/native/native.py
+++ b/openfl/native/native.py
@@ -11,7 +11,7 @@ from copy import copy
 from logging import getLogger
 from pathlib import Path
 
-from flatten_json import flatten_preserve_lists
+import flatten_json
 
 import openfl.interface.aggregator as aggregator
 import openfl.interface.collaborator as collaborator
@@ -54,7 +54,7 @@ def setup_plan(log_level='CRITICAL'):
 
 def flatten(config, return_complete=False):
     """Flatten nested config."""
-    flattened_config = flatten_preserve_lists(config, '.')[0]
+    flattened_config = flatten_json.flatten(config, '.')
     if not return_complete:
         keys_to_remove = [
             k for k, v in flattened_config.items()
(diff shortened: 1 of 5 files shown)


mansishr:

psfoley:
Thank you for fixing this, @mansishr! I think once some tests for this are in place, this PR is ready to merge.

openfl/native/native.py
```diff
@@ -70,11 +70,15 @@ def flatten(config, return_complete=False):
...
     Update the plan with the provided override and save it to disk.
+    If the override value is a list, keys corresponding to each list
+    value should be specified.
 
     For a list of available override options, call `fx.get_plan()`
 
     Args:
-        override_config : dict {"COMPONENT.settings.variable" : value}
+        override_config : dict {"COMPONENT.settings.variable" : value,
+                                "COMPONENT.settings.list_variable.0" : list_value_0,
```
psfoley: @mansishr does this mean that list items must be set individually? Or can a full list be provided here? i.e.

```
"COMPONENT.settings.list_variable" : [1,2,3]
```

mansishr: List items would need to be set individually for this to work

`{"COMPONENT.settings.list_variable.0" : 1, "COMPONENT.settings.list_variable.1" : 2, "COMPONENT.settings.list_variable.2" : 3}`

psfoley: This is not ideal for long lists. I would recommend adding some logic in the update plan method to do this transformation programmatically:

```
final_override_config = {}
for key, val in override_config.items():
    if type(val) == list:
        for idx,v in enumerate(val):
            final_override_config[f'{key}.{idx}'] = v
    else:
        final_override_config[key] = val
```

psfoley:
LGTM!

psfoley:
LGTM!

PR #435, opened 2022-05-24
Add TF version check before invoking keras model serialization

mansishr:
Adding TF version check before invoking hotfix function for keras model serialization.


openfl/plugins/frameworks_adapters/keras_adapter.py
```diff
@@ -45,7 +47,9 @@ def __reduce__(self):  # NOQA:N807
             cls.__reduce__ = __reduce__
 
         # Run the function
-        make_keras_picklable()
+        tf_version = float((tf.__version__).split(".")[0] + "." + (tf.__version__).split(".")[1])
```
MasterSkepticista: `tf.__version__` supports direct string boolean checks. We could simplify this to

```python
if tf.__version__ < '2.7.1':
  ...
```

MasterSkepticista: @mansishr A few minor corrections on the MNIST experiment (from the previous PR, couldn't add that into the previous one). Will push the commits here before this gets merged. Is that OK?

mansishr:
Looks good to me!

PR #441, opened 2022-06-08
Straggler Handling Interface

mansishr:
This PR identifies stragglers for each task and excludes them from sending task results to aggregator. The two policies created are:

1. Cutoff Time based straggler handling: Straggler is not able to finish tasks for a round within a given cutoff time
2. Percentage based straggler handling: Based on %age, there are enough collaborators with task results to proceed for next round.


openfl/component/aggregator/aggregator.py
```diff
@@ -506,11 +527,12 @@ def send_local_task_results(self, collaborator_name, round_number, task_name,
         # task dictionary
         for named_tensor in named_tensors:
             # sanity check that this tensor has been updated
-            if named_tensor.round_number != round_number:
-                raise ValueError(
-                    f'Collaborator {collaborator_name} is reporting results for the wrong round.'
-                    f' Exiting...'
-                )
+            # if named_tensor.round_number != round_number:
```
psfoley: This block looks like it can be removed

psfoley (changes requested):
Thanks Mansi for adding this interface.

A couple requests:

1. Could you add a new OpenFL workspace template that demonstrates how select a different straggler handling policy from the plan file? Also, I would expect to see changes within the plan.py file to initialize the referenced straggler policy. 
2. You should also add documentation about how to implement a new straggler handling policy.

psfoley (approved):

openfl/component/aggregator/aggregator.py
```diff
@@ -868,21 +887,37 @@ def _end_of_round_check(self):
...
-            ) for c in collaborators_needed
-        ])
+        
+        collaborators_done = []
+        for c in all_collaborators:
+            if self._collaborator_task_completed(
+                c, task_name, self.round_number):
+                collaborators_done.append(c)
+        
+        straggler_check = self.straggler_handling_policy.straggler_cutoff_check(
```
igor-davidyuk, line 900: is it a hardcoded method?

openfl-workspace/torch_cnn_mnist_straggler_check/plan/plan.yaml
```diff
@@ -0,0 +1,51 @@
...
+assigner :
+  defaults : plan/defaults/assigner.yaml
+  
+tasks :
+  defaults : plan/defaults/tasks_torch.yaml
+
+compression_pipeline :
+  defaults : plan/defaults/compression_pipeline.yaml
+
+straggler_handling_policy :
```
igor-davidyuk, line 47: There should be a default template that would be copied if user ignored this section

PR #452, opened 2022-07-05
Support plan updates for interactive API

### LINKED ISSUE #427: Support configuration of compression_pipeline by Director-Based Workflow
I'm trying to deploy and compare the different compression pipelines supported by the project.
But, as it seems, the configuration of the compression_pipline is not supported by the director-based workflow.

mansishr:
Fixes #427 
Plan settings can now be modified for the director based flow using override config.

### DIFF
--- a/docs/overriding_plan_settings.rst
+++ b/docs/overriding_plan_settings.rst
@@ -1,4 +1,4 @@
-.. # Copyright (C) 2020-2021 Intel Corporation
+.. # Copyright (C) 2020-2022 Intel Corporation
 .. # SPDX-License-Identifier: Apache-2.0
 
 .. _overriding_plan_settings:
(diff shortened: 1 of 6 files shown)


docs/overriding_plan_settings.rst
```diff
@@ -0,0 +1,44 @@
...
+
+
+Director Envoy Based Interactive API Interface
+==============================================
+Use **override_config** with FL_experiment.start:
+
+.. code-block:: python
+
+    import openfl.native as fx
+    print(fx.get_plan(plan_config_path='plan/plan.yaml'))
```
psfoley: I think it would be helpful to actually show the output of the `fx.get_plan` function call here. Then add a section that explains the key, value in the plan being overwritten, and then show how they provide these in the override config of the `fl_experiment.start` command

psfoley (changes requested):
Once a small modification to the documentation is addressed - this can be merged!

docs/overriding_plan_settings.rst
```diff
@@ -0,0 +1,97 @@
...
+                   override_config={'aggregator.settings.db_store_rounds': 1, 'compression_pipeline.template': 'openfl.pipelines.KCPipeline', 'compression_pipeline.settings.n_clusters': 2})
+
+
+Since 'aggregator.settings.db_store_rounds' is a field already present in the plan, its value gets replaced with the new value specified. Fields  'compression_pipeline.template' and 'compression_pipeline.settings.n_clusters' are new enteries that get added to the plan:
+
+.. code-block:: python
+
+    INFO     Updating aggregator.settings.db_store_rounds to 1...                                                                           native.py:102
+
+    WARNING  Did not find compression_pipeline.template in config. Make sure it should exist. Creating...                                   native.py:105
```
psfoley: Thanks Mansi for the extended documentation. The base plan ([here](https://github.com/intel/openfl/blob/develop/openfl-workspace/workspace/plan/plans/default/base_plan_interactive_api.yaml)) should be modified to include the default compression pipeline information (https://github.com/intel/openfl/blob/a018c4423c4d2ef5a929406912373a47e95c4183/openfl-workspace/default/plan/plan.yaml#L35-L36)

That will make it more clear for this example (and any compression pipeline) which values in the plan to modify.

We should also change the warning to an info message, because it will be somewhat common for custom fields to be passed into the plan

openfl-tutorials/interactive_api/Tensorflow_MNIST/workspace/Tensorflow_MNIST.ipynb
```diff
@@ -407,6 +407,18 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "480700cc-400b-4073-9daf-71b1f0d60d62",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# print the default federated learning plan\n",
+    "import openfl.native as fx\n",
+    "print(fx.get_plan(plan_config_path='plan/plan.yaml'))"
```
igor-davidyuk: We mix Interactive API with 'native' API here, which is confusing even for me

openfl/interface/interactive_api/experiment.py
```diff
@@ -62,6 +63,22 @@ def __init__(
         self.logger = getLogger(__name__)
         setup_logging()
 
+        self._initialize_plan()
```
igor-davidyuk, line 66: Why do we need an initialized plan at an Experiment creation time?

openfl-tutorials/interactive_api/Tensorflow_MNIST/workspace/Tensorflow_MNIST.ipynb
mansishr: Hi @igor-davidyuk, the idea behind it is that users should have the flexibility to make any changes to the plan before the experiment starts. Right now, only a few parameters can be modified while starting the experiment. Many other fields, like compression pipeline, can't be modified by the users directly through the jupyter notebook. This PR initializes the plan when the FLExperiment object is created and then users can display the plan settings to see the default values. Accordingly, changes could be made to the plan config using the override_config option.

openfl/interface/interactive_api/experiment.py
psfoley: The main reason is so that user know what the plan parameters are ahead of time (viewable with fx.get_plan) and can then override them at runtime.

openfl-tutorials/interactive_api/Tensorflow_MNIST/workspace/Tensorflow_MNIST.ipynb
igor-davidyuk: I understand the intention, I just say we should not mix different APIs. 
We actually did not address this comment

PR #453, opened 2022-07-06
Move `aggregation_functions` to the Interface package

igor-davidyuk:
I moved the `aggregation_functions` package from the `components` to the `interface` package; this move makes sense generally but also removes the `aggregator -> component -> aggregator` import cycle.
Tested on Kvasir interactive API example.
A python script `from openfl.databases import tensor_db` also works now.
A Pydeps graph for this branch is attached (`pydeps --no-show -o openfl1.svg --show-cycles openfl`)
openfl1


mansishr (approved):
LGTM!

PR #477, opened 2022-08-05
Update README.md

mkurisoo:
Several updates: Commands were changed to represent actual directory names (directory names listed appeared outdated); more steps were added for clarity; wording and formatting was changed in some areas.

### DIFF
--- a/openfl-tutorials/interactive_api/Tensorflow_MNIST/README.md
+++ b/openfl-tutorials/interactive_api/Tensorflow_MNIST/README.md
@@ -2,43 +2,62 @@
 
 ![mnist digits](http://i.ytimg.com/vi/0QI3xgXuB-Q/hqdefault.jpg "MNIST Digits")
 
-### 1. About dataset
+## I. About the dataset
 
 It is a dataset of 60,000 small square 28×28 pixel grayscale images of handwritten single digits
 between 0 and 9. More info at [wiki](https://en.wikipedia.org/wiki/MNIST_database).
 
-### 2. About model
+## II. About the model
 
-We use simple fully-connected neural network defined at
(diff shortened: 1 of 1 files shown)


mkurisoo:

openfl-tutorials/interactive_api/Tensorflow_MNIST/README.md
```diff
@@ -2,43 +2,67 @@
...
+  - For help with this step, visit the "Install the Package" section of the [OpenFL installation instructions](https://openfl.readthedocs.io/en/latest/install.html#install-the-package).
+<br/>
+ 
+### 1. Split terminal into 3 (1 terminal for the director, 1 for the envoy, and 1 for the experiment) and do the following in each terminal:
+   - Activate the virtual environment from step 0:
+   
+   ```sh
+   source venv/bin/activate
+   ```
+   - Ensure proxies are set. If you do not have sudo permissions on your machine, ensure your proxies are set in all terminals.
```
psfoley: Proxies are rarely necessary for users. Would recommend changing this wording to "If you are in a network environment with a proxy, ensure proxy environment variables are set in each of your terminals.". Would also recommend removing the 2nd sentence because sudo permissions are not required for this, and could be confusing.

mkurisoo: Thank you for the suggestion! I agree that the wording you suggested sounds less confusing. The only reason I mentioned sudo permissions was because when I was learning how to set up the proxies, a teammate mentioned that if I had sudo permissions I could set it up so that I wouldn't have to consistently type the proxy info into every terminal. I can see how adding that may confuse users though. I will revise it.

PR #482, opened 2022-08-11
Flax CIFAR10 CNN Interactive API Example

sun1lach:
- FLAX Interactive API example - jupyter notebook with documentation.
- New FLAX framework adapter introduced in `openfl.plugins`.
- Tensorflow (tfds) API based data loader and pre-processing in shard descriptor.
- Introduced flags to enable JAX compute on CPU and control XLA and TF GPU memory allocation.
- `Optax` gradient processing and optimization library and implemented sub-classing of `flax.training.train_state.TrainState` to process stateless parameters updates and access from `training` and `validation` loop resp.
- Updated README.md

### DIFF
--- a/openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
+++ b/openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
@@ -15,38 +15,38 @@ Definition provided in the notebook.
 ### 3. How to run this tutorial (without TLS and locally as a simulation):
 
 0. Pre-requisites:
+    
+    - Nvidia Driver >= 495.29.05
+    - CUDA >= 11.1.105
+    - cuDNN >= 8
+    
+    Activate virtual environment (Python - 3.8.10) and install packages from requirements.txt
 
-    Activate virtual environment (Python - 3.8.10) and install below packages.
-
-```
-pip install openfl==1.3 protobuf==3.19.4 --no-cache-dir
+```sh
+cd Flax_CNN_CIFAR
+pip install -r requirements.txt
 ```
 
 1. Run director:
 
 ```sh
-cd director_folder
+cd director
 ./start_director.sh
 ```
 
 2. Run envoy:
 
-Step 1: Activate virtual environment and install packages from the requirements.txt
-```
-cd envoy_folder
-pip install -r requirements.txt --no-cache-dir
-```
-Step 2: start the envoy
 ```sh
-./start_envoy.sh env_instance_1 envoy_config_1.yaml
+cd envoy
+./start_envoy.sh "envoy_identifier" envoy_config.yaml
 ```
 
 Optional: start second envoy:
 
-- Copy `envoy_folder` to another place and follow the same process as above:
+- Copy `envoy` folder to another place and follow the same process as above:
 
 ```sh
-./start_envoy.sh env_instance_2 envoy_config_2.yaml
+./start_envoy.sh "envoy_identifier_2" envoy_config_2.yaml
 ```
 
 3. Run `FLAX_CIFAR10_CNN.ipynb` jupyter notebook:
--- a/openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
+++ b/openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
@@ -1,0 +1,87 @@
+# Federated FLAX CIFAR-10 CNN Tutorial
+
+### 1. About dataset
+
+The CIFAR-10 dataset consists of 60000 32x32 colour images in 10 classes, with 6000 images per class. There are 50000 training images and 10000 test images.
+
+Define the below param in envoy.yaml config to shard the dataset across participants/envoy.
+- rank_worldsize
+
+### 2. About model
+
+A simple multi-layer CNN is used with XLA compiled and Auto-grad based parameter updates.
+Definition provided in the notebook.
+
+### 3. How to run this tutorial (without TLS and locally as a simulation):
+
+0. Pre-requisites:
+
+    Activate virtual environment (Python - 3.8.10) and install below packages.
+
+```
+pip install openfl==1.3 protobuf==3.19.4 --no-cache-dir
+```
+
+1. Run director:
+
+```sh
+cd director_folder
+./start_director.sh
+```
+
+2. Run envoy:
+
+Step 1: Activate virtual environment and install packages from the requirements.txt
+```
+cd envoy_folder
+pip install -r requirements.txt --no-cache-dir
+```
+Step 2: start the envoy
+```sh
+./start_envoy.sh env_instance_1 envoy_config_1.yaml
+```
+
+Optional: start second envoy:
+
+- Copy `envoy_folder` to another place and follow the same process as above:
+
+```sh
+./start_envoy.sh env_instance_2 envoy_config_2.yaml
+```
+
+3. Run `FLAX_CIFAR10_CNN.ipynb` jupyter notebook:
+
+```sh
+cd workspace
+jupyter lab FLAX_CIFAR10_CNN.ipynb
+```
+
+4. Visualization:
+
+```
+tensorboard --logdir logs/
+```
+
+
+### 4. Known issues
+
+1. #### CUDA_ERROR_OUT_OF_MEMORY Exception - JAX XLA pre-allocates 90% of the GPU at start
+
+- set XLA_PYTHON_CLIENT_PREALLOCATE to start with a small memory footprint.
+```
+%env XLA_PYTHON_CLIENT_PREALLOCATE=false
+```
+OR
+
+- Below flag to restrict max GPU allocation to 50%
+```
+%env XLA_PYTHON_CLIENT_MEM_FRACTION=.5
+```
+
+
+2. #### Tensorflow pre-allocates 90% of the GPU (Potential OOM Errors).
+
+- set TF_FORCE_GPU_ALLOW_GROWTH to start with a small memory footprint.
+```
+%env TF_FORCE_GPU_ALLOW_GROWTH=true
+```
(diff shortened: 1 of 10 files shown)


openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
```diff
@@ -0,0 +1,87 @@
...
+Definition provided in the notebook.
+
+### 3. How to run this tutorial (without TLS and locally as a simulation):
+
+0. Pre-requisites:
+
+    Activate virtual environment (Python - 3.8.10) and install below packages.
+
+```
+pip install openfl==1.3 protobuf==3.19.4 --no-cache-dir
```
MasterSkepticista: Can this be merged into `pip install -r requirements.txt`? Additionally, my sense is that requirements should be unique to an example (and therefore should be in the root directory of the example, not within envoy/director dirs).

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
```diff
@@ -0,0 +1,87 @@
...
+2. Run envoy:
+
+Step 1: Activate virtual environment and install packages from the requirements.txt
+```
+cd envoy_folder
+pip install -r requirements.txt --no-cache-dir
+```
+Step 2: start the envoy
+```sh
+./start_envoy.sh env_instance_1 envoy_config_1.yaml
```
MasterSkepticista: General tip: If `env_instance_x` is a name, name it as `Bob`, `Alice` etc. So that the reader understands it is not a flag or an argument that affects the execution. 

We've not been doing this on older examples (even I didn't for that matter). But I will make that change in my previous example.

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+Optional: start second envoy:
+
+- Copy `envoy_folder` to another place and follow the same process as above:
```
MasterSkepticista: Copy `envoy`

instead of 

Copy `envoy_folder`

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+- set XLA_PYTHON_CLIENT_PREALLOCATE to start with a small memory footprint.
+```
+%env XLA_PYTHON_CLIENT_PREALLOCATE=false
+```
+OR
+
+- Below flag to restrict max GPU allocation to 50%
+```
+%env XLA_PYTHON_CLIENT_MEM_FRACTION=.5
```
MasterSkepticista, line 90: Is this needed? I believe setting vars in the notebook worked?

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/envoy/cifar10_shard_descriptor.py
```diff
@@ -0,0 +1,109 @@
...
+        train_shard_size = int(len(datasets['train']) / worldsize)
+        test_shard_size = int(len(datasets['test']) / worldsize)
+
+        self.train_segment = f'train[{train_shard_size * (rank - 1)}:{train_shard_size * rank}]'
+        self.test_segment = f'test[{test_shard_size * (rank - 1)}:{test_shard_size * rank}]'
+
+        train_ds = tfds.as_numpy(dataset_builder.as_dataset(split=self.train_segment, batch_size=-1))
+        test_ds = tfds.as_numpy(dataset_builder.as_dataset(split=self.test_segment, batch_size=-1))
+
+        train_ds['image'] = jnp.float32(train_ds['image']) / 255.
```
MasterSkepticista, line 75: Is loading tensors in memory necessary? I would prefer an implementation where your example can plug in with any dataset (say `tf.data`) with a final conversion stage to `jnp`.

I don't know if it is possible with `ds.map(...)`. What is your observation?

MasterSkepticista (changes requested):

openfl/plugins/frameworks_adapters/flax_adapter.py
```diff
@@ -0,0 +1,100 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Custom model DeviceArray - JAX Numpy adapter."""
+import jax
+import jax.numpy as jnp
+import numpy as np
+import optax
+
+from .framework_adapter_interface import FrameworkAdapterPluginInterface
+
+_DELIM = '.'
```
MasterSkepticista: If this delimiter is the same for models in Jax/Flax, no need to declare as private global.

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
sun1lach: Would it be better to enclose `env_instance_x` in double quotes `"env_instance_x"` to let user know that it's a user defined string passed as a parameter? A generic naming like  `Bob, Alice` might take it a bit out of program's intent and probably naming convention for envoy instance. Let me know your thoughts on this.

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
sun1lach: `fx envoy start...` initializes shard descriptor and custom framework adapter at start and `jax, jaxlib` imports causes the GPU allocation upfront.
Same is the case when a new `train_state` gets created from the notebook.
Basically, 2 process gets spawned that utilized GPU memory and hence its essential to have environment variable set while running `envoy.sh` as well as in the notebook. 
./start_envoy.sh -> this takes care of setting the required env vars for the envoy automatically.

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/envoy/cifar10_shard_descriptor.py
sun1lach: If I interpreted it correctly, did you mean decouple the dataloader part from JAX/Flax DeviceArray representation?
This is doable, as JAX provides treemap conversion to transform and map any data representation to DeviceArray.

openfl/plugins/frameworks_adapters/flax_adapter.py
sun1lach: Used this delimiter to flatten the deeply nested tree of layers, kernel and bias.
Helpful to flatten and unflatten from `get_tensor_dict()` and `set_tensor_dict()` methods respectively (Basically DeviceArray to numpy and vice versa).

Another alternative was to use `flattendict` library. But avoided this import by having a delimiter.

MasterSkepticista: Having a delimiter is OK, but if it is well-known, need not declare it as a constant, is what I meant...

sun1lach: Sure (y)
I also have an idea to simplify the `get_tensor_dict` and `set_tensor_dict` logic as a whole. This could eliminate the need to have any delimiters. will try and update in the next commit.

sun1lach: Refactored this part. 
Improved the logic that handles multiple optax optimizer states.

sun1lach:
[Commit ](https://github.com/intel/openfl/pull/482/commits/21f561fedd92001d3f113bb3778210f79d72327e)

Refactored Flax adapter to support all optax optimizers and their state updates in `get_tensor_dict` and `set_tensor_dict`. Validated with adam, SGD, rmsprop, adagrad, yogi and adafactor optimizers.
Bumped `Tensorflow 2.8.2 -> 2.9.1`, `Flax 0.4.1 -> 0.5.6`, `tensorflow-datasets 4.4.0 -> 4.6.0`
Updated README.md

openfl-tutorials/interactive_api/Flax_CNN_CIFAR/README.md
```diff
@@ -0,0 +1,104 @@
...
+    - Variable `opt_vars` to keep track of generic optimizer variables.
+    - Method `update_state` to update the OpenFL `ModelInterface` registered state with the new_state returned within the `TaskInterface` registered training loop.
+
+2. Method `create_train_state`: Creates a new `TrainState` by encapsulating model layer definitions, random model parameters, and optax optimizer state.
+
+3. Method `apply_model` (`@jax.jit` decorated function): It takes a TrainState, images, and labels as parameters. It computes and returns the gradients, loss, and accuracy. These gradients are applied to a given state in the `update_model` method (`@jax.jit` decorated function) and a new TrainState instance is returned.
+
+### 4. How to run this tutorial (without TLS and locally as a simulation):
+
+0. Pre-requisites:
```
psfoley, line 27: Can this tutorial be modified to run on CPU as well?

sun1lach: Yes Patrick, just by setting this one variable **DEFAULT_DEVICE='CPU|GPU'** in `start_envoy.sh` and in workspace notebook.

**Details:**

The `jaxlib==0.3.10+cuda11.cudnn805` library version used in `requirements.txt` is by default CUDA device compatible and takes precedence during execution. If no CUDA device is present then the JAX execution falls back to CPU.

Another way we can enforce the execution on any platform is by using the environment variable `JAX_PLATFORMS='CPU|GPU|TPU'` (Updated README.md). Defined in both `start_envoy.sh` and workspace notebook.

This `JAX_PLATFORMS` environment variable need to be set prior to running `./start_envoy.sh` and notebook because the `import jax` statement allocates the GPU memory upfront so does the TensorFlow.

There are other supporting environment variables that goes along with `JAX_PLATFORMS` for the execution to work as expected without taking up GPU memory at its entirety, I am abstracting those away and user would now have to just set one variable `DEFAULT_DEVICE` in `start_envoy.sh` and workspace notebook to control the execution platform.

psfoley (approved):

PR #489, opened 2022-08-23
Envoys to review plan before experiment starts

mansishr:
Adding logic for the Director and Envoys to review the received FL experiment plan before starting the experiment. Set 'review_experiment: True' under 'params' in your director and envoy configs to enable review.
If Director declines an experiment, it is aborted right away, no aggregator is started and Envoys do not receive this experiment archive at all.
If any of the Envoys rejects the experiment, a set_experiment_failed request is sent to the director to stop the aggregator.

### DIFF
--- a/openfl/component/director/experiment.py
+++ b/openfl/component/director/experiment.py
@@ -7,6 +7,7 @@ import asyncio
 import logging
 from contextlib import asynccontextmanager
 from pathlib import Path
+from typing import Callable
 from typing import Iterable
 from typing import List
 from typing import Union
@@ -25,6 +26,7 @@ class Status:
     FINISHED = 'finished'
     IN_PROGRESS = 'in_progress'
     FAILED = 'failed'
+    REJECTED = 'rejected'
 
 
 class Experiment:
@@ -42,18 +44,15 @@ class Experiment:
     ) -> None:
         """Initialize an experiment object."""
         self.name = name
-        if isinstance(archive_path, str):
-            archive_path = Path(archive_path)
-        self.archive_path = archive_path
+        self.archive_path = Path(archive_path).absolute()
         self.collaborators = collaborators
         self.sender = sender
         self.init_tensor_dict = init_tensor_dict
-        if isinstance(plan_path, str):
-            plan_path = Path(plan_path)
-        self.plan_path = plan_path
+        self.plan_path = Path(plan_path)
         self.users = set() if users is None else set(users)
         self.status = Status.PENDING
         self.aggregator = None
+        self.run_aggregator_atask = None
 
     async def start(
             self, *,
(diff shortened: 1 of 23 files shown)


openfl/component/envoy/envoy.py
```diff
@@ -13,7 +13,9 @@
 from typing import Type
 from typing import Union
 
+from click import confirm
 from click import echo
+from click import open_file
```
igor-davidyuk: We better keep click out of this component and close to interface package

openfl/component/envoy/envoy.py
```diff
@@ -80,14 +85,35 @@ def run(self):
...
+                        echo('################ PLAN END ################')
+                        if confirm('Do you want to accept the experiment?'):
+                            echo('Experiment accepted by Collaborator!')
+                        else:
+                            raise Exception('Experiment rejected by Collaborator')
+                            self.director_client.set_experiment_failed(
+                                experiment_name,
+                                error_code=2,
+                                error_description='Collaborator rejected the experiment'
+                            )
```
igor-davidyuk: This user interaction code should be placed in a separate module. The first idea that came to my mind - put it somewhere in cli package and import this check from there

igor-davidyuk (changes requested):

psfoley (changes requested):
@mansishr Thanks for this PR. Plan agreement should implemented for the director as well (because the user of the frontend and director admin are not guaranteed to be the same.

openfl/component/director/experiment.py
```diff
@@ -65,10 +68,14 @@ async def start(
...
         try:
-            logger.info(f'New experiment {self.name} for '
-                        f'collaborators {self.collaborators}')
-
             with ExperimentWorkspace(self.name, self.archive_path):
+                if self.review_experiment:
+                    # Director to review the experiment before starting
+                    if not accept_file_post_review('plan', self.plan_path):
+                        self.status = Status.FAILED
+                        raise Exception('Experiment rejected by Director')
```
igor-davidyuk: Will this exception stop the Director service at all? We should not stop all operations because of one rejected experiment.
Upd: Ok, I see it is located under `try`

igor-davidyuk: it would be a reasonable move to ask Director for approval first, and only after that send the experiments to Envoys

igor-davidyuk:
Could you also add a section describing this feature to the docs?
And use it in one of the examples, may be just create an additional set of director/envoy configs to demonstrate

igor-davidyuk:
This is interesting, I tested the PR and saw that Envoys do not ask for review until the Director votes (in case all of the have the review feature on)
and when accepted, I got the following error:
[accepted.log](https://github.com/intel/openfl/files/9575768/accepted.log)

Another run: I declined an experiment on director and got these errors
[declined_director.log](https://github.com/intel/openfl/files/9575795/declined_director.log)
[declined_on_collaborator.log](https://github.com/intel/openfl/files/9575796/declined_on_collaborator.log)
+ processes hanged.

igor-davidyuk:
I conducted the following experiment:
1. Sent an experiment to Director
2. Accepted it as Director
3. Declined it as Envoy
4. Sent another experiment

Result: The aggregator stopped which is good, but the new experiment never started.
The first experiment's status was still `running`

openfl/component/director/experiment.py
```diff
@@ -85,6 +91,19 @@ async def start(
...
+        #     tls=tls,
+        #     root_certificate=root_certificate,
+        #     private_key=private_key,
+        #     certificate=certificate,
+        # )
+        # self.aggregator = aggregator_grpc_server.aggregator
+        # await self._run_aggregator_grpc_server(
+        #     aggregator_grpc_server=aggregator_grpc_server,
+        # )
+
```
igor-davidyuk, line 106:
```suggestion
    
```

openfl/component/director/experiment.py
```diff
@@ -69,6 +69,12 @@ async def start(
                         f'collaborators {self.collaborators}')
 
             with ExperimentWorkspace(self.name, self.archive_path):
+                # check for plan approval here
+                # if not self.get_plan_approval(experiment.name, experiment.plan_path):
+                #     experiment.status = Status.FAILED
+
+                # yields control back to event loop
+                # await asyncio.sleep(0)
```
igor-davidyuk, line 77:
```suggestion

```

openfl/interface/cli.py
```diff
@@ -173,6 +176,25 @@ def error_handler(error):
     raise error
 
 
+def review_plan_callback(file_name, file_path):
+    """Review plan callback for Director and Envoy."""
+    echo(style(
+        f'Please review the contents of {file_name} before proceeding...',
+        fg='green',
+        bold=True))
+    time.sleep(3)
```
aleksandr-mokrov, line 188: Please add a comment or named variable for understanding what 3 seconds mean and why it is 3 seconds.

aleksandr-mokrov (approved):

openfl/interface/cli.py
mansishr: That was added for the director/envoy to read the instruction to review the contents of the plan before the plan gets flashed.

psfoley (approved):

psfoley:
Closes https://github.com/intel/openfl/issues/386

Issue #492, opened 2022-08-26
UnboundLocalError: local variable 'request' referenced before assignment


CasellaJr:
For my experiments on CIFAR-10, until now I have used ResNet-18 or an  EfficientNet. In the jupyter notebook, I initialise these models in this way:
```
resnet18 = torchvision.models.resnet18(pretrained=False)
efficientnet_b0 = torchvision.models.efficientnet_b0(pretrained=False)
```
Then they works without problems, also even if I change normalization layers, number of classes and so on, no problem. The problem is when I use a different model, like a VGG16: `vgg16 = torchvision.models.vgg16(pretrained=False)` then I have the following error: 
```SetNewExperiment
                        if request.model_proto:
                    UnboundLocalError: local variable 'request' referenced before assignment
```
This is the complete log of the error:
```
[18:18:37] INFO     🧿 Starting the Director Service.                                                                                             director.py:50
           INFO     Sample shape: ['32', '32', '3'], target shape: ['32', '32', '3']                                                              director.py:74
           INFO     Starting server on 0.0.0.0:50051                                                                                      director_server.py:114
                    '32', '3']}, 'is_online': True, 'is_experiment_running': False, 'valid_duration': 120, 'last_updated': 1661537603.6085486,
                    'experiment_name': None}, 'env_two': {'shard_info': {'node_info': {'name': 'env_two'}, 'shard_description': 'Cifar10
                    dataset, shard number 2 out of 10', 'sample_shape': ['32', '32', '3'], 'target_shape': ['32', '32', '3']}, 'is_online':
                    True, 'is_experiment_running': False, 'valid_duration': 120, 'last_updated': 1661537604.1536303, 'experiment_name': None},
                    'env_seven': {'shard_info': {'node_info': {'name': 'env_seven'}, 'shard_description': 'Cifar10 dataset, shard number 7 out
                    of 10', 'sample_shape': ['32', '32', '3'], 'target_shape': ['32', '32', '3']}, 'is_online': True, 'is_experiment_running':
                    False, 'valid_duration': 120, 'last_updated': 1661537605.1729639, 'experiment_name': None}, 'env_three': {'shard_info':
                    {'node_info': {'name': 'env_three'}, 'shard_description': 'Cifar10 dataset, shard number 3 out of 10', 'sample_shape':
                    ['32', '32', '3'], 'target_shape': ['32', '32', '3']}, 'is_online': True, 'is_experiment_running': False, 'valid_duration':
[...]

PR #494, opened 2022-08-29
Introduce parameters to tune gRPC max message size 

igor-davidyuk:
This PR solves #492 and other related issues. The problem being solved - some model layers are bigger than our max_message_size threshold so experiments fail.
We had scattered message size settings in 5 places: director server, aggregator server, 2 director clients (frontend and envoy), and aggregator client. The biggest part of all our messages is model in most cases, so it would make sense to tune the message size in a centralized manner.
I took the easiest path here - just defined message size in a separate module and imported it everywhere it is needed. I am not sure if we should make it a part of the user interface and leave it for discussion. 

This PR also contains some fixes for the Histology example because I used it for testing


psfoley (approved):
Thanks Igor. Approved!

Issue #501, opened 2022-09-08
Allow setting ISV product ID and SVN in the manifest in OpenFL Gramine build


DL8:
**Is your feature request related to a problem? Please describe.**
Enclave developers are expected to define the following fields for each enclave they maintain:
1. `isvprodid` - used as a persistent high level identifier of the enclave. It is expected to remain constant per enclave (unlike `mrenclave`, which may change with each build)
2. `isvsvn` - security version of the enclave. It is expected to increment when a security vulnerability is fixed

These fields are part of SIGSTRUCT and therefore are signed in enclave build process. Currently these values are not found in OpenFL Gramine manifest, and therefore will be zero. This has two potential problems in production build:
1. Enclave developers may have other enclaves in the ecosystem, each with another product ID. Not allowing them to set the product ID may contradict with their product ID scheme
2. If a vulnerability is found and fixed in OpenFL, ISV SVN is not incremented, potentially leading to data disclosure and having other parties relying on a vulnerable enclave

**Describe the solution you'd like**
ISV product ID should be configurable in OpenFL Gramine build, e.g. via make flag with default 0.

ISV SVN may be handled in a similar way, but I am not sure this is the right thing to do: if a vulnerability is found on OpenFL, the SVN is expected to be incremented with the fix. If this is left for the developers to handle, they might easily miss that, leaving the updated version at risk of potential rollback attack or accepting a potentially vulnerable party in remote attestation.

PR #502, opened 2022-09-08
Add ISV product ID and ISV SVN fields to manifest

DL8:
ISV product ID and ISV SVN fields are defined and signed in SIGSTRUCT. In Gramine, they are defined in the manifest and then used on enclave signing in enclave build flow. Add the option to configure them in OpenFL manifest with new flags to the makefile: SGX_ISVPRODID and SGX_ISVSVN. Their default values are 0, to maintain backwards compatibility.

Currently ISV SVN is configurable in build time, but this is not necessarily the desired behavior. Required behavior is open for discussion and should be updated once a decision is made.

Related issue: #501

### DIFF
--- a/openfl-gramine/Makefile
+++ b/openfl-gramine/Makefile
@@ -1,5 +1,9 @@
 ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)
 
+# Enclave attributes to the manifest
+SGX_ISVPRODID ?= 0
+SGX_ISVSVN ?= 0
+
 # This is a signer key on the BUILDING machine
 SGX_SIGNER_KEY ?= /key.pem
 
@@ -20,6 +24,8 @@ openfl.manifest: openfl.manifest.template
 	gramine-manifest \
 		-Dlog_level=$(GRAMINE_LOG_LEVEL) \
 		-Darch_libdir=$(ARCH_LIBDIR) \
+		-Disvprodid=$(SGX_ISVPRODID) \
+		-Disvsvn=$(SGX_ISVSVN) \
 		-Dno_proxy=$(no_proxy) \
 		-Dhttp_proxy=$(http_proxy) \
 		-Dhttps_proxy=$(https_proxy) \
--- a/openfl-gramine/openfl.manifest.template
+++ b/openfl-gramine/openfl.manifest.template
@@ -51,6 +51,9 @@ sys.stack.size = "4M"
 sgx.thread_num = 512
 #sys.brk.max_size = "1M"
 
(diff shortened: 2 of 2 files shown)


DL8:

DL8:
recheck

igor-davidyuk:
Thank you for your contribution!
I must note that no example in https://github.com/gramineproject/examples feature these fields in manifests and the related section in Gramine docs is rather humble. Thus it is unclear how developers can come up with relevant product ID numbers and automate manifest versioning.

DL8:
> Thank you for your contribution! I must note that no example in https://github.com/gramineproject/examples feature these fields in manifests and the related section in Gramine docs is rather humble. Thus it is unclear how developers can come up with relevant product ID numbers and automate manifest versioning.

I guess the ISV product ID and SVN are not an important detail in Gramine examples, so zero is used implicitly. They are mentioned in [the samples of Intel's SGX SDK](https://github.com/intel/linux-sgx/tree/master/SampleCode), but I looked at some of them and they used zeroes.

More information about those fields can be found [here](https://blog.quarkslab.com/overview-of-intel-sgx-part-1-sgx-internals.html). Quote (emphasis is mine):
> Each enclave is associated with a `SIGSTRUCT` structure, which is signed by its author and contains the enclave measure, signer public key, version number (**`ISV`, reflecting the security level**) and product identifier (**`ISVPRODID`, to distinguish between enclaves from the same author**). It allows to ensure that the enclave hasn't been modified and then re-signed with a different key.

As I understand it, product ID convention is up to enclave developer. Assigning each enclave with a different number (e.g. 1 for server and 2 for client) is a good approach. Once chosen, I don't see any reason to change it.
I don't understand what is manifest versioning. Do you mean ISVSVN? This is not manifest version, but **enclave** security version. It should start with zero and increment when a security vulnerability is fixed (not 100% automatic).

igor-davidyuk:
Thank you for the detailed explanation, @DL8
We can start incrementing `ISVSVN` on our side. Another idea is to bind this number to the OpenFL version, as it is the main application running inside our enclave, do you think it is a good idea?

DL8:
@igor-davidyuk you shouldn't increment `ISVSVN` with every version, but only in case the new version includes security vulnerability fixes. Also, if the SVN was incremented, it should be mentioned in the release notes

igor-davidyuk:
Ok then, seems like there is nothing to add to this PR. Are you going to remove WIP?

DL8:
If you believe this is the right way to implement this, I will remove the WIP

igor-davidyuk (approved):
A good first approach!

PR #524, opened 2022-10-03
Fix torch==1.9.1 for Kvasir CI Test

mansishr:
Checking if version change helps for PyTorch Kvasir CI Test.


psfoley:
@mansishr I'm doing some cleanup of our open PR's. As this version change does not resolve the Kvasir failures, can this be closed?

PR #528, opened 2022-10-07
Break apart CI tests

psfoley:
This PR:
- Breaks existing CI tests into individual pipelines (one for linting, pytest, interactive API, taskrunner, etc.) so it's easier to see what is failing when submitting new functionality


.github/workflows/interactive-kvasir.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.8"
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
```
igor-davidyuk: Do we need flake8 anywhere other than the lint test?
The same for pytest

itrushkin:
Splitting job steps into different workflows causes them to run simultaneously in different environments: all workflows are triggered by pull request action or develop push action **at the same time**. We are not interested in whether experiments are successful if tests have not passed or if there are linter errors. But now workers performing experiment-related workflows will trigger always, even if we do not need it. We cannot control such behavior with workflows, but we could do this with job steps as it was before this pull request. We need to follow the sequence, and users still can see what is failing by following the workflow link in the pull request.  @psfoley What do you think?

CI badge in README.md is also broken after this PR.

psfoley:
Thanks @itrushkin for the comment! There were a few reasons to break apart the CI tests:
- **It makes it easier for users to address the underlying problem in their PR**: With a monolithic CI pipeline that fails and just shows a failed icon next to the commit, a user doesn't know why a failure occurred unless they click on the failed commit and comb through the logs. With individual CI pipelines that are correctly named, a user is presented with information immediately that allows them to categorize the error (lint issue? or particular workflow issue?)
- **Prevents accidental environment pollution**: For months there were sporadic issues with the Kvasir U-Net interactive API test. What appeared to be a failure because of this pipeline was actually likely the result of some set of processes that wasn't cleaned up properly. Individual CI tests create isolated environments, and resulted in resolving those sporadic issues.
- **Faster feedback for committers** - The CI pipelines now complete within 10 minutes of a commit, instead of 30-40 minutes. This is a lot better for users who are trying to fix a pipeline that only failed at the end of the previous script (straggler handling is an example of this)

The downsides to this that you called out are true. There are potential optimizations in terms of which tests to run to if a PR is only focused on documentation, but the criteria to run each action can be tuned in the individual CI pipeline (and I think there's also a way to chain pipelines together, though this does come at the cost of the speed of user feedback). The singular CI badge, as you mentioned, is now a broken link and needs to be fixed. My thinking is that we remove this, and replace it with a section in the README that displays the status for multiple builds, similar to other projects like [Tensorflow](https://github.com/tensorflow/tensorflow#official-builds) and [OnnxRuntime](https://github.com/microsoft/onnxruntime#build-pipeline-status)

PR #530, opened 2022-10-10
[Feature] Timing Fed Components

sun1lach:
Core timeout functionality is in place.
To apply timeouts to any function/method in openfl codebase, import and decorate the function with `@fedtiming`.

```
from openfl.utilities.fed_timer import fedtiming

@fedtiming(timeout=10)    #10 seconds
def some_function(*args, **kwargs):
    pass
```

Update: 20th Oct

The @fedtiming can now be decorated on normal python functions or async co-routine likewise. Usage remains the same.


openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,61 @@
...
+                    args=(), kwargs={}):
+        Thread.__init__(self, group, target, name, args, kwargs)
+        self._result = None
+        self._duration = None
+
+    def run(self):
+        if self._target is not None:
+            start = time.time()
+            self._result = self._target(*self._args, **self._kwargs)
+            self._exec_duration = round(time.time() - start, 3)
```
igor-davidyuk: There is a `_duration` field in `__init__`, should It be assigned here, or do I miss smth?

igor-davidyuk (commented):
Consider using `functool.wraps` to keep function method names for logging and debugging.
https://stackoverflow.com/questions/308999/what-does-functools-wraps-do

openfl/utilities/fed_timer.py
sun1lach: oops! Thanks for pointing it out.
Its a typo in line 20.
updating in next commit.

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,61 @@
...
+
+@parameterize
+def fedtiming(func, timeout):
+    def wrapper(*args, **kwargs):
+        start = time.time()
+        logger.info(f"Started: ({func.__name__})")
+
+        task = CustomThread(target=func, name=func.__name__, args=args, kwargs=kwargs)
+        task.start()
+        task.join(timeout)
```
igor-davidyuk: Does this mean that we will wait for the whole timeout and check the results only after that? So if a task is finished earlier we would not know?

sun1lach: Not really.

For example, for any single instance of a decorated function

if the actual execution time is 2 seconds and the timeout value is 5 seconds then everything runs only for 2 seconds it doesn't wait until the timeout 5 seconds are over.
So in this case the total execution time would be `min(2, 5)`.


If the actual execution time exceeds the timeout value then the `task.join(timeout)` kicks in and we check if the thread is still running/alive then we raise an exception, log the
`TimeoutError` and terminate that thread.
So the max total execution time in this case is just capped by the timeout value i.e., '5 seconds'

Hope it helps.

igor-davidyuk: Got it, thank you!

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,61 @@
...
+
+        result, duration = task.result()
+
+        if duration is None:
+            duration = round(time.time() - start, 3)
```
igor-davidyuk: Is it excessive? Can we just put the timeout time here?

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,61 @@
...
+        return repl
+    return layer        
+
+@parameterize
+def fedtiming(func, timeout):
+    def wrapper(*args, **kwargs):
+        start = time.time()
+        logger.info(f"Started: ({func.__name__})")
+
+        task = CustomThread(target=func, name=func.__name__, args=args, kwargs=kwargs)
```
igor-davidyuk: How does this threaded approach work with asynchronous services?

openfl/utilities/fed_timer.py
sun1lach: `timeout` time is the absolute value passed as a parameter.
above code in line 45 and 46, calculates the execution duration (the time it takes from start to completion of thread).

At line 44, if the duration variable is `None` means there was a timeout and the value assigned to line 45 should actually be a timeout duration.
So, Yes, it makes sense. we can put timeout time here. will update in next commit.

openfl/utilities/fed_timer.py
sun1lach: async methods run in the event loop in its own background context.
Is there a need to timeout and terminate the async methods? 

Although, I didn't try timeout decorator on any async method. Any specific method you have in mind that i can check?

igor-davidyuk: I am asking because one of the options for the aggregator service transformation was to make it asynchronous.
And if you look at the Director service, some of its RPCs are async. 
I am just wondering if there is a universal solution + could you specify in the PR descriptions what functions in our codebase could be targeted by this decorator?

sun1lach: the `@fedtiming` decorator runs the decorated function in an independent thread created using `CustomThread` class.
As long as the async methods awaits the result and has no fire and forget conditions, it can be decorated with `@fedtiming`.

I would strongly suggest not using the `@fedtiming` decorator on a async/sync method which has infinite loop conditions like `while True:`

Since async co-routine runs in its own thread and there are context switches which is taken care by event loop. The `@fedtiming` decorator also provides it own dedicated thread for execution but with the benefit of timeout.

Although, I haven't tested the async/await block with the `@fedtiming` decorator.
Theoretically I am expecting it to work but I won't be certain unless I implement my co-routine and test it with the `@fedtiming` decorator.

I am looking into async decorator.

sun1lach: @igor-davidyuk 
Did a quick run 
async co-routines currently doesn't work with the decorator and thread implementation.

we might require an async decorator and awaitable customthread to make it work for an async co-routines.

Do let me know if you have any thoughts on this and the direction we should be heading towards.
I am certain of the currently implemented timeout decorator usage with normal python functions.

sun1lach:
@igor-davidyuk 
@psfoley

Extended the capability of the `@fedtiming` decorator to support asynchronous co-routine methods as well.
Tested with below async methods in `director.py`

- start_experiment_execution_loop
- wait_experiment

Usage remains the same. Decorate the async methods with
`@fedtiming(timeout=<seconds>)`

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,126 @@
...
+        template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+        message = template.format(type(exc).__name__, exc.args)
+        # print(message)
+        logger.exception(traceback.format_exc())
+        time.sleep(2)
+        os._exit(status=os.EX_TEMPFAIL)
+
+    async def async_wrapper(*args, **kwargs):
+        task = init_task(*args, **kwargs)
+        logger.info(f"Started Async: ({func.__name__})")
```
sun1lach: Although, I didn't face any issue yet. Skimming through the code, I am suspecting the `func.__name__` might throw an exception if you happen to decorate `@fedtiming(timeout)` to a method which is already decorated with `@staticmethod`. Do let me know, if you encounter this. It's a minor change, incase.

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def exception_handler(exc):
+        template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+        message = template.format(type(exc).__name__, exc.args)
+        # print(message)
+        logger.exception(traceback.format_exc())
+        time.sleep(2)
+        os._exit(status=os.EX_TEMPFAIL)
+
+    async def async_wrapper(*args, **kwargs):
```
igor-davidyuk: `async_wrapper` and `wrapper` looks like the same code. Can we just have `wrapper` and hide the complexity inside `task.execute()` using `asyncio.iscoroutinefunction`. The code would be much simpler

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            task.sync_execute()
+        except Exception as exc:
+            exception_handler(exc)
+
+        result, duration = task.result()
+        logger.info(f"({func.__name__}) Execution took : {duration} second(s)")
+        return result
+
+    if asyncio.iscoroutinefunction(func):
```
igor-davidyuk: This is the second check for coroutine, can't we use just one?

igor-davidyuk (changes requested):
1. Please scan this module with flake8, it will highlight a lot of issues.
2. Please add unit tests covering different use cases.

openfl/utilities/fed_timer.py
sun1lach: Tried this approach as well of having a single wrapper but the `async` and `await` declaration requires the enclosing function to also be qualified as `async`. for that reason I had to create a `async_wrapper` as well.
There is a repetition in the code inside `wrapper` and `async_wrapper` decorator, I'll try to minimize that part.

sun1lach:
@igor-davidyuk @psfoley 
- Refactored from decorator function to decorator class.
- Removed redundant and repetitive code.
- sync and async decorator factory for better managing decorator state.
- Implemented `contextmanager` to better handle repetitive/redundant code across async and sync wrappers.
- flake8 compliant.

igor-davidyuk:
> * flake8 compliant.

Looks like our flake8 config is broken. There is no way flake8 would pass with no comment lines

sun1lach:
> > * flake8 compliant.
> 
> Looks like our flake8 config is broken. There is no way flake8 would pass with no comment lines

I see!
I tested the module with the by default/standard flake8 config in my local.
Do let me know if there is any custom flake8 config used in the OpenFL CI to make it compliant, I can validate the same in my local.

Also yes, will add comments and docstrings :)

sun1lach:
Looks like by default `flake` doesn't capture naming_conventions and docstrings etc according to `pycodestyle`.
https://stackoverflow.com/questions/60403545/flake8-not-giving-errors-warnings-on-missing-docstring-or-code-not-following-pep

sun1lach:
This change can be merged independently. 

Static timeout values defined in the `director_config.yml` would require code change in yml parse logic and in any one of the interactive example.

psfoley (changes requested):
@acharyasunil the changes look good on my end. Could you also add documentation for how someone would apply this feature to debug infrastructure / performance issues?

psfoley:
> 1. Please scan this module with flake8, it will highlight a lot of issues.
> 2. Please add unit tests covering different use cases.

From @igor-davidyuk's comment, @acharyasunil It looks like all that is missing at this point are unit tests. After that, this is ready to merge.

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Attributes:
+    target (function): decorated function
+    name (str): Name of the decorated function
+    *args (tuple): Arguments passed as a parameter to decorated function.
+    **kwargs (dict): Keyword arguments passed as a parameter to decorated function.
+
+    '''
+    def __init__(self, group=None, target=None, name=None, args=(), kwargs={}):
+        Thread.__init__(self, group, target, name, args, kwargs)
```
aleksandr-mokrov, line 30: Why not `super().__init__()`?

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,191 @@
...
+        '''
+        task = asyncio.create_task(
+            name=self._fn_name,
+            coro=self._target_fn(*self._args, **self._kwargs)
+        )
+
+        try:
+            await asyncio.wait_for(task, timeout=self._max_timeout)
+        except asyncio.TimeoutError:
+            raise asyncio.TimeoutError(f"Timeout after {self._max_timeout} second(s), Exception method: ({self._fn_name})") # noqa
```
aleksandr-mokrov: It's better avoid #noqa whenever it is possible. Just split it into two lines.

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,191 @@
...
+        task.start()
+        # Execution continues if the decorated function completes within the timelimit.
+        # If the execution exceeds time limit then
+        # the spawned thread is force joined to current/main thread.
+        task.join(self._max_timeout)
+
+        # If the control is back to current/main thread
+        # and the spawned thread is still alive then timeout and raise exception.
+        if task.is_alive():
+            raise TimeoutError(f"Timeout after {self._max_timeout} second(s), Exception method: ({self._fn_name})") # noqa
```
aleksandr-mokrov: It's better avoid #noqa whenever it is possible. Just split it into two lines.

openfl/utilities/fed_timer.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+        try:
+            start = time.perf_counter()
+            yield
+            logger.info(f"({self.task._fn_name}) Elapsed Time: {time.perf_counter() - start}")
+        except Exception as e:
+            template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+            message = template.format(type(e).__name__, e.args[0])
+            logger.exception(message)
+            time.sleep(2)
```
aleksandr-mokrov: It is better two use named variables instead magic numbers. It is not obvious why is this timeout needed?

tests/openfl/utilities/test_fed_timer.py
```diff
@@ -0,0 +1,139 @@
...
+    ])
+def test_check_sync_function_return_same_value_within_timelimit(input, expected):
+    """
+    Test that the decorated synchronous function return the output within the timeout threshold
+    Function call returns the expected output and it is asserted.
+    """
+
+    @fedtiming(timeout=3)
+    def some_sync_fn(value):
+        time.sleep(2)  # Simulate long running operation
```
aleksandr-mokrov: I suggest choosing the minimum possible timeout values for the tests. It's better to avoid long unit tests execution time.

tests/openfl/utilities/test_fed_timer.py
```diff
@@ -0,0 +1,139 @@
...
+import pytest
+import time
+
+from openfl.utilities.fed_timer import fedtiming
+from unittest import mock
+
+
+@pytest.mark.parametrize(
+    'input,expected', [
+        ('first', 'first'),
```
aleksandr-mokrov, line 16: Looks like one pair is enough. Each test include timeouts that increased time execution significantly. If you want to test that testing decorator doesn't change returning values, probably, it is better to do it separately, not in each test type and only for edge cases.

aleksandr-mokrov (changes requested):

tests/openfl/utilities/test_fed_timer.py
sun1lach: Sure, I thought the timeout decorator must work on any given data type. 
Will limit to just one pair in this case and also reduce the timeout period.
I tried to simulate the part where the execution completes within the time limit and exceeds the time limit.
Will choose a smaller timeout value (Y)

openfl/utilities/fed_timer.py
sun1lach: I think its pretty much the same thing.
using super() would benefit only during multiple inheritance situation like `class C(A, B)`.

openfl/utilities/fed_timer.py
sun1lach: Timeout is not really needed. I used it to just have a delay before the SIGKILL kicks in.
This functionality would be extended later to support graceful termination and this particular os._exit() call is not ideal.

tests/openfl/utilities/test_fed_timer.py
```diff
@@ -0,0 +1,111 @@
...
+    ])
+def test_check_sync_function_return_same_value_within_timelimit(input, expected):
+    """
+    Test that the decorated synchronous function return the output within the timeout threshold
+    Function call returns the expected output and it is asserted.
+    """
+
+    @fedtiming(timeout=2)
```
aleksandr-mokrov: Is it possible to reduce timeout even more to 0.1-0.3 seconds? Looks like it is enough for each platform where tests can be run.

aleksandr-mokrov (approved):

tests/openfl/utilities/test_fed_timer.py
sun1lach: Yup
Did a quick timeout functionality checks with sub-second timeout values.
Works as expected. Committed the changes

psfoley (approved):

PR #532, opened 2022-10-10
Support Python 3.9 and 3.10

psfoley:
Addresses https://github.com/intel/openfl/issues/522

This PR:
- Adds CI Tests for Python 3.9 and 3.10 (TaskRunner only so far)
- Removes dependence on grpcio 1.34.0, which is not supported by 3.10


igor-davidyuk (approved):
It is great we now support newer Python versions + bumped grpc

mansishr (approved):

PR #536, opened 2022-10-19
Introduce the `install requirements` option in director and envoy configs

igor-davidyuk:
Previously this was hardcoded: Director always skipped the installation of requirements and Envoy always installed related requirements.
Adding this new option is inspired by the trend of separating Local test runtime and Federated runtime in OpenFL. It will allow faster test runs.
Moreover, it resolves the issue that blocks installing requirements by Envoys met in Google Colab.


openfl/component/envoy/envoy.py
```diff
@@ -84,9 +86,9 @@ def run(self):
             self.is_experiment_running = True
             try:
                 with ExperimentWorkspace(
-                        self.name + '_' + experiment_name,
-                        data_file_path,
-                        is_install_requirements=True
+                        experiment_name=self.name + '_' + experiment_name,
```
MasterSkepticista: Consider using f-strings for readability

`experiment_name=f'{self.name}_{experiment_name}'`

MasterSkepticista (approved):

openfl/component/envoy/envoy.py
igor-davidyuk:
```suggestion
                        experiment_name=f'{self.name}_{experiment_name}',
```

PR #538, opened 2022-10-21
Fix log metric callback settings

mansishr:
This PR fixes a current bug w.r.t custom log metric callback function: Earlier, only tensorboard logging was enabled in the aggregator settings, overriding the log settings in plan. With this fix, users should be able to define a custom log metric callback function and that would be reflected in the aggregator settings.


igor-davidyuk (approved):
Looks good!

psfoley (approved):

PR #547, opened 2022-10-26
For gramine set write logs to false

mansishr:
Gramine doesn't support the use of POSIX semaphores, which is required by Python's `multiprocessing` package. By default, tensorboard logging is enabled which [invokes multiprocessing internally](https://github.com/intel/openfl/issues/503#issuecomment-1253563578) to write logs. Hence, `write_logs` should be disabled for `gramine_ready` workspaces.
Another solution might be to create a custom logging function that doesn't use multiprocessing but since we are using Gramine that has it's own way of logging through `log_level` in manifest, we might as well disable `write_logs` in plan.


igor-davidyuk (approved):
Great

psfoley (approved):

PR #550, opened 2022-10-27
Convert protobuf model to native format

### LINKED ISSUE #552: Load a trained pbuf model into general pytorch env
I would like to test the trained model externally, using an independent pytorch-based script that compares several models. The pbuf format is serialized differently that the regular .pth torch snapshots and can't be loaded using model.load_state_dict.

It would be great if you can provide either a tool for converting pbuf to pth, or a flag enabling to save the model in pth format during the learning rounds.

Thanks a lot.

psfoley:
This PR:
- Adds a new command, `fx model save`, that takes as input the `.pbuf` model file produced by a federated experiment and converts it to the native PyTorch / TensorFlow model representation for future use.

~~This PR is WIP~~. The core functionality to be refactored by @igor-davidyuk into the `utilities` folder, so this can also be used via API.

closes #552

### DIFF
--- a/openfl/interface/model.py
+++ b/openfl/interface/model.py
@@ -70,6 +70,10 @@ def get_model(
 ) -> TaskRunner:
     """
     Initialize TaskRunner and load it with provided model.pbuf.
+
+    Contrary to its name, this function returns a TaskRunner instance.
+    The reason for this behavior is the flexibility of the TaskRunner interface and
+    the diversity of the ways we store models in our template workspaces.
     """
 
     model_protobuf_path = Path(model_protobuf_path).absolute()
--- a/openfl/interface/model.py
+++ b/openfl/interface/model.py
@@ -15,6 +15,7 @@ from openfl.federated import Plan
 from openfl.federated import TaskRunner
 from openfl.protocols import utils
 from openfl.pipelines import NoCompressionPipeline
+from openfl.utilities.workspace import set_directory
 
 logger = getLogger(__name__)
 
@@ -76,16 +77,24 @@ def get_model(
     the diversity of the ways we store models in our template workspaces.
     """
 
-    model_protobuf_path = Path(model_protobuf_path).absolute()
+    # Here we change cwd to the experiment workspace folder
+    # because plan.yaml usually containe relaiteve paths to components.
+    workspace_path = Path(plan_config).resolve().parent.parent
+    plan_config = Path(plan_config).resolve().relative_to(workspace_path)
+    cols_config = Path(cols_config).resolve().relative_to(workspace_path)
+    data_config = Path(data_config).resolve().relative_to(workspace_path)
 
-    plan = Plan.parse(plan_config_path=Path(plan_config),
-                      cols_config_path=Path(cols_config),
-                      data_config_path=Path(data_config))
+    with set_directory(workspace_path):
+        plan = Plan.parse(
+            plan_config_path=plan_config,
+            cols_config_path=cols_config,
+            data_config_path=data_config
+        )
+        collaborator_name = list(plan.cols_data_paths)[0]
+        data_loader = plan.get_data_loader(collaborator_name)
+        task_runner = plan.get_task_runner(data_loader=data_loader)
 
-    collaborator_name = list(plan.cols_data_paths)[0]
-    data_loader = plan.get_data_loader(collaborator_name)
-    task_runner = plan.get_task_runner(data_loader=data_loader)
-
+    model_protobuf_path = Path(model_protobuf_path).resolve()
     logger.info(f'Loading OpenFL model protobuf:  🠆 {model_protobuf_path}')
 
     model_protobuf = utils.load_proto(model_protobuf_path)
--- a/openfl/interface/model.py
+++ b/openfl/interface/model.py
@@ -1,14 +1,19 @@
 # Copyright (C) 2020-2022 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-"""Plan module."""
+"""Model CLI module."""
 
 from logging import getLogger
+from pathlib import Path
+from typing import Union
 
-from click import echo
 from click import group
 from click import option
 from click import pass_context
 from click import Path as ClickPath
+from openfl.federated import Plan
+from openfl.federated import TaskRunner
+from openfl.protocols import utils
+from openfl.pipelines import NoCompressionPipeline
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


igor-davidyuk:
We want to introduce an additional piece of functionality in two forms:
- CLI: `fx model save` command called from the workspace must save the model on disk.
- Python API: `openfl.model_save(path)` must return the model object.

We start with `model.pbuf` path to the model snapshot that we need to decompress and load weights into the model object to save or return it
1. To save the model in a native format we need `save_native` method implemented in `TaskRunner`
This method is not implemented in all our template `TaskRunners` but this is fine as we may ask users to implement it.
2. To return the model as a Python object we need to access this object
Models are stored in different ways in our template `TaskRunners`. Moreover, there is a particular problem with TF1 examples. We may need to introduce an additional `get_model` method to all of our template `TaskRunners`.
3. To get access to the model we need to initialize `TaskRunner`
Model initialization happens in `TaskRunner` code and often requires `feature_shape` information to build a model.
4. To initialize `TaskRunner `we need the `plan.yaml` path and `DataLoader`
Again, model initialization is coupled with `DataLoader` initialization.
5. To initialize `DataLoader` we need `data.yaml` and `cols.yaml`
Here we use the first collaborator's name to get its `data_path` and load data to get the `feature_shape`. It works for our templates, which use synthetic or downloaded datasets, but in real-world usecases, users often may not have access to any data from their machine. A way to overcome this obstacle may be to define `feature_shape` in `plan.yaml` or hardcode it inside `DataLoader`.

With all said above, at this point, we can not guarantee `model save` API functioning to all users in the intended way.

igor-davidyuk:
As a solution to the abovementioned problems I propose the following approach:
1. For the CLI command we will rely on `TaskRunner`'s `save_native` method that users can implement or change if needed.
2. For the Python API part, we will return a `TaskRunner` object, thus allowing users to interact with the linked model.

These signatures allow us to mitigate issues caused by the diversity of model definition forms in our TuskRunners and yet deliver the expected functionality. At the same time, the data problem mentioned under point 5 is ignored, as we know it is already solved prior to executing `fx plan initialize` which follows a similar path.

Work left in this PR:
- [x] Update the docs
- [x] Write tests

openfl/interface/model.py
```diff
@@ -0,0 +1,87 @@
...
+@option('-c', '--cols-config', required=False,
+        help='Authorized collaborator list [plan/cols.yaml]',
+        default='plan/cols.yaml', type=ClickPath(exists=True))
+@option('-d', '--data-config', required=False,
+        help='The data set/shard configuration file [plan/data.yaml]',
+        default='plan/data.yaml', type=ClickPath(exists=True))
+@option('-i', '--input', 'model_protobuf_path', required=True,
+        help='The model protobuf to convert',
+        type=ClickPath(exists=True))
+@option('-o', '--output', 'output_filepath', required=False,
```
itrushkin: We could require this option as well for users to make sure they provide the path for the saved model. In case of someone forgot to specify it, it would be easier to notify them about the wrong command call and ask for the path. Otherwise, they have to search the save path in the console output.

igor-davidyuk: I understand the intention, but I would argue that we should make arguments `required` only if they are literally required.

itrushkin: `output_filepath` is required by the `_save` function for calling `TaskRunner.save_native`.

For example, [`torch.save`](https://pytorch.org/docs/stable/generated/torch.save.html)  and [`tf.keras.Model.save_weights`](https://www.tensorflow.org/api_docs/python/tf/keras/Model#save_weights) do require a file argument.

igor-davidyuk: Yes, but those functions are python function calls, in our case we have a CLI command which is called from a specific place, namely, an experiment workspace.
Honestly, I am not against making this argument required. @mansishr, what do you think?

aleksandr-mokrov: I believe that here is used good approach with default value. It is checked that file already exists and is asked user for the confirmation to rewrite. At the end, it is written to the output in witch path the model is saved.

aleksandr-mokrov (approved):

openfl/interface/model.py
```diff
@@ -76,16 +77,24 @@ def get_model(
...
-    plan = Plan.parse(plan_config_path=Path(plan_config),
-                      cols_config_path=Path(cols_config),
-                      data_config_path=Path(data_config))
-
-    collaborator_name = list(plan.cols_data_paths)[0]
-    data_loader = plan.get_data_loader(collaborator_name)
-    task_runner = plan.get_task_runner(data_loader=data_loader)
-
+    # Here we change cwd to the experiment workspace folder
+    # because plan.yaml usually containe relaiteve paths to components.
```
itrushkin:
```suggestion
    # because plan.yaml usually contains relative paths to components.
```

openfl/interface/model.py
mansishr: I agree, specifying a default path helps for our CLI command.

tests/openfl/interface/test_model_api.py
```diff
@@ -0,0 +1,38 @@
...
+
+from openfl import get_model
+from openfl.federated.task import TaskRunner
+
+
+@mock.patch('openfl.interface.model.utils')
+@mock.patch('openfl.interface.model.Plan')
+def test_get_model(Plan, utils):
+    "Test get_module returns TaskRunner."
+    plan_instanse = mock.Mock()
```
mansishr: nitpicking, typo for instance

mansishr (commented):
Looks good to me

tests/openfl/interface/test_model_api.py
igor-davidyuk:
```suggestion
    plan_instance = mock.Mock()
```

tests/openfl/interface/test_model_api.py
```diff
@@ -0,0 +1,38 @@
...
+    # Function call
+    result = get_model('plan_path', 'cols_path', 'data_path', 'model_protobuf_path')
+
+    # Asserts below
+    Plan.parse.assert_called_once()
+
+    utils.load_proto.assert_called_once()
+    utils.deconstruct_model_proto.assert_called_once()
+
+    plan_instanse.get_task_runner.assert_called_once()
```
igor-davidyuk:
```suggestion
    plan_instance.cols_data_paths = ['mock_col_name']
    Plan.parse.return_value = plan_instance

    plan_instance.get_task_runner.return_value = TaskRunner(data_loader=mock.Mock())
    TaskRunner.set_tensor_dict = mock.Mock()

    tensor_dict = mock.Mock()
    utils.deconstruct_model_proto.return_value = tensor_dict, {}

    # Function call
    result = get_model('plan_path', 'cols_path', 'data_path', 'model_protobuf_path')

    # Asserts below
    Plan.parse.assert_called_once()

    utils.load_proto.assert_called_once()
    utils.deconstruct_model_proto.assert_called_once()

    plan_instance.get_task_runner.assert_called_once()
```

mansishr (approved):

PR #537, opened 2022-10-20
Gramine manifest update

igor-davidyuk:
This PR is here to update our manifest template with the latest changes in Gramine guidelines.

PR #557, opened 2022-11-01
OpenFL-Gramine fixes for SGX execution

Giemp95:
This PR aims to make the OpenFL-Gramine workflow smoother for the user.

I have modified the openfl.manifest.template to include the mount of the /tmp directory since the lack of it was causing errors in the execution of the container.
Furthermore, I have updated the Dockerfile.gramine file so that now it will download and install the last development branch of the OpenFL software. This way, the version of OpenFL installed inside the container will be the same one used for development, and there will be no mismatch between OpenFL versions inside and outside the container. Once a new stable version of OpenFL is released, this will need to be changed to point to the last stable version.
Lastly, I've extended the Manual.md file to include many small things I learned while trying to make the OpenFL-Gramine example work. Now it should be easier for a newbie to run this application.

I have tested all the changes I made, and they are working as they are.
The only difference to this PR  is that, in my tests, the Docker image contained my project fork; in this PR I have changed that pointer to the developmental branch of OpenFL so that, if integrated, it will work out-of-the-box.

### DIFF
--- a/openfl-gramine/MANUAL.md
+++ b/openfl-gramine/MANUAL.md
@@ -36,6 +36,10 @@ export TEMPLATE_NAME=torch_cnn_histology
 fx workspace create --prefix $WORKSPACE_NAME --template $TEMPLATE_NAME
 cd $WORKSPACE_NAME
 ```
+Be sure that the Python versions used on the building node and inside the Docker container both support the requirement.txt files of your selected workspace.
+If that is not the case, proceed to manually modify your selected workspace requirement.txt file in the openfl-workspace installation folder to obtain a congruous environment.
+Also, if you are planning to use the SGX support, be sure that that requirements.txt file does not contain any reference to CUDA distributions.
+
 Modify the code and the plan.yaml, set up your training procedure. </br>
 Pay attention to the following: 
 - make sure the data loading code reads data from ./data folder inside the workspace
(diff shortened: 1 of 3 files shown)


openfl-gramine/MANUAL.md
```diff
@@ -2,16 +2,28 @@
...
+It is, by the way, not mandatory that these machines are physically different. A single machine can be used to carry out multiple functions; this choice is made only to make the following tutorial clearer.
+</br>
+In the following the requirements for the different machines will be presented.
+
 Building machine:
 - OpenFL
 - Docker should be installed, user included in Docker group
 
 Machines that will run an Aggregator and Collaborator containers should have the following:
+- OpenFL
```
DL8, line 20: are you sure that's the case? as far as i understand, the docker image is used, so aggregator and collaborators using it don't really need openfl installed

openfl-gramine/openfl.manifest.template
```diff
@@ -38,6 +38,10 @@ fs.mount.workspace.type = "chroot"
 fs.mount.workspace.path = "/workspace"
 fs.mount.workspace.uri = "file:/workspace"
 
+fs.mount.tmp.type = "chroot"
```
DL8, line 41: does gramine support a tmpfs-style mount? if /tmp is used only for temporary files, it may be better to tell gramine that /tmp is in memory, so there will be no leftovers after its execution

DL8: i looked in gramine docs and found the `tmpfs` filesystem type. i believe it can be more appropriate in this context

igor-davidyuk:
Some manifest updates were done here #537

openfl-gramine/MANUAL.md
Giemp95: Yes, you're right that the training will be carried out with the Docker container, in which OpenFL will already be present. However, the user will still need the OpenFL functionalities available outside the Docker container to carry out the exchange and validation of the certificates, as reported in step 7 of the actual OpenFL-gramine workflow reported in Manual.md.

PR #560, opened 2022-11-02
Add flake8 extensions to linter CI workflow

itrushkin:
This pull request contains:
 - Including installation of flake8 extensions
 - Making linter workflow fail on any non-ignored (see setup.cfg) error.
 - Removing linter rules specified directly in workflow code. All rules have already been specified in setup.cfg file in the repository root.
 - Fixing all linter errors 

Tested the behavior locally via [act](https://github.com/nektos/act).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 
+        pip install -r requirements-linters.txt 
         pip install .
     - name: Lint with flake8
       run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,7 +30,4 @@ jobs:
         pip install .
     - name: Lint with flake8
       run: |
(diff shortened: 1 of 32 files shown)


itrushkin:
New flake8 rules have been applied. I will convert this to a draft and fix all the issues here.

aleksandr-mokrov (approved):

igor-davidyuk (commented):
Great, thank you!
One thing that I would do is separate this into two PRs.
One will contain changes to CI and another one with linter highlited fixes.

itrushkin:
@igor-davidyuk It seems that CI changes affect GitHub actions immediately. I had linter errors after enabling the extensions, thus I had to fix them in the same pull request.

igor-davidyuk:
> @igor-davidyuk It seems that CI changes affect GitHub actions immediately. I had linter errors after enabling the extensions, thus I had to fix them in the same pull request.

Yeah, I thought it was the case. We could Merge a PR with fixes first to avoid this problem 🤣
But this is not that important, one combined PR is ok too.

PR #569, opened 2022-11-03
Configurable enclave size

### LINKED ISSUE #568: Enclave size should be configurable in Graminize operation

Enclave size of Graminized OpenFL is hardcoded to 16GB. Enclave size is not a one-size-fits-all and therefore should be configurable.

Add enclave size parameter to the makefile:
`make SGX_SIZE=16G`

If possible, it should also be passed as parameter to image build in `fx workspace graminize`

DL8:
Currently OpenFL enclave size is hardcoded to 16G. Add option to
configure it in build time:
Add the "-e" argument to "fx workspace graminize", to allow setting
the size of the enclave that will be used in the image. This setting
propagates as a Docker build argument to the makefile in openfl-gramine,
which eventually becomes an argument in the generated manifest.
Default size is 16G, to maintain backwards compatibility.

Fixes #568


igor-davidyuk (approved):
Great work!

mansishr (approved):
LGTM!

Issue #554, opened 2022-10-31
"fx workspace graminize" does not regard the version used to build the workspace


DL8:
**Describe the bug**

`fx workspace graminize` always uses latest version of OpenFL from Pip, regardless of version used in workspace creation. In case head of develop branch is used, version 1.4 is added to `requirements.txt` and `docker build` fails because there is no such version in Pip repository

**To Reproduce**

```bash
python -m venv venv
. venv/bin/activate
git clone https://github.com/intel/openfl.git
cd openfl
pip install .
fx workspace create --prefix workspace --template torch_unet_kvasir_gramine_ready
cd workspace
fx plan initialize -a `hostname --all-fqdns`
openssl genrsa -3 -out key.pem 3072
fx workspace graminize --rebuild -s key.pem
```

**Expected behavior**

Gramine-based Docker image is created based on the version used to build the workspace (in this case head of develop branch)

DL8:
Build log file:
[build.log](https://github.com/intel/openfl/files/9901947/build.log)

igor-davidyuk:
Thanks for reporting this issue!
It opens several questions:
1. Should we pin the OpenFL version in the experiment plan?
2. Should we ensure users create their Gramine-targeted experiments with the latest stable version of OpenFL?
3. Should we export OpenFL as a requirement for experiments?

@mansishr, what do you think?

mansishr:
@igor-davidyuk IMO exporting OpenFL as a requirement for experiments is the easiest approach for gramine ready workspaces, although it may look counterintuitive to users. I am not sure of the first option and the second option seems like an additional step that users might have to implement.

mansishr:
We have an external PR which fixes this issue #557

teoparvanov:
We are fundamentally reworking the graminization flow in OpenFL 1.7, so I'm closing this issue.

PR #570, opened 2022-11-03
OpenFL-Gramine fixes for SGX execution [reupload]

Giemp95:
[This PR is a copy of a previous one closed by error due to some branching reorganization]

This PR aims to make the OpenFL-Gramine workflow smoother for the user.

I have modified the openfl.manifest.template to include the mount of the /tmp directory since the lack of it was causing errors in the execution of the container.
Furthermore, I have updated the Dockerfile.gramine file so that now it will download and install the last development branch of the OpenFL software. This way, the version of OpenFL installed inside the container will be the same one used for development, and there will be no mismatch between OpenFL versions inside and outside the container. Once a new stable version of OpenFL is released, this will need to be changed to point to the last stable version.
Lastly, I've extended the Manual.md file to include many small things I learned while trying to make the OpenFL-Gramine example work. Now it should be easier for a newbie to run this application.

I have tested all the changes I made, and they are working as they are.
The only difference to this PR is that, in my tests, the Docker image contained my project fork; in this PR I have changed that pointer to the developmental branch of OpenFL so that, if integrated, it will work out-of-the-box.

[Comments]
DL8: "are you sure that's the case? as far as i understand, the docker image is used, so aggregator and collaborators using it don't really need openfl installed"
Answer: "Yes, you're right that the training will be carried out with the Docker container, in which OpenFL will already be present. However, the user will still need the OpenFL functionalities available outside the Docker container to carry out the exchange and validation of the certificates, as reported in step 7 of the actual OpenFL-gramine workflow reported in Manual.md."

DL8: "does gramine support a tmpfs-style mount? if /tmp is used only for temporary files, it may be better to tell gramine that /tmp is in memory, so there will be no leftovers after its execution"
DL8: "i looked in gramine docs and found the tmpfs filesystem type. i believe it can be more appropriate in this context"
Answer: "I've changed the mount type for /tmp from 'chmod' to 'tmpfs'"

igor-davidyuk: "Some manifest updates were done here https://github.com/intel/openfl/pull/537"

### DIFF
--- a/openfl-gramine/Dockerfile.gramine
+++ b/openfl-gramine/Dockerfile.gramine
@@ -5,7 +5,8 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN --mount=type=cache,target=/root/.cache/ \
     pip install --upgrade pip && \
-    pip install openfl
+    git clone --single-branch --branch develop https://github.com/intel/openfl.git && \
+    pip install ./openfl
 
 # install gramine
 RUN curl -fsSLo /usr/share/keyrings/gramine-keyring.gpg https://packages.gramineproject.io/gramine-keyring.gpg && \
--- a/openfl-gramine/MANUAL.md
+++ b/openfl-gramine/MANUAL.md
@@ -36,6 +36,10 @@ export TEMPLATE_NAME=torch_cnn_histology
 fx workspace create --prefix $WORKSPACE_NAME --template $TEMPLATE_NAME
 cd $WORKSPACE_NAME
 ```
+Be sure that the Python versions used on the building node and inside the Docker container both support the requirement.txt files of your selected workspace.
+If that is not the case, proceed to manually modify your selected workspace requirement.txt file in the openfl-workspace installation folder to obtain a congruous environment.
+Also, if you are planning to use the SGX support, be sure that that requirements.txt file does not contain any reference to CUDA distributions.
+
 Modify the code and the plan.yaml, set up your training procedure. </br>
 Pay attention to the following: 
 - make sure the data loading code reads data from ./data folder inside the workspace
@@ -47,7 +51,7 @@ Default workspaces (templates) in OpenFL differ in their data downloading proced
 - torch_cnn_histology
 - keras_nlp
 
-Also the other worksapces can be used by taking care of placing the dataset used under a data/ folder.
+Also the other workspaces can be used by taking care of placing the dataset used under a data/ folder and modifying the requirements.txt file so that it does not contain references to CUDA distributions.
 
 2. **Initialize the experiment plan** </br> 
 Find out the FQDN of the **Aggregator machine** and use it for plan initialization.
@@ -71,6 +75,9 @@ openssl genrsa -3 -out $KEY_LOCATION/key.pem 3072
 This key will not be packed into the final Docker image.
 
 4. **Build the Experiment Docker image**
+
+Before building the image, according to your SGX-capable processor, it could be necessary to turn off the logging TensorBoardX logging functionality from the plan.
+This can be done by simply setting `write_logs: false` in the plan/plan.yaml file.
 
 ```
 fx workspace graminize -s $KEY_LOCATION/key.pem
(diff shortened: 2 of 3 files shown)


mansishr:
Thanks @Giemp95 for submitting this PR! There is some overlap between this PR and #537 that we should resolve. Regarding ensuring that we use the same version of OpenFL for workspace and for docker image, the team needs to reach a consensus on the thread going on in Issues #554.

openfl-gramine/openfl.manifest.template
```diff
@@ -38,6 +38,10 @@ fs.mount.workspace.type = "chroot"
 fs.mount.workspace.path = "/workspace"
 fs.mount.workspace.uri = "file:/workspace"
 
+fs.mount.tmp.type = "tmpfs"
+fs.mount.tmp.path = "/tmp"
+fs.mount.tmp.uri = "file:/tmp"
```
DL8: the "uri" attribute is not relevant when using tmpfs, so this line can be removed

openfl-gramine/MANUAL.md
```diff
@@ -2,16 +2,28 @@
...
+It is, by the way, not mandatory that these machines are physically different. A single machine can be used to carry out multiple functions; this choice is made only to make the following tutorial clearer.
+</br>
+In the following the requirements for the different machines will be presented.
+
 Building machine:
 - OpenFL
 - Docker should be installed, user included in Docker group
 
 Machines that will run an Aggregator and Collaborator containers should have the following:
+- OpenFL
```
igor-davidyuk: You do not need OpenFL on running machines

openfl-gramine/openfl.manifest.template
```diff
@@ -38,6 +38,10 @@ fs.mount.workspace.type = "chroot"
...
+
```
igor-davidyuk: These changes are already in develop

igor-davidyuk (changes requested):

openfl-gramine/MANUAL.md
igor-davidyuk: >However, the user will still need the OpenFL functionalities available outside the Docker container to carry out the exchange and validation of the certificates, as reported in step 7 of the actual OpenFL-gramine workflow reported in Manual.md.

We must specify this, as our PKI is just a stub and users may have their own certificates obtained elsewhere.
Something like this:
```suggestion
- OpenFL (If used for certification through built-in tooling)
```

igor-davidyuk:
Thank you for the extensions of the manual! I know that working on gramine was hard lately because of many non-merged fixes required to run an example.
Moreover, managing OpenFL versions locally on the building machine and in the Docker container may be tricky and confusing. So this temporary stub for installing develop branch will not solve the problem as I tried it before 🥲, I started to think that we need some `Developer mode` for the `graminize` command that will allow us to replicate local OpenFL build in Docker images.

DL8:
@igor-davidyuk I like the developer mode idea. To extend it further, I would also want it to be possible to choose which repository and branch/revision to clone in the image build. The motivation is this: for development purposes, I have a private fork of OpenFL with a couple of branches I'm working on. To make it work, I had to manually hard-code my fork's address and the branch I'm using (and remember to switch the branch name as needed).

Regarding certification and PKI work, I think an example of how to do it without installing OpenFL should at the very least be documented. I also encountered some technical challenges with it, so I think it's worth a separate issue.

openfl-gramine/openfl.manifest.template
Giemp95: Done!

openfl-gramine/openfl.manifest.template
Giemp95: Thanks, I've rebased the manifest to the latest develop commit

Giemp95 (commented):
Review accepted, manifest rolled back to latest develop commit

openfl-gramine/openfl.manifest.template
```diff
@@ -18,25 +18,15 @@ fs.start_dir="/workspace"
...
-fs.mount.workspace.uri = "file:/workspace"
+fs.mounts = [
+  { uri = "file:{{ gramine.runtimedir() }}", path = "/lib" },
+  { uri = "file:{{ arch_libdir }}", path = "{{ arch_libdir }}" },
+  { uri = "file:/usr", path = "/usr" },
+  { uri = "file:/etc", path = "/etc" },
+  { uri = "file:/workspace", path = "/workspace" },
+
+  { type = "tmpfs", path = "/tmp" },
+]
```
igor-davidyuk: I do not get why these changes are here, they are already in develop. Can you rebase your branch?

Giemp95: Done

openfl-gramine/Dockerfile.gramine
```diff
@@ -5,7 +5,8 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 RUN --mount=type=cache,target=/root/.cache/ \
     pip install --upgrade pip && \
-    pip install openfl
+    git clone --single-branch --branch develop https://github.com/intel/openfl.git && \
+    pip install ./openfl
```
igor-davidyuk, line 9: Please remove this fix as we will introduce a flexible `development mode` instead

openfl-gramine/openfl.manifest.template
```diff
@@ -68,7 +71,6 @@ sgx.trusted_files = [
 ]
 
 sgx.allowed_files = [
-  "file:/tmp",
   "file:/etc",
```
igor-davidyuk, line 74: This is also already fixed

PR #587, opened 2022-11-14
CI for Windows

itrushkin:
### **CI support for Windows**
To adapt test code for Windows, it is recommended to use an OS-invariant tool. I would like to use Python files for it. Please comment if you have any concerns.
### Replacing test bash scripts with Python files
| Test | Tested on Windows | Tested on Ubuntu (20.04)
| --- | :---: | :---: |
| `test_hello_federation` | ✔️   | ✔️ |
| `dockerization_test` | :x: Requires switching Docker to Linux Engine | ✔️ |
| `test_double_ws_export` | ✔️  | ✔️ |
| `test_fets_challenge` | ❌ ([GANDLF lib issue](https://github.com/intel/openfl/actions/runs/3612836210/jobs/6088209680)) | ✔️ |
| `test_graminize` | N/A | ❌(see Graminize issue below) |
- [x] Include Windows jobs in CI workflows

### WIndows Issues found

1. **TaskRunner** | [Job link](https://github.com/intel/openfl/actions/runs/3482499425/jobs/5824890680#step:5:218)
`click` tries to write to a file with `cp-1252` encoding. Gets an error:
```
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-2: character maps to <undefined>
```
Fixed in #610.
2. **Graminize** | 
`fx workspace import` fails with an error:
```
ERROR: No matching distribution found for openfl==1.4
```


itrushkin:
I think we should not include the Graminize test in our CI. It fails even on Ubuntu because of PyPI OpenFL installation instead of the local one. The error is a gRPC version mismatch: https://github.com/intel/openfl/actions/runs/3648409877/jobs/6161798077#step:5:2461

psfoley (approved):
Very nicely done!

mansishr (approved):
LGTM, well done!

tests/github/test_fets_challenge.py
```diff
@@ -0,0 +1,104 @@
...
+    archive_name = f'{fed_workspace}.zip'
+    fqdn = socket.getfqdn()
+    template = args.template
+    rounds_to_train = args.rounds_to_train
+    col1, col2 = args.col1, args.col2
+    col1_data_path, col2_data_path = args.col1_data_path, args.col2_data_path
+    shutil.rmtree(fed_workspace, ignore_errors=True)
+    check_call(['fx', 'workspace', 'create', '--prefix', fed_workspace, '--template', template])
+    os.chdir(fed_workspace)
+    print(origin_dir)
```
aleksandr-mokrov: Looks like forgotten print, or comment is missing

openfl/utilities/utils.py
```diff
@@ -258,3 +260,11 @@ def change_tags(tags, *, add_field=None, remove_field=None) -> Tuple[str, ...]:
 
     tags = tuple(sorted(tags))
     return tags
+
+
+def rmtree(path, ignore_errors=False):
+    def remove_readonly(func, path, _):
+        "Clear the readonly bit and reattempt the removal"
+        os.chmod(path, stat.S_IWRITE)  # Windows can not remove read-only files.
```
aleksandr-mokrov: I suggest to add OS type checking to avoid file permission increase when it is not necessary. `if os.name == 'nt':`  for example.

tests/github/interactive_api_director/experiments/pytorch_kvasir_unet/data_loader.py
```diff
@@ -2,16 +2,22 @@
 from skimage import io
 import numpy as np
 from openfl.utilities import validate_file_hash
+import requests
```
aleksandr-mokrov, line 5: Imports are not in right order. Probably after these changes is it possible to remove `tests/github/interactive_api_director` from excluded paths for flake8?

itrushkin: Makes sense. I will create a separate pull request for this after merging this one.

aleksandr-mokrov (approved):

PR #610, opened 2022-11-21
[Windows] UnicodeDecodeError

itrushkin:
This is a fix for Windows that has default CP1252 encoding for I/O operations. We have some lines with Unicode characters that are written in `stdout`. The fix is enforcing UTF-8 encoding to all non-byte I/O operations.

The issue was found in #587: https://github.com/intel/openfl/actions/runs/3508716155/jobs/5877315977#step:5:211

I have tested it on Windows using `.github/workflows/taskrunner.yml` and `tests/github/test_hello_federation.py` from #587. ([Job result](https://github.com/intel/openfl/actions/runs/3539363355/jobs/5941173807))

In addition, I have replaced Linux-based separators with `os.sep` to use the correct paths on Windows systems.


itrushkin:
Waiting for #619 to be merged.

igor-davidyuk (approved):
Well done!

aleksandr-mokrov (approved):

PR #620, opened 2022-11-24
Add random delay in authentication failures

### LINKED ISSUE #617: Use random delay in authentication failures (gRPC TLS)
Use random delays when GRPC authentication fails so that the period needed to generate an authentication failure message is randomized. Attackers can use the length of time it takes to generate an authentication error message to guess the reason it failed.

igor-davidyuk:
Added a random delay before raising an exception when a certificate name does not match the corresponding collaborator name.
All other authentication is done automatically by gRPC during TLS handshake so we have no power over its behavior.

closes #617 

Signed-off-by: igor-davidyuk <igor.davidyuk@intel.com>

### DIFF
--- a/openfl/transport/grpc/aggregator_server.py
+++ b/openfl/transport/grpc/aggregator_server.py
@@ -5,6 +5,7 @@
 
 import logging
 from concurrent.futures import ThreadPoolExecutor
+from random import random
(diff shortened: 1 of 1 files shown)


psfoley (approved):
LGTM!

PR #623, opened 2022-11-28
Handling envoy disconnections on getting experiment.

aleksandr-mokrov:
If disconnection between a director and an envoy happens after an experiment is set but not yet started, the experiment will stuck waiting for the lost envoy to return. But the envoy will never return. 
This PR fix that problem and in addition handle an envoy restart before 0 round is completed. 
RPC method WaitExperiment doesn't use stream options, but it requires additional client logic in disconnection cases. Removing it makes code simpler and more fault-tolerant.

### DIFF
--- a/openfl/component/director/director.py
+++ b/openfl/component/director/director.py
@@ -112,6 +112,14 @@ class Director:
 
     async def wait_experiment(self, envoy_name: str) -> str:
         """Wait an experiment."""
+        experiment_name = self.col_exp.get(envoy_name)
+        if experiment_name and experiment_name in self.experiments_registry:
+            # Experiment already set, but the envoy hasn't received experiment
+            # name (e.g. was disconnected)
+            experiment = self.experiments_registry[experiment_name]
+            if experiment.aggregator.round_number == 0:
+                return experiment_name
+
         self.col_exp[envoy_name] = None
         queue = self.col_exp_queues[envoy_name]
         experiment_name = await queue.get()
--- a/openfl/transport/grpc/director_server.py
+++ b/openfl/transport/grpc/director_server.py
@@ -197,15 +197,13 @@ class DirectorGRPCServer(director_pb2_grpc.DirectorServicer):
                     break
                 yield director_pb2.ExperimentData(size=len(data), npbytes=data)
 
-    async def WaitExperiment(self, request_iterator, context):  # NOQA:N802
+    async def WaitExperiment(self, request, context):  # NOQA:N802
         """Request for wait an experiment."""
-        logger.info('Request WaitExperiment has got!')
-        async for msg in request_iterator:
-            logger.info(msg)
-            experiment_name = await self.director.wait_experiment(msg.collaborator_name)
-            logger.info(f'Experiment {experiment_name} was prepared')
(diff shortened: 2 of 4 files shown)


openfl/component/director/director.py
```diff
@@ -112,6 +112,15 @@ def get_experiment_data(self, experiment_name: str) -> Path:
 
     async def wait_experiment(self, envoy_name: str) -> str:
         """Wait an experiment."""
+        experiment_name = self.col_exp.get(envoy_name)
+        if experiment_name and experiment_name in self.experiments_registry:
+            # Experiment already set, but the envoy hasn't received experiment
+            # name (e.g. was disconnected)
+            experiment = self.experiments_registry[experiment_name]
+            logger.info(experiment.aggregator.round_number)
```
igor-davidyuk: Should we add adescription message?

openfl/component/director/director.py
```diff
@@ -112,6 +112,15 @@ def get_experiment_data(self, experiment_name: str) -> Path:
...
+            if experiment.aggregator.round_number == 0:
+                return experiment_name
+
         self.col_exp[envoy_name] = None
```
igor-davidyuk, line 123: Should we have this field freed when an experiment is finished? Why do we clear it here?

openfl/transport/grpc/director_server.py
```diff
@@ -197,15 +197,13 @@ async def GetExperimentData(self, request, context):  # NOQA:N802
...
-        logger.info('Request WaitExperiment has got!')
-        async for msg in request_iterator:
-            logger.info(msg)
-            experiment_name = await self.director.wait_experiment(msg.collaborator_name)
-            logger.info(f'Experiment {experiment_name} was prepared')
+        logger.info(f'Request WaitExperiment has got from envoy {request.collaborator_name}!')
+        experiment_name = await self.director.wait_experiment(request.collaborator_name)
+        logger.info(f'Experiment {experiment_name} is ready for {request.collaborator_name}')
 
-            yield director_pb2.WaitExperimentResponse(experiment_name=experiment_name)
```
igor-davidyuk, line 208: What was the original thought behind this stream? If I remember correctly, our workspace archives were too big for a single message. Has something changed?

openfl/component/director/director.py
aleksandr-mokrov: Removed, it is debug information.

openfl/component/director/director.py
aleksandr-mokrov: In fact it is only for first initialization here, and doesn't required removing after experiment.

openfl/transport/grpc/director_server.py
aleksandr-mokrov: Initially yes, but now an envoy received archives using GetExperimentData. And that API doesn't need in stream.

psfoley:
Thanks @aleksandr-mokrov ! @igor-davidyuk any other changes that need to be made here? Otherwise, I think this looks good to merge.

psfoley (approved):

igor-davidyuk (approved):
ЛГТМ!

PR #624, opened 2022-11-29
Resending Shard Info if the director has lost the information

aleksandr-mokrov:
We don't handle the case, when the director lost information about ShardDescriptor (after director restart). In that case we have to restart all envoys in the federation. 
In that PR suggested:
1. A director sends in response specific error on Health Check message from unknown envoy.
2. An envoy resends Shard Info on that error in health check task.


igor-davidyuk (approved):
Looks reasonable! Well done passing the RPC error code.

openfl/transport/grpc/exceptions.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
psfoley: Copyright date should be updated

aleksandr-mokrov: Done

PR #632, opened 2022-12-05
Experimental Workflow interface

### LINKED ISSUE #627: [Experimental] Workflow Interface
Adapted from the [OpenFL Roadmap](https://github.com/intel/openfl/blob/develop/ROADMAP.md)

All interfaces in OpenFL support the standard horizontal FL training workflow today:
1. The collaborator downloads the latest model from the aggregator
2. The collaborator performs validation with their local validation dataset on the aggregated model, and sends these metrics to the aggregator (aggregated_model_validation task)
3. The collaborator trains the model on their local training data set, and sends the local model weights and metrics to the aggregator (train task)
4. The collaborator performs validation with their local validation dataset on their locally trained model, and sends their validation metrics to the aggregator (locally_tuned_model_validation task)
5. The aggregator applies an aggregation function (weighted average, FedCurv, FedProx, etc.) to the model weights, and reports the aggregate metrics.

The [Task Assigner](https://github.com/intel/openfl/blob/develop/openfl-workspace/workspace/plan/defaults/assigner.yaml#L7-L9) determines the list of collaborator tasks to be performed, and both in the task runner API as well as the interactive API these tasks can be modified (to varying degrees). For example, to perform federated evaluation of a model, only the `aggregated_model_validation` task would be listed for the assigner's block of the federated plan. Equivalently for the interactive API, this can be done by only registering a single validation task. But there are *many* other types of workflows that can't be easily represented purely by training / validation tasks performed on a collaborator with a single model. An example is training a Federated Generative Adversarial Network (GAN); because this may be represented by separate generative and discriminator models, and could leak information about a collaborator dataset, the interface we provide should allow for better control over what gets sent over the network and how. Another common request we get is for validation with an aggregator's dataset after training. Today there is not a great way to enable this in OpenFL. 

The (currently experimental) OpenFL Workflow Interface has the following goals:

1. Simplify the federated workflow representation
2. Clean separation of workflow from runtime infrastructure
4. Help users better understand the steps in federated learning (weight extraction, tensor compression, etc.)
5. Interface makes it clear what is sent across the network
6. The placement of tasks and how they connect should be straightforward
7. Don't reinvent unless absolutely necessary

psfoley:
Closes https://github.com/intel/openfl/issues/627

This PR adds:

- A new workflow interface (inspired by the [Metaflow](https://github.com/Netflix/metaflow) project) - Define a chain of tasks to be performed on aggregator or collaborator. 
- Runtimes - A runtime defines the infrastructure where the workflow will be executed. 
    - [x] LocalRuntime - Run the workflow on a single node
        - Two backends supported: Execute a flow in a **single process** (default) or in multiple processes (for concurrent execution) with **[Ray](https://github.com/ray-project/ray)**
    - Additional runtimes supporting multi-node execution to be made available in a future release
- Partial compatibility with Metaflow Client for enhanced federated workflow debugging (note: all of these are optional)
    - [x] Save Data Artifacts
    - [x] Save stdout / stderr
    - [x] Generate Workflow DAG that can be loaded in HTML (only currently compatible in flows without loops)
- New Use Cases / Tutorials under the `openfl-tutorials/experimental` folder:
    - [x] Aggregator Validation after Collaborator Training
    - [x] Model Watermarking
    - [x] Privacy Meter - Measuring information leakage after model training
    - [x] Global Differential Privacy
    - [x] Vertical FL Examples
    - [x] Cyclic Institutional Incremental Training 

Remaining items before removing WIP tag:
- [x] Refactor logic in `LocalRuntime`
- [x] Merge PRs that ensure [no leakage between tasks](https://github.com/psfoley/openfl/pull/17)
- [x] Selective CI Pipelines for PRs labelled with `workflow_interface`
- [x] Comprehensive documentation
- [x] Review and update licenses

This PR includes contributions from:
@mansishr, @anindya-paul, @brandon-edwards, @changhongyan123, @scngupta-dsp, @ParthM-GitHub, @Greesh-N, @KeertiX, @igor-davidyuk, @aleksandr-mokrov, @itrushkin, @msheller, @hanboa


openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
```diff
@@ -0,0 +1,444 @@
...
+                self.next(self.aggregated_model_validation, foreach='round_collaborators', exclude=['private'])
+            else:
+                print(f'No collaborator selected for training at Round: {self.round}')
+                self.next(self.check_round_completion)
+        else:
+            print(f'No collaborator selected for training at Round: {self.round}')
+            self.next(self.check_round_completion)
+
+    # @collaborator                     # Uncomment this if you don't have GPU in the machine and want this application ro run on CPU instead 
+    @collaborator(num_gpus=1)           # Assuming GPU(s) is available in the machine
```
mansishr: Somehow Ray remote is not being invoked, all processes are going to the first cuda device.

anindya-paul: We set `backend='single_process'` by default as the Ray backend requires more system memory than the `single_process`, so we recommend ` backend='ray` on high-memory servers and systems.  The following will  demonstrate how to enable Ray backend through `LocalRuntime` if one wishes to experiment with Ray:
`local_runtime = LocalRuntime(aggregator=aggregator, collaborators=collaborators, backend='ray')`

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
```diff
@@ -0,0 +1,444 @@
...
+    aggregator = Aggregator()
+    aggregator.private_attributes = {}
+
+    # Setup collaborators with private attributes
+    collaborator_names = ['Portland', 'Seattle', 'Bangalore', 'Chandler', 'Austin', 
+                            'Hudson', 'Penang', 'Haifa', 'CostaRica', 'Guadalajara']
+    collaborators = [Collaborator(name=name) for name in collaborator_names]
+    
+    if torch.cuda.is_available():
+        device = torch.device(f'cuda:0')  # This will enable Ray library to reserve available GPU(s) for the task
```
mansishr, line 413: Devices should be maintained per collaborator. Right now, all processes are going to cuda 0.

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
psfoley: @mansishr You're correct that in the case of the `single_process` backend for the `LocalRuntime`, all models will be sent to the same device. This should be okay because with the `single_process` backend, collaborators execute sequentially so only one GPU should be used at a time for this example. 

If there are multiple GPUs and sufficient memory in the system to take advantage of concurrent collaborator training, the Ray backend (enabled by setting `local_runtime = LocalRuntime(aggregator=aggregator, collaborators=collaborators, backend='ray')`) will automatically set  `CUDA_VISIBLE_DEVICES` within the task, and each collaborator will be able to use their own dedicated GPU without the user needing to make any change to the workflow definition. 

It's worth noting though that setting `num_gpus=1` is only honored by the Ray backend - the `single_process` backend will just ignore this request. I'll add a warning / info message when these placement decorator arguments are ignored for backends where they are not supported

openfl/experimental/runtime/local_runtime.py
```diff
@@ -0,0 +1,177 @@
...
+    def __init__(
+        self,
+        aggregator=None,
+        collaborators=None,
+        backend="single_process",
+        **kwargs,
+    ):
+        """Use Local infrastructure to run the flow"""
+        super().__init__()
+        if backend not in ["ray", "single_process"]:
```
igor-davidyuk, line 61: In my understanding, `ray` and `single_process` backends may be classified as separate runtimes. Does it make sense to have separate  `RayLocalRuntime` and `SPLocalRuntime` classes both implementing the `Runtime` interface?

openfl/experimental/runtime/runtime.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) 2020-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+""" openfl.experimental.runtime module Runtime class."""
+
+
+class Runtime:
```
igor-davidyuk, line 13: The specification for this component needs elaboration.

openfl/experimental/runtime/local_runtime.py
psfoley: @igor-davidyuk I'm glad you raised this. My intention was to create a hierarchy and logically divide `Runtime` to define where the workflow would run, and the `Backend` would specify how it would be executed. That way in the future there it will be clear if the user selects `LocalRuntime` the workflow will be executed locally, and `FederatedRuntime` runs the workflow across multiple nodes, and the user doesn't need to think about the backend unless they have good reason to do so. @mansishr, @aleksandr-mokrov, @itrushkin what are your thoughts about this?

openfl/experimental/runtime/runtime.py
psfoley: Added base class methods

docs/flow_interface.rst
```diff
@@ -0,0 +1,287 @@
...
+================
+
+The Runtime defines where code will run, but the Runtime has a :code:`Backend` - which defines the underlying implementation of *how* the flow will be executed. :code:`'single_process'` is the default in the :code:`LocalRuntime`: it executes all code sequentially within a single python process, and is well suited to run both on high spec and low spec hardware. For users with large servers or multiple GPUs they wish to take advantage of, we also provide a `Ray <https://github.com/ray-project/ray>` backend. The Ray backend enables parallel task execution for collaborators, and optionally allows users to request dedicated GPUs for collaborator tasks in the placement decorator, as follows:
+
+.. code-block:: python
+    
+    ExampleDedicatedGPUFlow(FLSpec):
+        ...
+        # We request one dedicated GPU for this task 
+        @collaborator(num_gpus=1)
```
MasterSkepticista: What happens if `num_gpus` is set >1? Does the user have to write distributed dataparallel primitives? 
How are those devices enumerated to each process?
Alternatively, if the argument specifies `request_dedicated_gpu=True`, it would add clarity as to whether or not a user needs to write dataparallel primitives.

openfl-tutorials/experimental/Global_DP/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Global DP implementation using the Opacus library (https://opacus.ai/) to perform global differentially private federated learning. At each round, a subset of collaborators are selected using a Poisson distribution over all collaborators, the selected collaborators perform local training with periodic clipping of their model delta (with respect to the current global model) to bound their contribution to the average of local model updates. Gaussian noise is then added to the average of these local models at the aggregator. The result of this federated training is then differentially private with respect to the inclusion (or exclusion) of any one collaborator into the training. A Renyi differential privacy (RDP) accountant is used at the aggregator to determine the epsilon, delta for which the final model is epsilon, delta differentially private with respect to inclusion or exclusion of any one collaborator. This tutorial has two implementations which are listed below. The two implementations are statistically equivalent, but incorporate Opacus interfaces at two different levels. Both implementations use the same code for local training and periodic clipping.
```
MasterSkepticista: Nitpick: The first sentence might need refinement, plus a line or two on what DP is, when does it help.

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
```diff
@@ -0,0 +1,654 @@
...
+batch_size_train = 64
+batch_size_test = 1000
+learning_rate = 0.01
+momentum = 0.5
+log_interval = 10
+
+# Fixing the seed for result repeatation: remove below to stop repeatable runs
+# ----------------------------------
+random_seed = 5495300300540669060
+torch.manual_seed(random_seed)
```
MasterSkepticista: A more reliable way is to use local-scope RNGs. 
The current method fixes seed globally for the entire script and modules called thereafter. If certain modules change, they affect the state of all other modules later (even if repetitive states are guaranteed)
Details [here](https://pytorch.org/docs/stable/generated/torch.Generator.html#torch.Generator).

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
```diff
@@ -0,0 +1,654 @@
...
+        return F.log_softmax(x)
+
+
+def default_optimizer(model):
+    # indicate the default optimizer for training the target model (we have
+    # tested with SGD w/ momentum)
+    return optim.SGD(model.parameters(), lr=learning_rate, momentum=momentum)
+
+
+def FedAvg(models, previous_global_model=None, dp_params=None):
```
MasterSkepticista: General: Typehints and docstrings are helpful for end-users to write code without referring to the documentation.

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
```diff
@@ -0,0 +1,654 @@
...
+        self.loss = np.mean(train_losses)
+        self.training_completed = True
+
+        tmp_opt = deepcopy(self.optimizers[self.input])
+        tmp_opt.load_state_dict(self.optimizer.state_dict())
+        self.optimizer = tmp_opt
+        torch.cuda.empty_cache()
+        self.next(self.local_model_validation)
+
+    # Uncomment this if you don't have GPU in the machine
```
MasterSkepticista, line 494: Another possibility where different flag naming helps.
`request_dedicated_gpu=True` vs `num_gpus=1` leaves the choice to the runtime. `CPU` is allocated if `GPU` is unavailable (with a warning since user requested one)

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
```diff
@@ -0,0 +1,654 @@
...
+            clip_freq=self.dp_params["clip_frequency"],
+        )
+
+        # copy state dict from optimizer object from previous round to the newly
+        # instantiated optimizer for the same collaborator
+        if self.round > 0:
+            self.optimizer.load_state_dict(
+                deepcopy(self.optimizers[self.input].state_dict())
+            )
+            optimizer_to_device(optimizer=self.optimizer, device=self.device)
```
MasterSkepticista: Would it be more efficient to push the params to device from `state_dict()`, before calling `self.optimizer.load_state_dict(...)`?

openfl-tutorials/experimental/Global_DP/requirements_global_dp.txt
```diff
@@ -0,0 +1,8 @@
+numpy==1.23.3
```
MasterSkepticista, line 1: Strictly equal version requirements are primary causes of dependency loops. They work today, but not a few months down the line.
I suggest relaxing requirements for generic packages like `numpy`, `matplotlib` and others, wherever possible, to reduce conflicts when users install their own packages.

MasterSkepticista (commented):
Partial feedback on code reviewed so far:
- Please use typehints across functions, unless the function's behaviour is implicit or obvious. Typehints are a documentation in disguise. 
- Most functions have missing docstrings (by that, I mean an account of accepted values of an argument, default values etc., not just a literal expansion of what the function does)
- Given this new interface and growth of this repository, it would make sense to "consider" moving examples to a separate repository, and tag releases that are supposed to work for examples.

openfl/experimental/runtime/local_runtime.py
igor-davidyuk: Suppose we want to differentiate between runtimes and backends. In that case, we need a `backend` interface class to encapsulate the backend methods. 
`if else` approach is not sustainable if we would add more backends.

igor-davidyuk:
It is nice the PR comes with a vertical FL example.
I am wondering if it would be possible to implement cyclic training and cross-site validation schemes with FL Flow interface.

openfl/experimental/runtime/local_runtime.py
mansishr: I like the idea of dividing `Runtime` to define where the flow would run, making it clear for the user. About `Backend` being an interface class, for now we only support `Ray` so using `if else` looks okay to me.

docs/flow_interface.rst
mansishr: IMO, in that case, the workload is launched on a single GPU while other GPUs remain idle. According to [ray documentation](https://docs.ray.io/en/latest/ray-core/tasks/using-ray-with-gpus.html), calling `ray.remote()` internally calls `ray.get_gpu_ids()` which returns a list of available GPU IDs.

openfl/experimental/interface/fl_spec.py
```diff
@@ -0,0 +1,178 @@
...
+        """Takes backup of self before exclude or include filtering"""
+        return_objs = []
+        if "exclude" in kwargs:
+            exclude_backup = deepcopy(self)
+            return_objs.append(exclude_backup)
+        if "include" in kwargs:
+            include_backup = deepcopy(self)
+            return_objs.append(include_backup)
+        return return_objs
+
```
mansishr, line 121: Since we are not using `exclude_backup` and `include_backup` objects differently, could this be condensed to:

```
if "exclude" in kwargs or "include" in kwargs:
    return_objs = deepcopy(self)
return return_objs
```

openfl-tutorials/experimental/Global_DP/README.md
psfoley: @mansishr could you address this?

openfl/experimental/interface/fl_spec.py
psfoley: Fixed

docs/flow_interface.rst
psfoley: @MasterSkepticista `num_gpus=1` will likely be the most common usage, but there are other cases where using the functionality as implemented in ray will be useful. Setting  `num_gpus=0.5` allows two processes to run simultaneously on a single GPU. Users should be able to tune this parameter based on the size of their model / details of their workflow. One of the other possibilities is [an experiment that involves an ensemble of models](https://gitlab.com/frontierdevelopmentlab/astronaut-health/crisp), and by setting `num_gpus=2` the user could send different models to different GPUs. @mansishr mentioned one of the ways that device ids can be enumerated, but the other way is using the standard `os.environ["CUDA_VISIBLE_DEVICES"]`

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
mansishr: Won't make much of a difference for the MNIST example but @anindya-paul what do you think?

openfl-tutorials/experimental/Global_DP/requirements_global_dp.txt
mansishr: I agree, maybe we could change it to `numpy>=1.23.3` to use latest versions. @anindya-paul what do you think?

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
mansishr: Fixed

openfl-tutorials/experimental/Global_DP/requirements_global_dp.txt
anindya-paul: I have seen sometimes the latest version of NumPy causes issues. For example, the Privacy Meter example breaks with numpy==1.24.0 and hence we had to fix the Numpy version requirement there. In my experience, Numpy is one of the worst offenders in breaking interfaces and also sometimes causing confusing errors, so I'd suggest fixing the version that we have tested and knowing that it works with the release.

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
anindya-paul: It doesn't matter in this case as `self.optimizers` dictionary is initialized to CPU and hence state of `self.optimizers[self.input]` needs to move from CPU to GPU even if we load `self.optimizer` state to GPU first.

docs/flow_interface.rst
anindya-paul: @psfoley: following along the line what @MasterSkepticista asked: if a user enables PyTorch "[DISTRIBUTEDDATAPARALLEL](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html)" primitive for multi-GPU model training, will that cause conflict with Ray distributed orchestration?

openfl-tutorials/experimental/Global_DP/Workflow_Interface_Mnist_Implementation_1.py
psfoley: See the point above about keeping `num_gpus=` for consistency with Ray (and use cases where varying the number of requested GPUs is needed). I like the idea of fallback to CPU in the case where none are available - will add this as an issue to be addressed in the future

psfoley:
> Partial feedback on code reviewed so far:
> 
> * Please use typehints across functions, unless the function's behaviour is implicit or obvious. Typehints are a documentation in disguise.
> * Most functions have missing docstrings (by that, I mean an account of accepted values of an argument, default values etc., not just a literal expansion of what the function does)
> * Given this new interface and growth of this repository, it would make sense to "consider" moving examples to a separate repository, and tag releases that are supposed to work for examples.

@MasterSkepticista thanks for the feedback! I've added type hints and expanded docstrings for user facing functions. @anindya-paul is working on some of the other comments specific to Global DP and Privacy Meter.

psfoley:
> It is nice the PR comes with a vertical FL example. I am wondering if it would be possible to implement cyclic training and cross-site validation schemes with FL Flow interface.

@igor-davidyuk Excellent idea. I've added a cyclic training example notebook to this PR here that compares that approach with standard horizontal FL under various IID / Non-IID scenarios: https://github.com/intel/openfl/blob/554f674b71b7bc5346e65a98752d4959210492eb/openfl-tutorials/experimental/Workflow_Interface_103_Cyclic_Institutional_Incremental_Learning.ipynb

openfl/experimental/runtime/local_runtime.py
psfoley: @igor-davidyuk and I discussed this offline. I agree a formal interface for the backend is more sustainable. I've added an issue for this to be addressed in the next release: https://github.com/intel/openfl/issues/676

openfl/experimental/placement/placement.py
```diff
@@ -0,0 +1,151 @@
...
+from typing import Callable
+
+
+class RayExecutor:
+    def __init__(self):
+        self.remote_functions = []
+        self.remote_contexts = []
+
+    def ray_call_put(self, ctx, func):
+        remote_to_exec = make_remote(func, num_gpus=func.num_gpus)
```
mansishr, line 21: Docstrings can be added here

openfl/experimental/placement/placement.py
```diff
@@ -0,0 +1,151 @@
...
+        ref_ctx = ray.put(ctx)
+        self.remote_contexts.append(ref_ctx)
+        self.remote_functions.append(remote_to_exec.remote(ref_ctx, func.__name__))
+        del remote_to_exec
+        del ref_ctx
+
+    def get_remote_clones(self):
+        clones = deepcopy(ray.get(self.remote_functions))
```
mansishr, line 29: add docstrings

tests/github/experimental/testflow_reference.py
```diff
@@ -0,0 +1,369 @@
...
+                ] = matched_ref_dict.get(collab)
+
+    if not reference_flag:
+        print(
+            f"{bcolors.OKGREEN}  Pass : Reference test passed for collaborators. {bcolors.ENDC}"
+        )
+
+
+def validate_agg_attr_ref(agg_attrs, agg_obj):
+    attr_flag = False
```
mansishr, line 292: add docstring.

tests/github/experimental/testflow_reference_with_exclude.py
```diff
@@ -0,0 +1,335 @@
...
+        if attr_collab_dict.get(collab.input):
+            attr_collab_dict.get(collab.input)[attr] = attr_id
+        else:
+            attr_collab_dict[collab.input] = {}
+            attr_collab_dict.get(collab.input)[attr] = attr_id
+    return attr_collab_dict, collab_attr_list
+
+
+def find_match_ref_at_step(collab_attr_list, all_collborators):
+    collab_names = all_collborators.keys()
```
mansishr, line 288: add docstring.

tests/github/experimental/testflow_reference.py
psfoley: Added

tests/github/experimental/testflow_reference_with_exclude.py
psfoley: Added

mansishr (approved):
LGTM, great work!

Issue #638, opened 2022-12-11
Introduce `development mode` to `fx workspace graminize` command


igor-davidyuk:
Add the option to install OpenFL from a local repo or from a specific branch in the gramine-ready docker image.
More context in https://github.com/intel/openfl/pull/570 comment section.

teoparvanov:
This has been resolved with https://github.com/securefederatedai/openfl/pull/1140

PR #639, opened 2022-12-11
Add development mode to graminize command

### LINKED ISSUE #554: "fx workspace graminize" does not regard the version used to build the workspace

itrushkin:
## New Features
 -  `--openfl-path` option to pass to `fx workspace graminize` command. Allows both local and remote OpenFL repository paths.
   1. Local path
   Accepts every path on the system that exists.
   2. Remote path
   Accepts the URL matching repository regular expression. 

 -  New Graminize workflow in GitHub actions (using local repo from the corresponding PR).
   Previously used PyTorch UNet template could not be handled by a GitHub-hosted runner due to memory restrictions. Decreasing of batch size and dataset size does not help, thus I switched the Graminize test to `torch_cnn_histology` template.

closes: #554 #638


openfl/interface/workspace.py
```diff
@@ -450,9 +450,12 @@ def dockerize_(context, base_image, save):
         default=True, type=bool,
         help='Dump the Docker image to an archive')
 @option('--rebuild', help='Build images with `--no-cache`', is_flag=True)
+@option('--developer-mode', required=False,
+        default=False, type=bool, is_flag=True,
+        help='Use the local OpenFL package in the Docker image')
 @pass_context
 def graminize_(context, signing_key: Path, enclave_size: str, pip_install_options: Tuple[str],
-               save: bool, rebuild: bool) -> None:
+               save: bool, rebuild: bool, developer_mode: bool) -> None:
```
igor-davidyuk: `COPY . .` in the Dockerfile will not work for this command. The reason is `fx workspace graminize` command, as well as `fx workspace dockerize` are designed to be called from an FL workspace, not OpenFL repo.
I imagine the additional parameter to be either a local path to OpenFL repo or a URL for a remote repo.
Something like `openfl_source: Union[Path, str]`, with additional parameter `branch` for the remote option

itrushkin: Got it. We could do this with 1 parameter: the branch could be specified in the URL itself.

rahulga1:

MasterSkepticista:
Gramine-based TEE support has been revamped (#1140), per latest requirements, closing this PR.

PR #648, opened 2022-12-14
Write keys with restricted permissions (#637)

### LINKED ISSUE #637: Generated aggregator and collaborator private keys are stored with world-readable permissions
**Describe the bug**

`write_key` function write private keys to disk with world-readable permissions. Consequently, keys generated by `generate-cert-request` (both on aggregator and collaborator sides) are in higher risk of being disclosed

**To Reproduce**

1. Create a new workspace
2. On the aggregator side, generate new keypair using `fx aggregator generate-cert-request`
3. On the collaborator side, generate new keypair using `fx collaborator generate-cert-request`

**Expected behavior**

Generated private key is saved with no access to anyone but the user

DL8:
Prior to this fix, keys generated by "fx" command line were saved with default permissions. Depending on the current OS configuration, this may include read permissions to other users.
To fix this, write_key() function was modified such that it will write the key with no permissions to other users.

Fixes #637

### DIFF
--- a/openfl/cryptography/io.py
+++ b/openfl/cryptography/io.py
@@ -3,6 +3,7 @@
 
 """Cryptography IO utilities."""
 
+import os
 from hashlib import sha384
 from pathlib import Path
 from typing import Tuple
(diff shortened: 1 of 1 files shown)


psfoley (approved):
Thank you for fixing this, @DL8!

PR #649, opened 2022-12-14
Fix TensorFlow version comparison

itrushkin:
## Issue
We have a logic that makes Keras models picklable for TensorFlow versions less than 2.7 where picklability is included. Starting with TensorFlow 2.10, string comparison with `'2.7.1'` does not work because `'2.10.1' < '2.7.1'`.
## Fix
Use `packaging.version` module to compare versions properly.


mansishr (approved):
Looks good to me, thanks Ilya!

aleksandr-mokrov (approved):

PR #652, opened 2022-12-15
Added HPU adaptations for Pytorch_MedMNIST_2D

Supriya-Krishnamurthi:
Added HPU adaptations for Pytorch_MedMNIST_2D and removed the depricated API load_habana_module

### DIFF
--- a/openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/README.md
+++ b/openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/README.md
@@ -1,0 +1,181 @@
+# MedMNIST 2D Classification Tutorial
+
+![MedMNISTv2_overview](https://raw.githubusercontent.com/MedMNIST/MedMNIST/main/assets/medmnistv2.jpg)
+
+For more details, please refer to the original paper:
+**MedMNIST v2: A Large-Scale Lightweight Benchmark for 2D and 3D Biomedical Image Classification** ([arXiv](https://arxiv.org/abs/2110.14795)), and [PyPI](https://pypi.org/project/medmnist/).
+
+
+## **Habana Tutorials**
+
+## **About model and experiments**
+
+We use a simple convolutional neural network and settings coming from [the experiments](https://github.com/MedMNIST/experiments) repository.
+<br/>
+
+#### The name of the file/example that contain HPU adaptations start with "HPU". 
+For example: HPU_PyTorch_MedMNIST_2D.ipynb placed under workspace folder contains the required HPU adaptations.
+
+ All the execution steps mention in last section (**V. How to run this tutorial**) remain same for HPU examples but as pre-requisite it needs some additional environment setup and Habana supported package installations which is explained below from **section I to IV**.
+
+ **Note:** By default these experiments utilize only 1 HPU device
+
+ <br/>
+
+ ## **I. AWS DL1 Instance Setup**
+
+ This example was tested on AWS EC2 instance created by following the instructions mentioned [here](https://docs.habana.ai/en/latest/AWS_EC2_DL1_and_PyTorch_Quick_Start/AWS_EC2_DL1_and_PyTorch_Quick_Start.html) . 
+ 
+ Test setup - Habana 1.7 and Ubuntu 20.04
+
+ <br/>
+
+ ## **II. Set Up SynapseAI SW Stack**
+
+ - To perform an installation of the full driver and SynapseAI software stack independently on the EC2 instance, run the following command:
+ 
+ ```
+ wget -nv https://vault.habana.ai/artifactory/gaudi-installer/latest/habanalabs-installer.sh
+chmod +x habanalabs-installer.sh
+./habanalabs-installer.sh install --type base
+```
+You can refer the [Habana docs](https://docs.habana.ai/en/latest/Installation_Guide/Bare_Metal_Fresh_OS.html#set-up-synapseai-sw-stack) mentioned [GitHub repository](https://github.com/HabanaAI/Setup_and_Install) for detailed instructions.
+
+<br/>
+
+ ## **III. HPU Pytorch Installation**
+
+ For this example make sure to install the PyTorch package provided by Habana. These packages are optimized for Habana Gaudi HPU. Installing public PyTorch packages is not supported.
+ Habana PyTorch packages consist of:
+ - **torch** - PyTorch framework package with Habana support
+
+- **habana-torch-plugin** - Libraries and modules needed to execute PyTorch on single card, single node and multi node setup.
+
+- **habana-torch-dataloader** - Habana multi-threaded dataloader package.
+
+- **torchvision** - Torchvision package compiled in torch environment. No Habana specific changes in this package. 
+ 
+ Run the following command to install the above Habana PyTorch environment
+
+ ```
+ ./habanalabs-installer.sh install --type pytorch --venv
+ ``` 
+
+ The -- venv flag installs the relevant framework inside the virtual environment. To activate a virtual environment please perform the following:
+ 
+ 
+ ```
+  cd $HOME/habanalabs-venv
+  source ./bin/activate
+ ```
+
+The default virtual environment folder is $HOME/habanalabs-venv. To override the default, run the following command:
+
+ ```
+ export HABANALABS_VIRTUAL_DIR=/path/to/dir
+ ```
+
+ You can refer the [Habana docs](https://docs.habana.ai/en/latest/Installation_Guide/Bare_Metal_Fresh_OS.html#set-up-synapseai-sw-stack) mentioned [GitHub repository](https://github.com/HabanaAI/Setup_and_Install) for detailed instructions.
+
+ </br>
+
+ ## **IV. HPU Adaptations For PyTorch Examples**
+
+The following set of code additions are required in the workspace notebook to run a model on Habana. The following steps cover Eager and Lazy modes of execution.
+
+### 1. Target the Gaudi HPU device:
+
+```
+device = torch.device("hpu")
+```
+### 2. Move the model to the device:
+
+**There is a dependency in the order of execution (moving model to HPU and intializing optimizer). The workaround is to execute this step before initializing any optimizers.**
+
+```
+model.to(device)
+```
+### 3. Import the Habana Torch Library:
+
+```
+import habana_frameworks.torch.core as htcore
+```
+### 4. Enable Lazy execution mode by setting the environment variable shown below. 
+Do not set this environment variable if you want to execute your code in Eager mode:
+
+```
+os.environ["PT_HPU_LAZY_MODE"] = "1"
+```
+### 5. In Lazy mode, execution is triggered wherever data is read back to the host from the Habana device. 
+For example, execution is triggered if you are running a topology and getting loss value into the host from the device with `loss.item()`. Adding a `mark_step()` in the code is another way to trigger execution:
+
+```
+htcore.mark_step()
+```
+
+The placement of `mark_step()` is required at the following points in a training script:
+
+* Right after `optimizer.step()` to cleanly demarcate training iterations,
+* Between `loss.backward` and `optimizer.step()` if the optimizer being used is a Habana custom optimizer.
+
+Refer [getting started with PyTorch](https://www.intel.com/content/www/us/en/developer/articles/technical/get-started-habana-gaudi-deep-learning-training.html#articleparagraph_cop_711677074) for detailed explaination and PyTorch Habana architecture. Sample example can be found [here](https://docs.habana.ai/en/latest/PyTorch/Getting_Started_with_PyTorch_and_Gaudi/Getting_Started_with_PyTorch.html)
+
+<br/>
+
+## **V. How to run this tutorial (without TLC and locally as a simulation):**
+### 0. If you haven't done so already, create a virtual environment, install OpenFL, and upgrade pip:
+  - For help with this step, visit the "Install the Package" section of the [OpenFL installation instructions](https://openfl.readthedocs.io/en/latest/install.html#install-the-package).
+<br/>
+ 
+### 1. Split terminal into 3 (1 terminal for the director, 1 for the envoy, and 1 for the experiment)
+<br/> 
+
+### 2. Do the following in each terminal:
+   - Activate the virtual environment from step 0:
+   
+   ```sh
+   source venv/bin/activate
+   ```
+   - If you are in a network environment with a proxy, ensure proxy environment variables are set in each of your terminals.
+   - Navigate to the tutorial:
+    
+   ```sh
+   cd openfl/openfl-tutorials/interactive_api/PyTorch_MedMNIST_2D
+   ```
+<br/>
+
+### 3. In the first terminal, run the director:
+
+```sh
+cd director
+./start_director.sh
+```
+<br/>
+
+### 4. In the second terminal, install requirements and run the envoy:
+
+```sh
+cd envoy
+pip install -r requirements.txt
+./start_envoy.sh env_one envoy_config.yaml
+```
+
+Optional: Run a second envoy in an additional terminal:
+  - Ensure step 2 is complete for this terminal as well.
+  - Run the second envoy:
+```sh
+cd envoy
+./start_envoy.sh env_two envoy_config.yaml
+```
+<br/>
+
+### 5. In the third terminal (or forth terminal, if you chose to do two envoys) run the Jupyter Notebook:
+
+```sh
+cd workspace
+jupyter lab Pytorch_MedMNIST_2D.ipynb
+```
+- A Jupyter Server URL will appear in your terminal. In your browser, proceed to that link. Once the webpage loads, click on the Pytorch_MedMNIST_2D.ipynb file. 
+- To run the experiment, select the icon that looks like two triangles to "Restart Kernel and Run All Cells". 
+- You will notice activity in your terminals as the experiments runs, and when the experiment is finished the director terminal will display a message that the experiment was finished successfully.  
+ 
--- a/openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/envoy/envoy_config.yaml
+++ b/openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/envoy/envoy_config.yaml
@@ -1,0 +1,10 @@
+params:
+
+optional_plugin_components: {}
+
+shard_descriptor:
+  template: medmnist_shard_descriptor.MedMNISTShardDescriptor
+  params:
+    rank_worldsize: 1, 1
+    datapath: data/.
+    dataname: bloodmnist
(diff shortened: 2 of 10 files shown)


openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The following command zips the workspace and python requirements to be transfered to collaborator nodes\n",
+    "fl_experiment.start(model_provider=MI, \n",
+    "                    task_keeper=TI,\n",
+    "                    data_loader=fed_dataset,\n",
+    "                    rounds_to_train=3,\n",
+    "                    opt_treatment='RESET',\n",
+    "                    device_assignment_policy='CUDA_PREFERRED')"
```
igor-davidyuk: The `envoy_config` does not contain any GPU devices registered, therefore this line does not make any difference and may be removed

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
```diff
@@ -0,0 +1,627 @@
...
+    "\n",
+    "val_custom_params={'criterion':criterion, 'task':task, 'device':'hpu'}\n",
+    "\n",
+    "@TI.add_kwargs(**val_custom_params)\n",
+    "@TI.register_fl_task(model='model', data_loader='val_loader', device='device')\n",
+    "def validate(model, val_loader, device, criterion, task):\n",
+    "\n",
+    "    val_loader = tqdm.tqdm(val_loader, desc=\"validate\")\n",
+    "    model.eval()\n",
+    "    model.to(device)\n",
```
igor-davidyuk, line 477: What is the device name for HPU? It is not obvious how tensors end up in HPU memory

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/README.md
```diff
@@ -0,0 +1,179 @@
+# MedMNIST 2D Classification Tutorial
```
MasterSkepticista, line 1: It is essential to note in all HPU notebooks so far, that these experiments utilize only 1 HPU device. Habana instances are not provided as single-device nodes, hence to avoid surprise when users see only 1 device being utilized for compute.

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/README.md
```diff
@@ -0,0 +1,179 @@
...
+import habana_frameworks.torch.core as htcore
+```
+### 4. Enable Lazy execution mode by setting the environment variable shown below. 
+Do not set this environment variable if you want to execute your code in Eager mode:
+
+```
+os.environ["PT_HPU_LAZY_MODE"] = "1"
+```
+### 5. In Lazy mode, execution is triggered wherever data is read back to the host from the Habana device. 
+For example, execution is triggered if you are running a topology and getting loss value into the host from the device with loss.item(). Adding a mark_step() in the code is another way to trigger execution:
```
MasterSkepticista: Across all markdown documents, encapsulate all code snippets, package names, or flags used mid-sentences with single backticks. 

Without ticks:
> Execution is triggered if you are running a topology and getting loss value into the host from the device with loss.item(). Adding a mark_step() in the code is another way to trigger execution

With ticks:
> Execution is triggered if you are running a topology and getting loss value into the host from the device with `loss.item()`. Adding a `mark_step()` in the code is another way to trigger execution

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/envoy/medmnist_shard_descriptor.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
MasterSkepticista: Update header with 2023

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/envoy/envoy_config.yaml
```diff
@@ -0,0 +1,11 @@
+params:
+  cuda_devices: []
```
MasterSkepticista: This can be removed.

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
```diff
@@ -0,0 +1,627 @@
...
+    "                targets = targets.to(torch.float32).to(device)\n",
+    "                loss = criterion(outputs, targets)\n",
+    "                m = nn.Sigmoid()\n",
+    "                outputs = m(outputs).to(device)\n",
+    "            else:\n",
+    "                targets = torch.squeeze(targets, 1).long().to(device)\n",
+    "                loss = criterion(outputs, targets) \n",
+    "                m = nn.Softmax(dim=1)\n",
+    "                outputs = m(outputs).to(device)\n",
+    "                targets = targets.to(torch.float32)\n",
```
MasterSkepticista:
```python
targets = targets.to(torch.float32)
# targets = targets.float().resize_(len(targets), 1)
# right way to use float() is tensor.to(torch.float32) else gives RuntimeError: expected scalar type Long but found Float
# The resize_(len(targets), 1) operation in above line causes below graph compile error, hence removed it
# RuntimeError: FATAL ERROR :: MODULE:BRIDGE syn compile encountered : Graph compile failed
```
1. Why does it need to be cast to float? AFAIK, `nn.CrossentropyLoss()` works on long tensors.
2. From the comment block, it is not clear what `resize_` is supposed to do, other than the fact that it throws an exception. `resize_` manipulates tensor shape. Alternatively, if `resize_` is not required, which other function is taking its role?

Overall, I think both the code and the comment has the potential to create unnecessary confusion. Generally, a combination of below statements is used without any explicit typecasts:
```python
inputs = inputs.to(device)  # inputs = [bs, H, W, C] or [bs, C, H, W] both are equivalent
targets = targets.to(device)  # targets = [bs] a.k.a 1D array of long integers

# Forward pass
optimizer.zero_grad()
outputs = model(inputs)  # outputs = [bs, num_classes]
pred = outputs.argmax(1)  # pred = [bs] of long integers, same as targets
loss = criterion(outputs, targets)  # loss = [bs] of floats

# Backward and step
loss.backward()
optimizer.step()  # And habana-specific mitigation if any.

# Log
acc += pred.eq(targets).sum().cpu().numpy()
```

MasterSkepticista (changes requested):

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: Hi @shreyasnc1, I also had this query, casting to float and resize_() are not done in training part, it can be seen only in validation. These two operations do not have any impact on the accuracy. Do we need to keep this?

The comment was added just to bring it to the notice that these lines were part of original code and that resize if kept is causing that error

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: i] To get the details about HPU device like HPU device name, index number we need to import below package:
import habana_frameworks.torch.hpu as hthpu

The output of the below API's for current example is highlighted below:

hthpu.get_device_name() - Returns the name of the HPU device --> **'GAUDI'**
hthpu.current_device() - Returns the index of the current selected HPU device --> **0**

ii] First we initialize `device = 'hpu'` , then move Model and Tensors to device('HPU') memory using `model.to(device)` , `input.to(device)`, `targets.to(device)` etc

We can verify this by checking the device that the tensor is present in before and after moving to HPU device by using `tensor.device`, ex: `input.device`, `target.device` . 

o/p before moving to HPU: `device(type='cpu')`
o/p after moving to HPU: `device(type='hpu', index=0)`

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: @MasterSkepticista @shreyasnc1 @acharyasunil The actual issue with resize_() operation mentioned in the original code was with the way the arguments are passed to it. For ex: The target was initially in shape `([128,1])` which is 2D. But we need target to be in 1D. `resize_(len(targets), 1)` mentioned in code does the opposite, it resizes the tensor of length 128 to 2D `([128,1])`.To resize it to 1D we need to pass argument as `resize_(len(targets), )`. So the working snippet after correction: `targets.to(torch.float32).resize_(len(targets),)`

This resolves the issue but still casting to float and resizing is not needed as the computations works on `long` tensors and the squeeze operation has already reduced target dimension from 2D to 1D `targets = torch.squeeze(targets, 1).long().to(device)` . Let me know if we need to keep these float and resize operations or not

**Note:** Even on CPU  `resize_(len(targets), 1)` gives dummy accuracy score like ex: 19 or 20
But with `resize_(len(targets), )` gives valid accuarcy score like ex: 83.66 or 86.08

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
igor-davidyuk: But the `device` variable will be set to `cpu` by the Collaborator in this method, this is why I ask

Supriya-Krishnamurthi: We have set `device='hpu'` in custom_params which are passed to add_kwargs() before moving model or tensors to device

`train_custom_params={'criterion':criterion,'task':task, 'device':'hpu'}`
`val_custom_params={'criterion':criterion, 'task':task, 'device':'hpu'}`

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
sun1lach: `reshape` or `view` are a better alternative. 
Please check for the correctness of the train and validation loop by comparing previous with the modified/removed `resize_` or `squeeze` operations.

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/envoy/medmnist_shard_descriptor.py
```diff
@@ -0,0 +1,128 @@
...
+        """Return the target shape info."""
+        return ['1', '1']
+
+    @property
+    def dataset_description(self) -> str:
+        """Return the dataset description."""
+        return (f'MedMNIST dataset, shard number {self.rank}'
+                f' out of {self.worldsize}')
+
+    def download_data(datapath: str = 'data/',
```
sun1lach, line 94: I believe this method needs a `@staticmethod` decorator for it to be called as `MedMNISTShardDescriptor.download_data(...)` in line 117

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
igor-davidyuk: That is an interesting case of overwriting the task contract parameters. It is not obvious which one will be passed in runtime. It may be better to hardcode the device in the task code.

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: Yes, view is the better option. I have tried with it, but we do not need these operations because of the reasons mentioned above. I have validated the the example after removing it. It is working fine.

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: The parameters passed in [add_kwargs()](https://openfl.readthedocs.io/en/latest/running_the_federation.html#id4:~:text=%40TI.-,add_kwargs,-(**task_settings) ) overrides the local variables, and the device passed as HPU here gets the precedence and it persists. For the current example when I print the location of the tensors, variables and model parameters in both train and validation task, it shows that they reside on HPU as shown below:

**Train Task:**
image

**Validation Task**
image

Can you help me to know if I am missing something.

MasterSkepticista (approved):

sun1lach:
Looks like Supriya doesn't get the option to add folks to review the PR.
The above reviewer removal was a glitch.
adding reviewers again
@aleksandr-mokrov @psfoley @shreyasnc1 @igor-davidyuk

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
igor-davidyuk: I understand that it works, but you should not rely on this behavior, I mean overriding parameters.
Moreover, you probably want to emphasize that the device used for training is HPU, so I still think it is better to put
`device='HPU:0'` in both tasks' code.

Supriya-Krishnamurthi: Got it, sure will make that change.

igor-davidyuk: Please make a comment there, something like:
```
# This task is designed to run on remote nodes equipped with HPU
# so we hardcode the device name here.
```

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
```diff
@@ -0,0 +1,622 @@
...
+    "        x = x.view(x.size(0), -1)\n",
+    "        x = self.fc(x)\n",
+    "        return x\n",
+    "\n",
+    "# NOTE: This is a workaround to move model to HPU before initializing any optimizers \n",
+    "device = torch.device('hpu')\n",
+    "model = Net(in_channels=n_channels, num_classes=n_classes).to(device)\n",
+    "    \n",
+    "# define loss function and optimizer\n",
+    "if task == \"multi-label, binary-class\":\n",
```
igor-davidyuk: I do not get what you intended to do with this code. You must understand that the notebook is not executed remotely, please place all the logic inside OpenFL interface classes

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
```diff
@@ -0,0 +1,622 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4ff463bd",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "TI = TaskInterface()\n",
+    "\n",
+    "train_custom_params={'criterion':criterion,'task':task}\n",
```
igor-davidyuk: Please do not pass complex Python objects or variables here. There is no logic on Collaborator's side that will pass `'multi-label, binary-class'` as task variable

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: Done

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: This was part of original interactive api [medmnist code](https://github.com/intel/openfl/tree/develop/openfl-tutorials/interactive_api/PyTorch_MedMNIST_2D/workspace). But currently for HPU example I have modified it according to your suggestion

openfl-tutorials/interactive_api/HPU/PyTorch_MedMNIST_2D/workspace/HPU_Pytorch_MedMNIST_2D.ipynb
Supriya-Krishnamurthi: Got it, done.

shreyasnc1 (approved):

MasterSkepticista (approved):

igor-davidyuk (approved):
Good work!

aleksandr-mokrov (approved):
Nice work! 
habanalabs required python 3.8. It is hardcoded in habanalabs-installer.sh. Probably it will be helpful to add the note to README

Supriya-Krishnamurthi:
> Nice work! habanalabs required python 3.8. It is hardcoded in habanalabs-installer.sh. Probably it will be helpful to add the note to README

This change will be tracked in https://github.com/intel/openfl/pull/709

PR #662, opened 2022-12-20
Add tag name argument to graminize

DL8:
Set the graminized image name in "fx workspace graminize" using the "-t" argument. The tag name is formatted with the "worspace_name" input, such that "{workspace_name}" format specifier in the tag name will be replace by the workspace name

If not specified, the workspace name will be used as tag (current behavior)

### DIFF
--- a/openfl/interface/workspace.py
+++ b/openfl/interface/workspace.py
@@ -443,6 +443,11 @@ def dockerize_(context, base_image, save):
              'Must be a power-of-2.\n'
              'Default is 16G.'
         )
+@option('-t', '--tag', required=False,
+        type=str, multiple=False, default='',
+        help='Tag of the built image.\n'
+             'By default, the workspace name is used.'
+        )
 @option('-o', '--pip-install-options', required=False,
         type=str, multiple=True, default=tuple,
         help='Options for remote pip install. '
@@ -453,8 +458,8 @@ def dockerize_(context, base_image, save):
         help='Dump the Docker image to an archive')
 @option('--rebuild', help='Build images with `--no-cache`', is_flag=True)
 @pass_context
-def graminize_(context, signing_key: Path, enclave_size: str, pip_install_options: Tuple[str],
-               save: bool, rebuild: bool) -> None:
+def graminize_(context, signing_key: Path, enclave_size: str, tag: str,
+               pip_install_options: Tuple[str], save: bool, rebuild: bool) -> None:
     """
     Build gramine app inside a docker image.
 
@@ -501,6 +506,10 @@ def graminize_(context, signing_key: Path, enclave_size: str, pip_install_option
 
     workspace_path = Path.cwd()
     workspace_name = workspace_path.name
+
+    if not tag:
+        tag = workspace_name
+
     context.invoke(export_, pip_install_options=pip_install_options)
     workspace_archive = workspace_path / f'{workspace_name}.zip'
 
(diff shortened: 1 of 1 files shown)


openfl/interface/workspace.py
```diff
@@ -504,10 +510,12 @@ def open_pipe(command: str):
 
     grainized_ws_dockerfile = SITEPACKS / 'openfl-gramine' / 'Dockerfile.graminized.workspace'
 
+    tag = tag.format(workspace_name=workspace_name)
```
igor-davidyuk: This is not readable. Would be better to pass `None` as default

DL8: done

openfl/interface/workspace.py
```diff
@@ -504,10 +510,15 @@ def open_pipe(command: str):
 
     grainized_ws_dockerfile = SITEPACKS / 'openfl-gramine' / 'Dockerfile.graminized.workspace'
 
+    if tag is None:
+        tag = workspace_name
+    else:
+        tag = tag.format(workspace_name=workspace_name)
```
igor-davidyuk:
```suggestion

```
`tag` is already a string

DL8: this is meant to allow the user to add the workspace name to the tag. for example, if the name of the workspace is `test_workspace` then `-t 'prefix-{workspace_name}-suffix'` will translate into `prefix-test_workspace-suffix`

itrushkin: I believe that having **both**  `workspace_name` and `tag` command parameters instead of template formatting could improve the understanding and the user experience.

igor-davidyuk: 1. `workspace_name` is not a parameter to the `graminize` command, it is a part of the context as the command must be called from some workspace (it has a name already).
2. regarding `-t 'prefix-{workspace_name}-suffix'` example - this is rather exotic use case, supporting it in our CLI reduces the readability and transparency of the code.  I believe users that want to have `prefix-test_workspace-suffix` image tag should pass that value as a string.

DL8: i needed a way to override the name of the tag/workspace when working on automated image build and execution. i agree that having separate tag and workspace names could be too confusing, so perhaps the following may be more appropriate
1. tag and workspace name will be the same (current behavior in `graminize` and `dockerize`)
2. instead of tag name (`-t`), add an argument to override workspace name: `-w workspace_name`
3. if provided, the workspace name will be overridden and used as tag for the image
4. if not, apply current behavior

igor-davidyuk: As I said above, `workspace_name` is not a parameter to the `graminize` command, it is a part of the context as the command must be called from some workspace (it has a name already)

DL8: note that i recommitted a simpler version in which only the workspace name can be overridden and the tag is always the same

openfl/interface/workspace.py
```diff
@@ -498,7 +503,12 @@ def open_pipe(command: str):
     echo('\n ✔️ DONE: Building base gramine-openfl image')
 
     workspace_path = Path.cwd()
-    workspace_name = workspace_path.name
+    if workspace_name is None:
+        workspace_name = workspace_path.name
+    else:
+        # Prevent potential path traversal issues:
+        # If workspace name is a path, use only the last element
```
igor-davidyuk: This is another surprising assumption 😁

mansishr: I agree with Igor, this might fail based on the `workspace_name` variable.

openfl/interface/workspace.py
```diff
@@ -443,6 +443,11 @@ def dockerize_(context, base_image, save):
              'Must be a power-of-2.\n'
              'Default is 16G.'
         )
+@option('-t', '--tag', required=False,
+        type=str, multiple=False, default=None,
```
igor-davidyuk:
```suggestion
        type=str, multiple=False, default='',
```

openfl/interface/workspace.py
```diff
@@ -501,6 +506,10 @@ def open_pipe(command: str):
 
     workspace_path = Path.cwd()
     workspace_name = workspace_path.name
+
+    if tag is None:
```
igor-davidyuk:
```suggestion
    if not tag:
```

igor-davidyuk (approved):
LGTM!

itrushkin (approved):

mansishr (approved):
LGTM!

PR #664, opened 2022-12-22
Suppress pandas FutureWarning

itrushkin:
## Issue
Training output mostly consists (depending on the number of neural network parameters) of the following warnings:

```
FutureWarning: In a future version, object-dtype columns with all-bool values will not be included in reductions with bool_only=True. Explicitly cast to bool dtype instead.
```
## Solution
Declare all column types on TensorDB initialization.

### DIFF
--- a/openfl/databases/tensor_db.py
+++ b/openfl/databases/tensor_db.py
@@ -38,7 +38,9 @@ class TensorDB:
             'tags': 'object',
             'nparray': 'object'
         }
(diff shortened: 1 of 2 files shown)


igor-davidyuk (approved):
Great work!
We also recently introduced a CLI option to suppress warnings here https://github.com/intel/openfl/pull/644/files

aleksandr-mokrov (approved):

mansishr (approved):

PR #665, opened 2022-12-22
Cleanup temporary files (SDLE task)

mansishr:
Temporary files need to be cleaned to free up the resources.

### DIFF
--- a/openfl/interface/collaborator.py
+++ b/openfl/interface/collaborator.py
@@ -160,6 +160,7 @@ def generate_cert_request(collaborator_name, data_path, silent, skip_package):
         from shutil import copytree
         from shutil import ignore_patterns
         from shutil import make_archive
+        from shutil import rmtree
         from tempfile import mkdtemp
         from os.path import basename
         from os.path import join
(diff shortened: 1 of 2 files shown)


igor-davidyuk (approved):
The cleaning is good.
But it is not obvious to me that `make_archive` creates the file in the current directory and not in the root directory that is passed to it. Could you leave a comment to make this clear?
I have another concern regarding `tempfile.mkdtemp`, which probably will not work in Gramine. An alternative would be creating a temp directory inside the workspace.

mansishr:
`make_archive` takes `basename` (name of the path to create) as the first argument. For OpenFL, we are not specifying a path, just a name so the archive is created under `cwd`. `root_dir` is used to specify the root, and all paths are taken relative to this root directory. Check an example [here](https://docs.python.org/3/library/shutil.html#shutil-archiving-example-with-basedir).

psfoley:
@mansishr Before merging, could you rebase with the signoff using these instructions? https://github.com/intel/openfl/pull/665/checks?check_run_id=10259301135

psfoley (approved):

PR #667, opened 2022-12-27
Parameterize PKI certs location

mansishr:
For manual PKI exchange, certificates are stored under 'cert' directory inside workspace by default. This PR decouples the location of certificates from workspace in that the user can now specify certificate path to store certs which can reside outside the workspace.

- For CA, `cert_dir` can be passed, where all CA certs and keys can reside. To store `signing-ca.crt`, certificate chain (`cert_chain.crt`) and `signing-ca.key`, use `cert_path`and `key_path `  to specify locations.
- For both aggregator and collaborator, use `cert_path` and `key_path` to store aggregator (or collaborator) certificate and key, respectively.
- Introducing `fx workspace participants` that let's the CA/aggregator select a subset of certified collaborators for a federation by modifying `plan/cols.yaml`.


tests/github/test_pki_cert_location.sh
```diff
@@ -0,0 +1,146 @@
...
+
+cd ${FED_DIRECTORY}
+fx aggregator uninstall-cert -c ${CERT_PATH_AG}
+cd ${COL1_DIRECTORY}/${FED_WORKSPACE}
+fx collaborator uninstall-cert -c ${CERT_PATH_COL1}
+cd ${COL2_DIRECTORY}/${FED_WORKSPACE}
+fx collaborator uninstall-cert -c ${CERT_PATH_COL2}
+
+rm -rf ${FED_DIRECTORY}
+
```
itrushkin: This test should be rewritten to Python code after merging of #587.

openfl/interface/workspace.py
```diff
@@ -232,6 +234,11 @@ def certify():
     echo('1.  Create Root CA')
     echo('1.1 Create Directories')
 
+    if cert_path:
+        CERT_PATH = Path(cert_path).absolute()
```
itrushkin: This is a local variable, not a constant. The name should use lowercase letters to avoid misunderstanding.

openfl/interface/workspace.py
```diff
@@ -232,6 +234,11 @@ def certify():
...
+        (CERT_PATH / 'cert').mkdir(parents=True, exist_ok=True)
+        CERT_DIR = CERT_PATH/ 'cert' # NOQA
```
itrushkin: Could we avoid reassigning the constant value imported from another module?

openfl/interface/aggregator.py
```diff
@@ -36,7 +36,13 @@ def aggregator(context):
         default='plan/cols.yaml', type=ClickPath(exists=True))
 @option('-s', '--secure', required=False,
         help='Enable Intel SGX Enclave', is_flag=True, default=False)
-def start_(plan, authorized_cols, secure):
+@option('-c', '--cert_path',
+        help='The cert path where pki certs will reside', required=False)
+@option('--fqdn', required=False, type=click_types.FQDN,
+        help=f'The fully qualified domain name of'
+             f' aggregator node [{getfqdn_env()}]',
+        default=getfqdn_env())
```
itrushkin: Aggregator FQDN could be read from the plan. According to the CLI flow, the plan should be initialized by the step of aggregator start.

itrushkin (changes requested):

openfl/interface/workspace.py
mansishr: `CERT_DIR` path is either fixed (`Path('cert')`) or provided by the user at runtime now. Avoiding reassigning this variable might be difficult.

igor-davidyuk: Try introducing a local variable

openfl/interface/aggregator.py
igor-davidyuk: I agree with Iliya, this option overrides information in `plan.yaml`.
Why would we need this?

openfl/interface/aggregator.py
```diff
@@ -36,7 +36,13 @@ def aggregator(context):
         default='plan/cols.yaml', type=ClickPath(exists=True))
 @option('-s', '--secure', required=False,
         help='Enable Intel SGX Enclave', is_flag=True, default=False)
-def start_(plan, authorized_cols, secure):
+@option('-c', '--cert_path',
```
igor-davidyuk, line 39: What if we allow user to provide paths to certificates and keys? This way we do not implicitly rely on the user certificate folder to replicate our internal structure. For example, we will not expect aggregator private key be in format `{CERT_DIR}/server/agg_{common_name}.key`

mansishr: I tried to keep the certs structure consistent but we can do that too. It would also involve users to provide a path where certs will be stored `{CERT_DIR}/agg_{common_name}.key`.  @psfoley @itrushkin @aleksandr-mokrov what do you think?

itrushkin: We should not force users to keep our internal PKI folder structure. I think it is better to split this argument and use paths for each file.

mansishr: Added separate paths to certs and keys to not rely on internal PKI folder structure.

theakshaypant:
@mansishr can you please resolve the conflicts and request for review again?

PR #669, opened 2023-01-03
[CI] Process GitHub API rate limit error

itrushkin:
## Issue
CI job `test-insecure-client` fails because of exceeding the GitHub API rate limit. We fetch `step-ca` binaries from GitHub by calling `api.github.com` which has a rate limit for unauthorized clients. When we create multiple pull requests in a short period of time, some tests may fail due to GitHub request denials.

## Solution
Change step-ca URLs from `api.github.com` to `dl.step.sm`.

### DIFF
--- a/openfl/component/ca/downloader.py
+++ b/openfl/component/ca/downloader.py
@@ -38,9 +38,6 @@ def download_step_bin(prefix='.', confirmation=True):
     Donwload step binaries from github.
 
     Args:
-        url: address of latest release
-        grep_name: name to grep over github assets
-        architecture: architecture type to grep
         prefix: folder path to download
         confirmation: request user confirmation or not
     """
@@ -52,6 +49,13 @@ def download_step_bin(prefix='.', confirmation=True):
 
 
 def download_step_ca_bin(prefix='.', confirmation=True):
+    """
+    Donwload step binaries from github.
+
+    Args:
+        prefix: folder path to download
+        confirmation: request user confirmation or not
+    """
     system, arch = get_system_and_architecture()
     ext = FILE_EXTENSIONS[system]
     binary = f'step-ca_{system}_{VERSION}_{arch}.{ext}'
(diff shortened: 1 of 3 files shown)


tests/github/pki_insecure_client.py
```diff
@@ -45,9 +45,21 @@ def install_ca():
         'fx', 'pki', 'install',
         '-p', str(CA_PATH),
         '--password', str(CA_PASSWORD)
-    ], stdin=subprocess.PIPE)
+    ], stdin=subprocess.PIPE, stdout=subprocess.PIPE)
     time.sleep(3)
-    pki_install.communicate('y'.encode('utf-8'))
+    stdout, _ = pki_install.communicate('y'.encode('utf-8'))
+    while 'API rate limit exceeded' in stdout.decode('utf-8'):
+        minutes_to_wait = 5
```
igor-davidyuk: Would it be better to move this constant to the top of the script?
Also, is there any meaning in 5 minutes delay, can we make it smaller?

psfoley:
Marking `WIP` while @itrushkin looks into alternate download location for `step-ca`

itrushkin:
Removed WIP

openfl/component/ca/downloader.py
```diff
@@ -0,0 +1,71 @@
+# Copyright (C) 2020-2021 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import platform
+import urllib.request
+import shutil
+
+from click import confirm
+
+VERSION = '0.16.0'
```
mansishr, line 10: Is there a specific reason to use this version when latest versions are available?

itrushkin: The latest `step-ca` release has a different folder structure from what we expect to have in our PKI interface. The transition could not be done smoothly and requires changes in our code that should be addressed in different PR.

mansishr (approved):

igor-davidyuk (approved):
Looks good

PR #670, opened 2023-01-03
ci: add branch value for pypiPublish pushOn parameter

soda480:
Add `pypiPublishPushOn` parameter to designate valid branch for pushing to PyPi. This is a parameter that was recently added to the `pypiPublish` function and is needed to designate which branches are allowed to publish to PyPi.

Signed-off-by: Emilio Reyes <emilio.reyes@intel.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -129,7 +129,9 @@ pipeline {
                 }
                 stage('Publish Package') {
                     steps {
-                        pypiPublish()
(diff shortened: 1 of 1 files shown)


soda480:
I am closing this PR for now; we will be updating the `pypiPublish` function in our jenkins-common-pipelines shared library to not require the `pypiPublishPushOn` parameter.

PR #673, opened 2023-01-05
CI badges

itrushkin:
## New
- Ubuntu test sequence
- Windows test sequence
- 2 badges in the README.md

The test sequence for both operating systems is following:
- Linter (removed on Windows because of https://github.com/savoirfairelinux/flake8-copyright/pull/16)
- PyTest
- Dockerization (removed on Windows because of inability to switch to Windows containers inside Github Actions).
- Interactive API (UNet Kvasir example)
- CLI (Keras CNN MNIST template)

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,8 +1,9 @@
 name: Ubuntu (latest)
 
 on:
-  schedule:
-    - cron: '0 0 * * *'
+  push:
+    branches:
+      - ci_badges
 
 permissions:
   contents: read
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,4 +1,4 @@
-name: ['Ubuntu (latest)']
+name: Ubuntu (latest)
 
 on:
   schedule:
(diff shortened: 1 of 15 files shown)


itrushkin:
Badges could not be rendered. According to [Github documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule), scheduled workflows will trigger only after workflow files appear in the default (develop) branch. In order for badges to render, this PR should be merged. Both workflows have been tested locally.

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,101 @@
+name: Ubuntu (latest)
+
+on:
+  push:
+    branches:
+      - ci_badges
```
psfoley: `ci_badges` is the name of this branch, which I would expect would be deleted after merge. Should the condition to run be pushing to `develop` instead?

itrushkin: Yes, this is for debugging purpose. I am checking the pipeline on GitHub Actions.

itrushkin:
Badges in this branch have been rendered. Resetting the trigger to scheduled.

psfoley (approved):
LGTM!

Issue #692, opened 2023-01-16
Informative prompt in certification process


DL8:
**Is your feature request related to a problem? Please describe.**

Certification process using `fx` command line works as follows (for both aggregator and collaborator):
1. CSR is generated using `generate-cert-request`
2. CSR is signed using `certify`

Currently the outputs of these commands are not informative enough to provide a reliable way to verify the CSR: `certify` prints the hash of the CSR file, but `generate-cert-request` doesn't, so there is no real way to validate the CSR

**Describe the solution you'd like**

1. Prior to writing the CSR to disk, `generate-cert-request` calculates and prints its hash
2. `certify` prints the hash
3. User is expected to compare the two hashes and approve the request only if they match. It may be useful to emphasize this expectation, for example in openfl documentation and/or as part of the prompt
4. Adding a warning to the `certify -s` flag can be considered
5. Printing additional data from the CSR (e.g. subject/SAN) can be considered

**Describe alternatives you've considered**

It is possible to manually calculate the CSR file's hash after it was written to disk, but a malicious actor may be able replace it with another one before the actual check happens. Therefore, this solution is not secure

bjklemme-intel:
OpenFL_Team1 is working on this

PR #693, opened 2023-01-16
Print CSR hash in generate-cert-request (#692)

DL8:
Currently, there is no real way to verify CSR properly when using fx command line: since generate-cert-request does not print the CSR hash, it is not really possible to know in certify that the CSR being signed is the correct one. To mitigate this, add a print with the CSR hash in generate-cert-request. User is then expected to compare the hashes and approve the CSR only if they match


theakshaypant:
Closing this PR as #813 added this functionality to openfl.

Issue #672, opened 2023-01-04
Installing OpenFL from develop doesn't work


DL8:
**Describe the bug**
`fx` command fails with the following after installing OpenFL from repo:
```
$ fx
Traceback (most recent call last):
  File "/home/vmuser/venv_test/bin/fx", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/home/vmuser/venv_test/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3252, in <module>
    def _initialize_master_working_set():
  File "/home/vmuser/venv_test/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3235, in _call_aside
    f(*args, **kwargs)
  File "/home/vmuser/venv_test/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3264, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/home/vmuser/venv_test/lib/python3.8/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/home/vmuser/venv_test/lib/python3.8/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/vmuser/venv_test/lib/python3.8/site-packages/pkg_resources/__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'jsonpointer>1.13; extra == "format-nongpl"' distribution was not found and is required by jsonschema
```

A second installation on the same environment fixes this issue, but it's not very elegant to demand invoking `pip install` twice

**To Reproduce**
```bash
python -m venv venv_test
. venv_test/bin/activate
git clone https://github.com/intel/openfl.git
pip install ./openfl
fx
```

**Expected behavior**
`fx` runs without dependency errors

MasterSkepticista:
Could you mention the Python version?

MasterSkepticista:
Could you mention the `pip` version?
Besides, it is generally recommended to upgrade `pip`, `setuptools` and `wheel` before installing packages in a new virtual environment.
You may give this a try (in a fresh venv):
```bash
python -m venv venv_test
. venv_test/bin/activate
# Upgrade base package helpers
python -m pip install -U pip setuptools wheel
git clone https://github.com/intel/openfl.git
pip install ./openfl
fx
```

DL8:
Python version:
```
$ python --version
[...]

PR #695, opened 2023-01-17
Align graminize process with setup documentation

DL8:
Per fix for #672, seup documentation was updated to upgrade pip, setuptools and wheel. Align base Dockerfile used in graminize to reflect this change


psfoley (approved):
LGTM!

PR #697, opened 2023-01-17
Remove FastEstimator until compatible dependencies are available

psfoley:
Due to several incompatibilities introduced by the latest torch version (https://github.com/intel/openfl/pull/661 and https://github.com/intel/openfl/pull/657), and downstream pinned package versions required by FastEstimator, prior examples no longer work with OpenFL.  This PR removes those examples, tutorials, and the FastEstimator task runner until a compatible combination of packages can be found in the future.


aleksandr-mokrov (approved):

mansishr:
Need to remove fastestimator reference from __init__.py : https://github.com/intel/openfl/blob/develop/openfl/federated/__init__.py#L18

psfoley:
> Need to remove fastestimator reference from **init**.py : https://github.com/intel/openfl/blob/develop/openfl/federated/__init__.py#L18

https://github.com/intel/openfl/pull/702

PR #698, opened 2023-01-17
Pytorch version is updated to 1.13.1

aleksandr-mokrov:
Pytorch version is updated to 1.13.1. Experiments were tested with python3.10

### DIFF
--- a/.github/workflows/fets-challenge.yml
+++ b/.github/workflows/fets-challenge.yml
@@ -36,7 +36,7 @@ jobs:
       run: |
         cd fets_challenge
         pwd
-        pip install torch==1.13.1+cpu torchvision==0.14.1+cpu torchaudio===0.13.1 -f https://download.pytorch.org/whl/cpu/torch_stable.html
+        pip install torch==1.8.2+cpu torchvision==0.9.2+cpu torchaudio===0.8.2 -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html
         pip install onnx==1.12.0
         pip install -e .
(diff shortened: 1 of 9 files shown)


psfoley (approved):

itrushkin:
Should we also bump versions being installed in CI workflows? For example, in `.github/workflows/interactive-kvasir.yml`.

.github/workflows/fets-challenge.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       run: |
         cd fets_challenge
         pwd
-        pip install torch==1.8.2+cpu torchvision==0.9.2+cpu torchaudio===0.8.2 -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html
+        pip install torch==1.13.1+cpu torchvision==0.14.1+cpu torchaudio===0.13.1 -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html
```
psfoley: The GaNDLF Package has a hard dependency on 1.8.2. This should be reverted.

psfoley: https://github.com/mlcommons/GaNDLF/blob/0.0.14/setup.py#L88

aleksandr-mokrov:
> Should we also bump versions being installed in CI workflows? For example, in .github/workflows/interactive-kvasir.yml.
> The GaNDLF Package has a hard dependency on 1.8.2. This should be reverted.
@itrushkin, @psfoley all versions are bumped, excluding fets-challenge.yml that uses GaNDLF Package.

PR #728, opened 2023-02-08
Allow grpcio <=1.48.2 for Raspberry Pi installation

### LINKED ISSUE #727: How strict is the grpcio~=1.48.2 requirement?
I was trying to run openfl on a raspberry pi 3 B+ running debian 11 bullseye, and grpcio~=1.48.2 did not work because of a glibc version error. Downgrading to grpcio==1.44 seems to fix the problem, and openfl works fine from what I can tell (although I haven't tested this extensively). So, I was wondering if any features not in grpcio==1.44, but in grpcio~=1.48.2 are used? If not, would it make sense to depend on grpcio==1.44 instead?

psfoley:
grpcio 1.48.2 throw a glibc error on Raspberry Pi installation. This change will still install 1.48.2 by default, but allow for user downgrades afterwards

Closes https://github.com/intel/openfl/issues/727

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -143,7 +143,7 @@ setup(
         'docker',
         'dynaconf==3.1.7',
         'flatten_json',
-        'grpcio~=1.48.2',
+        'grpcio<=1.48.2',
         'ipykernel',
         'jupyterlab',
         'numpy',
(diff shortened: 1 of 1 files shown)


MasterSkepticista:
Closing this PR due to inactivity.

PR #736, opened 2023-02-16
Resolve AST parsing error in Jupyter Notebook

### LINKED ISSUE #724: Workflow Interface: AST parsing error can arise in jupyter notebooks
When using the new experimental workflow interface in a jupyter notebook, an error produced at any point during execution may lead to an AST parsing error when calling `flow.run()`. The error is thrown by metaflow's DAG generation code:

image

Because the error is generated at the metaflow level, I think the DAG UI genertion code should likely be disabled by default, and we should provide an argument to enable it instead.

KeertiX:
**ISSUE** :
- Described in #724 

**STEPS TO REPRODUCE** : 
- Described in #724 
	
**FIX DESCRIPTION / SUMMARY OF CHANGES**: 

1. Refactor Flowgraph related functionality (graph and steps creation) to InspectFlow()
    Rationale: There is no need to create the graph and steps in MetaflowInterface and these steps can be performed only 
    if the user desires to generate FlowGraph

2. Fetch the source code of flow instead of module in FlowGraph()
    Rationale: Only flow source code is required for generation of Graph. 
    Understanding is that this change should also obviate the need to reload the module as it should always fetch the 
    updated source code for the flow
	
**VERIFICATION SUMMARY**:

Verified if the fix is working in Google colab, Jupyter notebook and python file
- Add bug in any of the jupyter notebook cell/google colab/py file and run. 
- Fix the issue and rerun the cell till flflow.run to verify whether its throwing AST error
- For supported flows, verified that graph html is generated properly

Closes #724

### DIFF
--- a/openfl/experimental/utilities/metaflow_utils.py
+++ b/openfl/experimental/utilities/metaflow_utils.py
@@ -201,7 +201,9 @@ class FlowGraph(FlowGraph):
         self._postprocess()
 
     def _create_nodes(self, flow):
-        tree = ast.parse(getsource(flow)).body
+        module = __import__(flow.__module__)
+        tree = ast.parse(getsource(module)).body
+        # tree = ast.parse(getsource(flow)).body
         root = [
             n
             for n in tree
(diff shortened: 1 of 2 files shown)


tanwarsh:
This PR is not required anymore as this issue is already fixed as a part of PR [910](https://github.com/securefederatedai/openfl/pull/910)
Confirmed by trying to reproduce this issue.

Added syntax error.
Screenshot 2024-10-17 at 10 08 42 AM

Rerun after fixing the error:
Screenshot 2024-10-17 at 10 14 06 AM

PR #758, opened 2023-03-13
Testflow for verifying stdout redirection to Metaflow datastore

KeertiX:
**BACKGROUND**:
- Output of the collaborator steps (executed in Ray context) can be  observed in aggregator steps (executed out of Ray 
   context)
- Observed while executing a test case to validate stdout capturing functionality. The Test case is a very simple Test case and 
   not representative of actual Federated Learning workloads 
- When the Test case was modified to include a NN model and some inferencing / ML processing the same issue could not 
   be observed / reproduced
- As ray backend is not recommended for processing of tiny tasks therefore the proposal is to commit this Test case so that any issues in stdout redirection functionality are exposed with a TC that is representative of a FL workload

**TEST CASE DESCRIPTION**: 
1. Test case verifies that the stdout of the Aggregator and Collaborator steps is captured properly in METAFLOW datastore
2. Test case uses a simple MNIST model to verify this functionality
3. Test case is verified with 4 collabs for 10 rounds, 20 rounds, 50 rounds and 100 rounds

**VERIFICATION SUMMARY**:
   Verified the test case in Google colab, Jupyter notebook and python file.

### DIFF
--- a/openfl/experimental/placement/placement.py
+++ b/openfl/experimental/placement/placement.py
@@ -26,15 +26,7 @@ class RayExecutor:
         del ref_ctx
 
     def get_remote_clones(self):
-        # clones = deepcopy(ray.get(self.remote_functions))
-        clones = []
-        while len(self.remote_functions):
-            ready_ids, self.remote_functions = ray.wait(self.remote_functions)
-            if len(ready_ids)>1:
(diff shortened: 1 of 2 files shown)


psfoley (commented):
Approved

PR #807, opened 2023-04-25
[Bug: 768] FX CLI: Separate create, cert gen commands

### LINKED ISSUE #768: FX CLI: Separate user creation from TLS certificate request generation
## Background
Today in the `fx` cli, there is a combined step that create a user for an experiment and also generates a certificate signing request for them. While the use of mutual TLS is the default in OpenFL, we should not impose it on users for all environments (for example, there can be issues defining a FQDN on internal/personal infrastructure).  

## Task
1. Split `fx collaborator generate-cert-request` into two commands:
    - `fx collaborator create -n {NAME} -d {DATA_PATH: optional}` : 
    - `fx collaborator generate-cert-request -n {NAME}`
 
## TODO items for maintainers
1. Clarify how users should be added to the `plan/cols.yaml` file when not using TLS

bjklemme-intel:
This change separates existing command
"fx collaborator.py generate-cert-request" command into two commands.
"fx collaborator create -n {NAME} -d {DATA_PATH: optional}". "fx collaborator generate-cert-request -n {NAME}".

Fixes #768

### DIFF
--- a/docs/running_the_federation.rst
+++ b/docs/running_the_federation.rst
@@ -1047,6 +1047,7 @@ Importing the Workspace
 
     .. code-block:: console
 
+       fx collaborator create -n {COL_LABEL} -d {DATA_PATH:optional}
        fx collaborator generate-cert-request -n {COL_LABEL}
 
 
--- a/openfl/interface/collaborator.py
+++ b/openfl/interface/collaborator.py
@@ -61,6 +61,24 @@ def start_(plan, collaborator_name, data_config, secure):
     plan.get_collaborator(collaborator_name).run()
 
 
+@collaborator.command(name='create')
+@option('-n', '--collaborator_name', required=True,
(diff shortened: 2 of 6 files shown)


mansishr (changes requested):
Thanks for submitting, the PR looks good! There will need to be some more changes:
1. Changes to `init` function in [native](https://github.com/securefederatedai/openfl/blob/develop/openfl/native/native.py#L203)
2. Changes to `create_collaborator` [function name](https://github.com/securefederatedai/openfl/blob/develop/openfl/native/native.py#L211)
3. Changes to the [documentation](https://github.com/securefederatedai/openfl/blob/develop/docs/running_the_federation.rst#step-2-configure-the-federation)

ejillela:
@mansishr I have addressed the changes you requested. please continue your review

psfoley (approved):

PR #810, opened 2023-04-27
Improve workspace requirements import

### LINKED ISSUE #767: Improve workspace requirements import
## Background
Users have reported long installation times for pip dependencies installed when importing a workspace (applies to multiple OpenFL workspaces). 

## Task
1. Understand underlying issues with importing pip dependencies across different virtual environments
2. Modify import logic in `openfl/interface/workspace.py`

danhe1:
This adds two commands **fx workspace export** and **fix workspace import** to implement dependencies import across different virtual environments. 

An example of using those two commands.

1. Create a workspace named workspace.
`fx workspace create --prefix ./workspace --template keras_cnn_mnist`
2. Export the requirements in workspace, which will generate an archive file workspace.zip in the directory workspace/.
`cd workspace && fx workspace export`
3. Change the virtual environment.
4. Create workspace2 with the archived file.
`mkdir workspace2 && cd workspace2 && fx workspace import --archive ../workspace/workspace.zip`

Fixes #767

### DIFF
--- a/openfl/interface/workspace.py
+++ b/openfl/interface/workspace.py
@@ -17,7 +17,6 @@ from click import pass_context
 from click import Path as ClickPath
 
 from openfl.utilities.path_check import is_directory_traversal
-from openfl.utilities.workspace import dump_requirements_file
 
 
 @group()
@@ -121,6 +120,7 @@ def export_(pip_install_options: Tuple[str]):
     """Export federated learning workspace."""
     from os import getcwd
     from os import makedirs
+    from os.path import isfile
     from os.path import basename
     from os.path import join
     from shutil import copy2
@@ -139,9 +139,6 @@ def export_(pip_install_options: Tuple[str]):
     except Exception:
         echo(f'Plan file "{plan_file}" not found. No freeze performed.')
 
-    # Dump requirements.txt
-    dump_requirements_file(prefixes=pip_install_options, keep_original_prefixes=True)
-
     archive_type = 'zip'
     archive_name = basename(getcwd())
     archive_file_name = archive_name + '.' + archive_type
@@ -158,7 +155,10 @@ def export_(pip_install_options: Tuple[str]):
     makedirs(f'{tmp_dir}/data', exist_ok=True)
     copytree('./src', f'{tmp_dir}/src', ignore=ignore)  # code
     copytree('./plan', f'{tmp_dir}/plan', ignore=ignore)  # plan
-    copy2('./requirements.txt', f'{tmp_dir}/requirements.txt')  # requirements
+    if isfile('./requirements.txt'):
+        copy2('./requirements.txt', f'{tmp_dir}/requirements.txt')  # requirements
+    else:
+        echo('No requirements.txt file found.')
 
     try:
         copy2('.workspace', tmp_dir)  # .workspace


openfl/interface/workspace.py
```diff
@@ -139,8 +150,8 @@ def export_(pip_install_options: Tuple[str]):
     except Exception:
         echo(f'Plan file "{plan_file}" not found. No freeze performed.')
 
-    # Dump requirements.txt
-    dump_requirements_file(prefixes=pip_install_options, keep_original_prefixes=True)
+    # No need to dump all dependencies, just use template requirements.txt to export to wheel packages
```
psfoley: These comments should be removed

openfl/interface/workspace.py
```diff
@@ -203,9 +224,7 @@ def import_(archive):
         check_call([
             executable, '-m', 'pip', 'install', '--upgrade', 'pip'],
             shell=False)
-        check_call([
-            executable, '-m', 'pip', 'install', '-r', 'requirements.txt'],
-            shell=False)
+        _import_whls_to_venv()
```
psfoley: This could break if different collaborators are using a different python version than the aggregator. Because this assumes that all dependencies are captured in `requirements.txt` anyway, I would recommend falling back to the original behavior (i.e. `pip install -r requirements.txt`) in the case that the import fails

psfoley:
@danhe1 Saving wheels in the workspace increases the zip file size significantly - as an example the `torch_cnn_mnist` workspace increases from 66 MB to 1.8 GB. You've resolved the crux of this issue - which is that the packages in the virtual environment that are *not* explicitly part of the original `requirements.txt` should not be saved because this leads to many potential version conflicts.

To merge this, I would recommend instead just removing the call to `dump_requirements_file`, and skip the dumping / import of wheels

psfoley (changes requested):
@danhe1 Saving wheels in the workspace increases the zip file size significantly - as an example the torch_cnn_mnist workspace increases from 66 MB to 1.8 GB. You've resolved the crux of this issue - which is that the packages in the virtual environment that are not explicitly part of the original requirements.txt should not be saved because this leads to many potential version conflicts.

To merge this, I would recommend instead just removing the call to dump_requirements_file, and skip the dumping / import of wheels

The lint check should be fixed prior to merge: https://github.com/securefederatedai/openfl/actions/runs/4826556296/jobs/8610922793?pr=810

danhe1:
> @danhe1 Saving wheels in the workspace increases the zip file size significantly - as an example the torch_cnn_mnist workspace increases from 66 MB to 1.8 GB. You've resolved the crux of this issue - which is that the packages in the virtual environment that are not explicitly part of the original requirements.txt should not be saved because this leads to many potential version conflicts.
> 
> To merge this, I would recommend instead just removing the call to dump_requirements_file, and skip the dumping / import of wheels
> 
> The lint check should be fixed prior to merge: https://github.com/securefederatedai/openfl/actions/runs/4826556296/jobs/8610922793?pr=810

We have provided two solutions. The solution at this PR is to remove the call to dump_requirements_file. And we also have a solution of dumping the wheels only appears in the requrements.txt. https://github.com/danhe1/openfl/tree/issue-767-dev-dump-req. The ideas are the same, just to ignore the dependency packages.

psfoley (approved):

PR #813, opened 2023-04-27
This prints out the hash of the CSR to disk for both the aggregator and

### LINKED ISSUE #692: Informative prompt in certification process

bjklemme-intel:
collaborator. The user then compares and approves this hash with the hash printed out of the file to validate the CSR. In addition, a warning message is pritned if certify is run in silent mode.

Fixes securefederatedai#692

### DIFF
--- a/openfl/cryptography/io.py
+++ b/openfl/cryptography/io.py
@@ -111,3 +111,21 @@ def read_csr(path: Path) -> Tuple[CertificateSigningRequest, str]:
     # TODO: replace assert with exception / sys.exit
     assert (isinstance(csr, x509.CertificateSigningRequest))
     return csr, hasher.hexdigest()
+
+
+def get_csr_hash(certificate: CertificateSigningRequest) -> str:
+    """
+    Get hash of cryptography certificate.
+
+    Args:
+        certificate : Cryptography CSR object
+
+    Returns:
+        Hash of cryptography certificate / csr
+    """
+    hasher = sha384()
+    encoded_bytes = certificate.public_bytes(
+        encoding=serialization.Encoding.PEM,
+    )
+    hasher.update(encoded_bytes)
+    return hasher.hexdigest()
--- a/openfl/cryptography/io.py
+++ b/openfl/cryptography/io.py
@@ -102,15 +102,13 @@ def read_csr(path: Path) -> Tuple[CertificateSigningRequest, str]:
     Returns:
         Cryptography CSR object
     """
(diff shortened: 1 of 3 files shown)


openfl/cryptography/io.py
```diff
@@ -111,3 +111,21 @@ def read_csr(path: Path) -> Tuple[CertificateSigningRequest, str]:
     # TODO: replace assert with exception / sys.exit
     assert (isinstance(csr, x509.CertificateSigningRequest))
     return csr, hasher.hexdigest()
+
+
+def get_csr_hash(certificate: CertificateSigningRequest) -> str:
```
psfoley, line 114: I like the separation of the CSR hashing functionality. I would just recommend refactoring the `read_csr` function to reuse the `get_csr_hash` function internally

openfl/interface/aggregator.py
```diff
@@ -183,6 +189,7 @@ def certify(fqdn, silent):
 
         if confirm('Do you want to sign this certificate?'):
 
+            echo('Make sure the two hashes above are the same.')
```
psfoley: Per the requirement:

"User is expected to compare the two hashes and approve the request only if they match. It may be useful to emphasize this expectation, for example in openfl documentation and/or as part of the prompt"

At this point the decision has already been made to sign the certificate. A message prompting for verification should appear before line 190

psfoley (changes requested):
Once the comments have been resolved, this PR can be merged!

psfoley (approved):

PR #814, opened 2023-04-28
Support for federated evaluation

### LINKED ISSUE #788: Support for federated evaluation
Add an example of federated evaluation using [aggregator based workflow](https://openfl.readthedocs.io/en/latest/running_the_federation.html#aggregator-based-workflow) or the [director based workflow](https://openfl.readthedocs.io/en/latest/running_the_federation.html#director-based-workflow). Federated evaluation refers to model evaluation on the client/collaborator side. Once the global model is sent to the collaborators, they could perform one pass of local evaluation. Evaluation wouldn't need any training/construction of optimizers/gradient descent.

For aggregator based workflow:
- An existing workspace like `torch_cnn_mnist` could be taken as a base example from the list of [supported templates for OpenFL](https://github.com/securefederatedai/openfl/tree/develop/openfl-workspace) to create a new template for federated evaluation. 

- Changes to the `plan/plan.yaml` will be needed:
  - Since there's no training involved, parameter `rounds_to_train` should be set to 1
  - Task `aggregator_model_validation` should be assigned instead of the [default tasks](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/workspace/plan/defaults/tasks_torch.yaml).

- Changes to the `src/pt_cnn.py` will be needed:
  - No optimizer initialization needed.

wenjiewang-intel:
This adds a new tutorial example Support for federated evaluation with aggregator based workflow.

The example adds federate evaluation after the model has been trained. In the plan, we set parameter `rounds_to_train` to 1 and make collaborators only do `aggregator_model_validation` task.

Fixes  #788

### DIFF
--- a/openfl-tutorials/Federated_Pytorch_MNIST_Tutorial_Evaluation.ipynb
+++ b/openfl-tutorials/Federated_Pytorch_MNIST_Tutorial_Evaluation.ipynb
@@ -1,0 +1,1135 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Federated PyTorch MNIST Tutorial"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Requirement already satisfied: torch in /usr/local/lib/python3.10/dist-packages (1.13.1)\n",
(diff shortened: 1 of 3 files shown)


psfoley:
Thanks for the submission. The issue (https://github.com/securefederatedai/openfl/issues/788) requested creating a template for the aggregator based workflow, but the PR created an example for (and several changes to) the Python native API which will soon be deprecated. There is also not corresponding documentation that we would help users understand the steps necessary to perform federated evaluation. For the intent of the hackathon, this particular PR will be closed, but please feel free to create a new PR that adds documentation and removes the changes to the python native API

PR #816, opened 2023-04-28
Create interfaces for OpenFL components with methods that can be impl…

### LINKED ISSUE #802: Create interfaces for OpenFL components with methods that can be implemented
All OpenFL components like aggregator, collaborator, aggregation algorithms are directly implemented which makes it difficult to write new s/w that would want to customize a few of these components. For ex; Add/test new algos, a new add-on based on OpenFL. 
Either the base OpenFL code has to be tweaked or inherit these classes and re-implement the code to suit the requirement

If OpenFL components had interfaces defined, these interfaces can be implemented by someone looking to re-use OpenFL.
An option to directly use OpenFL's components should also be provided

**Describe alternatives you've considered**
TBD

danhe1:
Create interfaces for OpenFL components with methods that can be implementedCreate interfaces for OpenFL components with methods that can be implemented.

Create AggregatorInterface and CollaboratorInterface in openflinterface/interactive_api/experiment.py Support passing the aggregator object and collaborator object in the method FLExperiment.start() If aggregator object or collaborator object is None, the default aggregator and collaborator objects will be created at director and envoy services.

Known issue:
Since the AggregatorInterface and CollaboratorInterface are for ad-hoc algorithm development. We should modify the definitions of the model, data loader and task interfaces, since we are going to support some algorithms other than Keras, TensorFlow or PyTorch.

Fixes securefederatedai#802

### DIFF
--- a/openfl/federated/plan/plan.py
+++ b/openfl/federated/plan/plan.py
@@ -335,7 +335,19 @@ class Plan:
 
         defaults[SETTINGS]['log_metric_callback'] = log_metric_callback
         if self.aggregator_ is None:
-            self.aggregator_ = Plan.build(**defaults, initial_tensor_dict=tensor_dict)
+            # Try loading the aggregator from the serialized object file
+            api_layer = self.config.get('api_layer', {})
+            api_settings = api_layer.get('settings', {})
+            obj_file = api_settings.get('aggregator_interface_file')
+            if obj_file is not None:
+                self.aggregator_ = self.restore_object(obj_file)
+
+            if self.aggregator_ is None:  # Failed to load the object file
+                self.aggregator_ = Plan.build(**defaults, initial_tensor_dict=tensor_dict)
+            else:
+                # Pass the settings to the deserialized object
+                self.aggregator_.initialize(**defaults[SETTINGS],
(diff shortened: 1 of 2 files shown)


psfoley:
@danhe1 and team, thanks for the submission. The maintainer team has determined that there will be a larger refactoring of the code necessary to support the interfaces proposed in https://github.com/securefederatedai/openfl/issues/802, so we are going to close this PR for now. If you're interested in staying involved here, I'll loop you into future conversations in this space. I expect this will come in around the OpenFL ~1.7 timeframe.

danhe1:
> @danhe1 and team, thanks for the submission. The maintainer team has determined that there will be a larger refactoring of the code necessary to support the interfaces proposed in #802, so we are going to close this PR for now. If you're interested in staying involved here, I'll loop you into future conversations in this space. I expect this will come in around the OpenFL ~1.7 timeframe.

Hi, Patrick. Thanks for your comment. I agree that we need to modify a lot of related code for this issue. We are glad if we can continuing contributing to this issue. By saying OpenFL 1.7, I guess this means the 2nd half of this year. We are wondering how should we get involved in the development? What part are we responsible to? Maybe we can talk more about it when we get started.

PR #825, opened 2023-05-16
Fix warnings and issues in docs

akantak:


docs/log_metric_callback.rst
```diff
@@ -39,8 +39,8 @@ You could also add your custom metric logging function by defining the function
         your code 
 
 
-Example of a Metric Callback
-============================
+Example of a MLFlow's Metric Callback
```
psfoley: I would suggest removing the `a`:

Example of MLFlow's Metric Callback

docs/log_metric_callback.rst
```diff
@@ -104,8 +104,8 @@ Follow the steps below to write your custom callback function instead. As an exa
         template : src.mnist_utils.callback_name
 
 
-Example of a Metric Callback
-============================
+Example of a TensorBoard's Metric Callback
```
psfoley: I would suggest removing the `a`:

Example of TensorBoard's Metric Callback

psfoley:
Thanks @akantak for the fixes! I left two comments about phrasing. After those are addressed, I think this is ready to merge

akantak:
fixed @psfoley

psfoley (approved):

PR #867, opened 2023-08-01
Update 'grpcio' to '1.53.0'

ashahba:
This PR fixes the following `CWE`'s:
https://cwe.mitre.org/data/definitions/440.html
https://cwe.mitre.org/data/definitions/617.html

Signed-off-by: Abolfazl Shahbazi <abolfazl.shahbazi@intel.com>

PR #868, opened 2023-08-01
Update Tensorflow, gRPC, Protobuf dependencies

psfoley:
This PR updates the Tensorflow workspace requirements to 2.13, which allows the use of significantly updated versions of the grpcio / protobuf packages

Fixes https://github.com/mlcommons/GaNDLF/pull/702

**Update**
In addressing the broken CI tests, this PR has also brought in the following:
- New `KerasSerializer` to address missing `build` function in Keras legacy optimizer
- Refactoring `openfl/utils` and `openfl/component/ca` to minimize unneeded imports before `fx workspace create` command. This is needed because on Windows when Python modules are loaded into memory they cannot be overwritten on disk. Because we do a pip installation from code, any version differences between loaded and new packages will throw an error. This has the extra benefit of `fx` CLI commands now responding near instantly.
- Fixing dockerization CI test by pinning pyzmq (thanks to @mansishr)
- Python 3.11 support


setup.py
kminhta, line 1: Will we also need to update `grpcio-tool>=1.56.2`? Referencing [setup.py #L159](https://github.com/psfoley/openfl/blob/tf-1.13/setup.py#L159) based on PR #867

.github/workflows/taskrunner.yml
```diff
@@ -24,10 +24,16 @@ jobs:
       uses: actions/setup-python@v3
       with:
         python-version: "3.8"
-    - name: Install dependencies
+    - name: Install dependencies ubuntu
+      if: matrix.os == 'ubuntu-latest'
```
ashahba, line 28: If we are on latest `ubuntu` or `windows` versions, we may want to have `python-version: "3.9"` or  even `python-version: "3.10" as our default test matrix environment.

Also we need to add a workflow for Python 3.11.

setup.py
psfoley: Done! Thanks @kta-intel

PR #870, opened 2023-08-03
pinned tensorboardX

porteratzo:
openfl 1.5 requires protobuf 3.19.6, tensorboardX 2.6 is compatible with that protobuf, after that newer protobuf is needed, 
tensorboardX 2.6.2 did not pin protobuf although it probably needs at least 4.22.3(inferred from 2.6.1 pin). 
we should pin tensorboardX<=2.6 since tensorboardX needs a higher version of protobuf anyway

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -153,7 +153,7 @@ setup(
         'rich',
         'scikit-learn',
         'tensorboard',
-        'tensorboardX',
+        'tensorboardX<=2.6',
         'tqdm',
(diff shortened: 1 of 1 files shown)


psfoley:
Approved. For other's context, this is a stopgap until https://github.com/securefederatedai/openfl/pull/868 is merged, and the version restriction for tensorboardX can be loosened again afterwards.

PR #895, opened 2023-11-08
Fixes #718 WandB in OpenFL and example to change number of epochs per round

CasellaJr:
With this PR it is possible to use WandB for tracking metrics of OpenFL experiments. For the Aggregator and each Collaborator (thanks to @Giemp95) will be created a new WandB run, that will be grouped under the name of the Aggregator for the sake of cleanliness. 
Moreover, in the PyTorch_TinyImageNet.ipynb you can find how to track the metrics, and how to modify the number of epochs per round.


CasellaJr:
Sorry @psfoley, there was a typo in the aggregator.py. I fixed it.

openfl/component/aggregator/aggregator.py
```diff
@@ -18,6 +18,8 @@
 from openfl.utilities import TensorKey
 from openfl.utilities.logs import write_metric
 
+import wandb
```
psfoley, line 21: This would require `wandb` as an import for OpenFL, which is something we want to avoid.

openfl/component/aggregator/aggregator.py
```diff
@@ -18,6 +18,8 @@
...
+LOG_WANDB = True
```
psfoley, line 22: This should be configurable by the definer of the experiment, i.e. pass this through the OpenFL Plan

openfl/component/aggregator/aggregator.py
```diff
@@ -55,6 +57,17 @@ def __init__(self,
                  log_metric_callback=None,
                  **kwargs):
         """Initialize."""
+        #INITIALIZE WANDB WITH CORRECT NAME
+        #The following variable, my_aggregator_name, can be changed to whatever you want. Right now I suppose that the name of the collaborators of the federation is in the format DATASETNAME_ENV_NUMBER
+        my_aggregator_name = '_'.join(set(element.split('_')[0] for element in authorized_cols))
+        if LOG_WANDB:
+            wandb.init(project="my_project", entity="my_group", group=f"{my_aggregator_name}", tags=["my_tag"],
```
psfoley, line 64: The name of the project, entity, group, etc. should also be configurable through the plan

openfl/component/aggregator/aggregator.py
```diff
@@ -55,6 +57,17 @@ def __init__(self,
                  log_metric_callback=None,
                  **kwargs):
         """Initialize."""
+        #INITIALIZE WANDB WITH CORRECT NAME
+        #The following variable, my_aggregator_name, can be changed to whatever you want. Right now I suppose that the name of the collaborators of the federation is in the format DATASETNAME_ENV_NUMBER
+        my_aggregator_name = '_'.join(set(element.split('_')[0] for element in authorized_cols))
```
psfoley, line 62: This should be moved into the `if LOG_WANDB` block because it's not used elsewhere. Also, it's a bit confusing that the `my_aggregator_name` variable is a combined list of collaborators. May want to modify this to:

```suggestion
        my_aggregator_name = 'Aggregator_' + '_'.join(set(element.split('_')[0] for element in authorized_cols))
```

openfl/component/aggregator/aggregator.py
```diff
@@ -55,6 +57,17 @@ def __init__(self,
...
+        if LOG_WANDB:
+            wandb.init(project="my_project", entity="my_group", group=f"{my_aggregator_name}", tags=["my_tag"],
+                       config={
+                           "num_clients": 4,
+                           "rounds": 100
```
psfoley, line 67: I'm not familiar with the wandb configuration, but the same guidance applies here as earlier comments. I would expect `num_clients` and `rounds` to be configurable or based on existing parameters used elsewhere

openfl/component/collaborator/collaborator.py
```diff
@@ -14,6 +14,8 @@
 from openfl.protocols import utils
 from openfl.utilities import TensorKey
 
+import wandb
+LOG_WANDB = True
```
psfoley, line 18: Same comment as above. Should only import this if `wandb` is required in the plan

openfl/component/collaborator/collaborator.py
```diff
@@ -133,6 +135,14 @@ def set_available_devices(self, cuda: Tuple[str] = ()):
 
     def run(self):
         """Run the collaborator."""
+        if LOG_WANDB:
+            wandb.init(project="my_project", entity="my_group", tags=["my_tags"],
+                       config={
+                           "num_clients": 4,
+                           "rounds": 100,
+                       },
+                       name=self.collaborator_name
```
psfoley, line 144: Same comment as above. Make project, entity, tags configurable. Use client count and rounds from existing plan variables.

psfoley (changes requested):
Hi @CasellaJr - thanks for this contribution. There's some changes that are required before this can be approved. Those changes fall into two categories:
1. `wandb` cannot be assumed to be present in all federations. It should be made optional, and ideally this would be configurable in the FL Plan. The `LOG_WANDB` variable you defined I think aims to make it's use optional, but this is not something a user could set without modifying `aggregator.py` or `collaborator.py` directly in the current version. See the [`get_aggregator`](https://github.com/securefederatedai/openfl/blob/develop/openfl/federated/plan/plan.py#L328-L355) function in `plan.py` as an example of how the aggregator object gets constructed, and how the plan.yaml parameters are used to initialize the object. 
2. Hard coded variables. The wandb project name and configuration should be made configurable (again through the plan).

theakshaypant:
@CasellaJr can you please resolve the comments/conflicts and request for review again?

psfoley:
Hi @CasellaJr -  OpenFL will be adding a callback API in 1.8 so that WandB and other MLOps tools can be more easily inserted into the experiment. Closing this for now

PR #897, opened 2023-11-27
incorporating support for federated evaluation 

kminhta:
Request: #788 

This PR adds an additional workspace template for federated evaluation using `torch_cnn_mnist` as a base

Changes made to `plan/plan.yaml` to set `rounds_to_train` to 1 and set `aggregator_model_validation` as the only task to be executed

Changes made to `src/pt_cnn.py` to remove optimizer initialization and loss function, which are no longer needed for a single inference run

Added .rst file to docs to introduce federated eval and walk through example (**Note**: landing page TBD)

### DIFF
--- a/docs/federated_evaluation.rst
+++ b/docs/federated_evaluation.rst
@@ -34,7 +34,7 @@ In order to adapt this template for federated evaluation, the following modifica
 
 .. code-block:: yaml
 
-    # Copyright (C) 2023 Intel Corporation
+    # Copyright (C) 2020-2023 Intel Corporation
     # Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
     
     aggregator :
--- a/docs/federated_evaluation.rst
+++ b/docs/federated_evaluation.rst
@@ -1,6 +1,5 @@
-
 Federated Evaluation with OpenFL
-==============================
+=================================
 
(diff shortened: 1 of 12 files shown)


psfoley:
@kta-intel the federated evaluation documentation you've added here is excellent. Can you confirm that the page is indexed correctly (i.e. shown on the side pane when you build locally)? If so this is ready to merge.

kminhta:
> @kta-intel the federated evaluation documentation you've added here is excellent. Can you confirm that the page is indexed correctly (i.e. shown on the side pane when you build locally)? If so this is ready to merge.

Great catch! I added it under the [manual](https://github.com/securefederatedai/openfl/blob/202bf02cd703a0ffaf957ae74eabf910a10b10ed/docs/manual.rst?plain=1#L38) tab. Let me know if this looks good or if you think it should be added elsewhere instead. It builds locally without issue

psfoley:
Thanks for the quick turnaround Kevin. [manual](https://github.com/securefederatedai/openfl/blob/202bf02cd703a0ffaf957ae74eabf910a10b10ed/docs/manual.rst?plain=1#L38) is a good spot for the documentation. Only items remaining are 3 minor linting errors. Once those are fixed this can be merged.

psfoley (approved):

PR #902, opened 2023-12-14
Update README.md

sbakas:
Made changes to reflect the transition of the Bakas' group from UPenn to IU.


psfoley (approved):
Thanks for the clarifications and updates to the background, @sbakas! Approved

PR #926, opened 2024-02-16
Update Ray

psfoley:
This PR:
- Updates Ray dependency for Workflow API to 2.9.2
- Removes the extra requirements file from the MedMNIST example


ifahim:
Looks good to me.

PR #937, opened 2024-03-04
Install Python 3.10 from deadsnakes PPA

ashahba:
This PR attempts to reduce number of `Trivy` scan issues by building and installing latest `Python 10` version from source.

### DIFF
--- a/openfl-docker/Dockerfile.base
+++ b/openfl-docker/Dockerfile.base
@@ -20,47 +20,40 @@ FROM base as python-dev
 
 ARG PYTHON=python3.10
 ARG PIP=pip3.10
-ARG PYTHON_VER=3.10.13
 
 RUN apt-get update && \
     apt-get install --no-install-recommends --fix-missing -y \
-      build-essential \
-      ca-certificates \
       curl \
-      libbz2-dev \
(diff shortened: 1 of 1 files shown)


openfl-docker/Dockerfile.base
```diff
@@ -53,9 +111,9 @@ WORKDIR /openfl
 COPY . .
 
 # Install pip
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && rm -rf get-pip.py
-RUN pip install --no-cache-dir . 
+RUN pip install --no-cache-dir . && \
+    pip uninstall -y setuptools
```
psfoley: `setuptools` is a required package: https://github.com/securefederatedai/openfl/actions/runs/8137042934/job/22234556956?pr=937#step:5:12376

psfoley (changes requested):
Thanks @ashahba for helping address the Trivy issues. I've added a single review comment about adding back `setuptools` because the removal is breaking OpenFL in docker

PR #944, opened 2024-03-13
multiple collaborators per certificate

hasan7n:
This PR adds the possibility of having multiple collaborator nodes use the same certificate for communication with the aggregator. The goal is to separate the dataset "identity" from its owner identity. We can think of "collaborator name" now as a "dataset name/ID".

`cols.yaml`'s `collaborators` key is now expected to have one of the following two structures:
- A dictionary representing a mapping between an authorized collaborator name and the corresponding certified common name. Multiple collaborator names can have the same certified common name.
- list of strings: as before. In this case, each string will be interpreted as both the collaborator name and its certified common name. (for backward compatibility)

In the code, the `Plan` object will now have both the `authorized_cols` property and a new property `cn_mapping`. I kept the `authoized_cols` property to minimize changes to the codebase since it is used in many places as a list of strings.


hasan7n:
possible TODOs:
1. unit tests
2. Update `fx collaborator certify` and `generate-cert-request` to accept multiple cols per cert.
3. Director-based flow: I haven't tested this yet. It looks like the list of collaborators is received via a request. Should we add a new optional request param for associated expected CNs?
4. Experimental code is not updated
5. documentation

openfl/component/aggregator/aggregator.py
```diff
@@ -225,7 +227,7 @@ def valid_collaborator_cn_and_id(self, cert_common_name,
         # FIXME: '' instead of None is just for protobuf compatibility.
         #  Cleaner solution?
         if self.single_col_cert_common_name == '':
-            return (cert_common_name == collaborator_common_name
+            return (cert_common_name == self.cn_mapping[collaborator_common_name]
```
hasan7n, line 230: TODO: handle error

hasan7n: possible key error

manuelhsantana:
Hi @hasan7n,

Great work on enabling multiple collaborators to use a shared certificate. To ensure everything works as expected, could you add unit tests for both the new dictionary mapping in cols.yaml and the backward-compatible list format?

Additionally, could you outline how we might manually test this feature to verify its functionality?

Thanks for your efforts!

Best, Manuel Santana

Issue #855, opened 2023-06-30
Code example uses image of text instead of text


ChrisXKeroack:
**Describe the bug**
Code example uses considerable amount of image of text in use for what should be presented as text

**To Reproduce**
1. https://openfl.readthedocs.io/en/latest/install.html 
2. start a screenreader
3. Scroll down to  ‘output of the fx command’ graphic

notice:  ‘output of the fx command’ graphic gives its filename as its alt text, and is conveying considerable image-of-text information  that should also be available in text (sighted user will get information that person reliant on screenreader can’t)

**Expected behavior**
Text content should be presented in a code block similar to other commandline instructions demonstrated elsewhere on the page.

**Screenshots**
![image](https://github.com/securefederatedai/openfl/assets/104644211/30ca4d1f-9321-4ace-aa33-5bcb6d998706)

**Desktop:**
This issue is not OS or browser specific, but for tracking purposes, testing was done on Win11, in Edge Version 114.0.1823.58 (Official build) (64-bit), and the screenreader in use was the common 3rd party screenreader NVDA (http://www.nvaccess.org).

**Smartphone:**
This issue is not device, OS, or browser specific.

**Additional context**
This is an accessibility issue, please annotate this bug with the 'accessibility' label.

aryanxk02:
I can take up this issue @psfoley.

theakshaypant:
Closed with #960

PR #960, opened 2024-04-04
openfl bash message

aryanxk02:
This PR adds a code block to the installation.rst file, as mentioned in #855, the terminal image of the installation has been replaced with the actual content.


psfoley (approved):
Thank you for fixing this, @aryanxk02! Approved

PR #963, opened 2024-04-26
Stop suppressing gandlf config errors

VukW:
If gandlf config cannot be parsed properly (typos, wrong values, etc), now the error is just suppressed without any meaningful error and `gandlf_config` variable is just broken.

Fix ensures
1. runner fails if gandlf config cannot be parsed,
2. if any errors, they are printed to logs with their stack trace


psfoley (approved):

psfoley:
Thanks for the contribution @VukW! Approved and merged.

PR #967, opened 2024-05-02
Minor updates to docker Makefile and scan job

ashahba:
This PR:
- updates `Makefile` and `Trivy` workspace file so that the `docker build` always pulls latest base `ubuntu:22.04` first
-  updates `Makefile` so that the container is saved in `gzip` format rather than `tar` so the saved image has a smaller footprint:
```
$ du -sh openfl.tar*
1.2G 	openfl.tar
463M	openfl.tar.gz
```
- Removes stock ssh keys installed under `/etc/ssh/` by openssh server, since they are not needed and also `Trivy` flags them incorrectly as `secrets` left on the container.

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -17,7 +17,7 @@ jobs:
 
       - name: Build an image from Dockerfile
         run: |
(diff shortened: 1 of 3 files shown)


psfoley (approved):
Thank you @ashahba! Approved

PR #980, opened 2024-05-30
Standardize linting and formatting

MasterSkepticista:
Note: This PR only impacts `openfl.experimental` namespace, and should be a non-breaking change. 

Changes:
* Sorted imports using `isort`
* Formatted files using `black`
* Linted using `flake8`

Configuration for `isort` and `black` is included in `pyproject.toml`. `flake8` continues to use `setup.cfg`

For future development:
* Lint using `./shell/lint.sh`
* Format using `./shell/format.sh`
* Use absolute imports as discussed [here](https://github.com/securefederatedai/openfl/pull/980#discussion_r1628970037).

### DIFF
--- a/openfl/experimental/runtime/__init__.py
+++ b/openfl/experimental/runtime/__init__.py
@@ -2,8 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 """ openfl.experimental.runtime package Runtime class."""
 
-from .federated_runtime import FederatedRuntime
-from .local_runtime import LocalRuntime
-from .runtime import Runtime
-
-__all__ = ["FederatedRuntime", "LocalRuntime", "Runtime"]
+from openfl.experimental.runtime.federated_runtime import FederatedRuntime
+from openfl.experimental.runtime.local_runtime import LocalRuntime
+from openfl.experimental.runtime.runtime import Runtime
--- a/openfl/experimental/runtime/__init__.py
+++ b/openfl/experimental/runtime/__init__.py
@@ -3,9 +3,8 @@
 
 """ openfl.experimental.runtime package Runtime class."""
 
-from .runtime import Runtime
-from .local_runtime import LocalRuntime
 from .federated_runtime import FederatedRuntime
-
+from .local_runtime import LocalRuntime
+from .runtime import Runtime
 
 __all__ = ["FederatedRuntime", "LocalRuntime", "Runtime"]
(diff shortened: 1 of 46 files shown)


psfoley (commented):
Thanks @MasterSkepticista. Looks like there's just a couple remaining lint issues before approval. Will the `namex` changes be included in a future PR?

MasterSkepticista:
Yes, this PR is for non-breaking formatting changes. `namex` will impact API surface, and may impact project structure. That will be for a future PR.

MasterSkepticista:
@psfoley Added relevant ignores in `setup.cfg`

Could not get rid of `F821` without invoking other warnings. `noqa: #` does not seem to work with multi-line code.
How does `ModelHeader` and `TensorProto` not throw an error when code reaches there? The error is valid, no definitions for those are imported anywhere in the file.

openfl/experimental/runtime/__init__.py
```diff
@@ -1,11 +1,9 @@
...
 # SPDX-License-Identifier: Apache-2.0
-
 """ openfl.experimental.runtime package Runtime class."""
 
-from .runtime import Runtime
-from .local_runtime import LocalRuntime
 from .federated_runtime import FederatedRuntime
-
+from .local_runtime import LocalRuntime
+from .runtime import Runtime
```
psfoley: Changing the import order is leading to this circular import error (confirmed this by changing the order manually locally on develop, which is passing). This should either be reverted or fixed: https://github.com/securefederatedai/openfl/actions/runs/9381169993/job/25829824994?pr=980#step:5:19

MasterSkepticista: Since `__init__.py` files are useful for API exports, until we transition to API export decorators, we need to:
* Export classes using absolute paths:
```python
# Before (order-sensitive, fails)
from .federated_runtime import FederatedRuntime
from .local_runtime import LocalRuntime
from .runtime import Runtime
__all__ = ["FederatedRuntime", "LocalRuntime", "Runtime"]

# After (order-agnostic)
from openfl.experimental.runtime.federated_runtime import FederatedRuntime
from openfl.experimental.runtime.local_runtime import LocalRuntime
from openfl.experimental.runtime.runtime import Runtime
```
* Use absolute imports in source code:
```python
# Before (order sensitive, fails)
from openfl.experimental.runtime import Runtime

# After (order-agnostic)
from openfl.experimental.runtime.runtime import Runtime
```

This gives the intended behavior: users can access imported objects without long paths in user code, and order of imports within source code does not lead to errors.

This fix will be used for all subsequent tests.

MasterSkepticista:
Fixed imports in the entire `openfl.experimental` interface.

psfoley (approved):

PR #983, opened 2024-06-05
Allow `feature_shape` to be passed to `fx plan initialize`

psfoley:
This PR:
- Fixes the `feature_shape` argument that can be passed to `fx plan initialize`. By passing this argument (which allows for list types arguments, such as `[1,28,28]`), loading the data loader for the model owner / aggregator can be avoided. This is important because the aggregator may not have access to local data to generate the initial model.

### DIFF
--- a/openfl/interface/plan.py
+++ b/openfl/interface/plan.py
@@ -65,8 +65,6 @@ def initialize(context, plan_config, cols_config, data_config,
             echo(f'{p} is out of the openfl workspace scope.')
             sys.exit(1)
 
-    logger.info("input shape = {input_shape}")
-
     plan_config = Path(plan_config).absolute()
     cols_config = Path(cols_config).absolute()
     data_config = Path(data_config).absolute()
--- a/openfl/interface/plan.py
+++ b/openfl/interface/plan.py
@@ -12,6 +12,7 @@ from click import pass_context
 from click import Path as ClickPath
 
 from openfl.utilities.path_check import is_directory_traversal
+from openfl.utilities.click_types import ListOption
 
 logger = getLogger(__name__)
 
@@ -36,8 +37,8 @@ def plan(context):
         default='plan/data.yaml', type=ClickPath(exists=True))
 @option('-a', '--aggregator_address', required=False,
         help='The FQDN of the federation agregator')
-@option('-f', '--feature_shape', required=False,
-        help='The input shape to the model')
+@option('-f', '--feature_shape', cls=ListOption, required=False,
+        help='The input shape to the model (i.e. [1,28,28])')
 @option('-g', '--gandlf_config', required=False,
         help='GaNDLF Configuration File Path')
 def initialize(context, plan_config, cols_config, data_config,
@@ -54,6 +55,9 @@ def initialize(context, plan_config, cols_config, data_config,
     from openfl.protocols import utils
     from openfl.utilities.split import split_tensor_dict_for_holdouts
     from openfl.utilities.utils import getfqdn_env
+    from openfl.utilities.mocks import MockDataLoader
+
+    #logger.info(f'Feature_shape = {feature_shape}')
 
     for p in [plan_config, cols_config, data_config]:
         if is_directory_traversal(p):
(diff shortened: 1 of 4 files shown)


openfl/interface/plan.py
```diff
@@ -54,6 +55,7 @@ def initialize(context, plan_config, cols_config, data_config,
     from openfl.protocols import utils
     from openfl.utilities.split import split_tensor_dict_for_holdouts
     from openfl.utilities.utils import getfqdn_env
+    from openfl.utilities.mocks import MockDataLoader
```
MasterSkepticista: Imports within functions are generally discouraged. Can we move these to the top?
I recognize this needs a fix throughout the repository.

openfl/utilities/click_types.py
```diff
@@ -3,6 +3,7 @@
 """Click types module."""
```
MasterSkepticista: Nit, module one-liner summary: "Custom input type definitions for Click"

openfl/utilities/click_types.py
```diff
@@ -30,6 +31,17 @@ def convert(self, value, param, ctx):
             self.fail(f'{value} is not a valid ip adress name', param, ctx)
         return value
 
+class ListOption(click.Option):
+
```
MasterSkepticista, line 34: One-liner docstring, to be consistent with other classes in this file.

openfl/interface/plan.py
```diff
@@ -36,8 +37,8 @@ def plan(context):
         default='plan/data.yaml', type=ClickPath(exists=True))
 @option('-a', '--aggregator_address', required=False,
         help='The FQDN of the federation agregator')
-@option('-f', '--feature_shape', required=False,
-        help='The input shape to the model')
+@option('-f', '--feature_shape', cls=ListOption, required=False,
```
MasterSkepticista: Could be renamed to `input_shape`?

MasterSkepticista (commented):
While this method works for single-input models with static shapes, it may not work for multi-tensor input models, or single-input models with custom input `dtype`s.

A general approach could be for the aggregator to take `InputSpec` as an argument. `InputSpec` is could be a dictionary of list of `TensorSpec[TensorShape, DType]`.

Example multi-input-output model:
```shell
# Can generalize if `ast` parses it as a dict instead
--input_shape {'input_0': [1, 240, 240, 4], 'output_1': [1, 240, 240, 1]}

# Assumes first input to take the given shape, if `ast` parses as list.
--input_shape [1, 240, 240, 1]
```

openfl/interface/plan.py
psfoley: Fixed.

psfoley:
> While this method works for single-input models with static shapes, it may not work for multi-tensor input models, or single-input models with custom input `dtype`s.
> 
> A general approach could be for the aggregator to take `InputSpec` as an argument. `InputSpec` is could be a dictionary of list of `TensorSpec[TensorShape, DType]`.
> 
> Example multi-input-output model:
> 
> ```shell
> # Can generalize if `ast` parses it as a dict instead
> --input_shape {'input_0': [1, 240, 240, 4], 'output_1': [1, 240, 240, 1]}
> 
> # Assumes first input to take the given shape, if `ast` parses as list.
> --input_shape [1, 240, 240, 1]
> ````

This is a good point. I tested that the existing code (that makes use of `ast.literal_eval`) can handle dictionaries as well. The main caveat is that the dictionary needs to be wrapped in quotes to be passed via command line. I've tried to make it clear in the help string that this is a requirement, and provided your example (wrapped in double quotes) as a reference for users.

kminhta:
This looks good to me, overall. It successfully avoids the need of loading the data at the aggregator to initialize the model, which is a great addition.

One comment: this won't always guarantee that the model is initialized using the specified input values or that the dataloader will load in data with the specified shape at the collaborators. For example, the `torch_cnn_mnist` workspace hard codes the model parameters in [src/taskrunner.py](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch_cnn_mnist/src/taskrunner.py#L36) so it will ultimately ignore the values passed through `fx plan initialize --input_shape` and initialize regardless, and even if we use `.get_feature_shape()` during initialization, only the channel size matters since the HxW isn't required in for torch conv layers.

**In my opinion, this is not a breaking issue and this PR is good to go.** However, going forward we should consider ways to reduce any potential confusion caused by this from a user perspective, such as having the existing workspace templates erroring out if an invalid `input_shape` is provided or issuing a user warning if it is ignored.

openfl/utilities/mocks.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2020-2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+"""Mock objects to eliminate extraneous dependencies"""
+
+
+class MockDataLoader:
```
teoparvanov, line 6: I suppose you don't extend `DataLoader`, because only the `get_feature_shape()` method is required during plan initialization? In this case, would it make sense to call this class `FeatureShapeLoader`?

teoparvanov (approved):
Looks great, thanks @psfoley !

Issue #973, opened 2024-05-15
Update and consolidate Keras and TensorFlow taskrunner ~~openfl-workspaces under a single TensorFlowTaskRunner class~~


kminhta:
_2024-12-3 edit: update Keras Task Runner to support 3.x directly. It should rely only on Keras API to leave the door open to support keras-TF, keras-JAX, keras-torch workspaces_

---
The current [TensorFlow TaskRunner](https://github.com/securefederatedai/openfl/blob/develop/openfl/federated/task/runner_tf.py) is outdated and currently only supports 1.x. The current [Keras TaskRunner](https://github.com/securefederatedai/openfl/blob/develop/openfl/federated/task/runner_keras.py) is also outdated and only supports TF implementation. 

Since stock Keras has support for multiple frameworks (i.e. torch, tf, and JAX), and the recommended practice by the TF team is to use [Keras by default with TF](https://www.tensorflow.org/guide/keras), **the current keras and tensorflow taskrunner openfl-workspaces should be updated and consolidated under a single ~~tensorflow~~ keras taskrunner class**

This will require:
- Updating necessary [openfl.federated.task](https://github.com/securefederatedai/openfl/tree/develop/openfl/federated/task) files (i.e. updating KerasTaskRunner ~~to TensorFlowTaskrunner and removing/deprecating the original TensorFlowTaskrunner~~)
- Updating workspaces (tf_* and keras_* workspaces should run on the new taskrunner, keras_* workspaces should update naming convention to tf)
- Fixing how optimizer weights are being updated because legacy optimizer is [no longer be supported](https://github.com/keras-team/keras/blob/master/keras/src/optimizers/__init__.py#L113) in the latest keras

kminhta:
Completed by #1174

PR #985, opened 2024-06-06
Update TensorFlow Task Runner and related workspaces

kminhta:
Related Issue: #973 

**Summary:**
This PR aims to update the TensorFlow Task Runner to use Keras as the high-level API, which is in line with [best practices](https://www.tensorflow.org/guide/keras) as well as updates existing TF workspaces. This enables the usage of non-legacy optimizers (which will be deprecated in future versions of TF/Keras)

**Specifically, this PR:**
- Creates a new `TensorFlowTaskRunner` class in `openfl.federated.task.runner_tf` which borrows heavily from the `KerasTaskRunner` task. Major difference is in handling the weights of the optimizer which was necessitated by the removal of the `.get_weight()` and `.weights()` attributes from the optimizer. This new `TensorFlowTaskRunner` extracts weights from the `.variables()` attribute
   - Also updated the train and validation task names to `train_validation` and `task_validation` to be consistent with the torch taskrunner
- Archived old `TensorFlowTaskRunner` as `TensorFlowTaskRunner_v1` within `openfl.federated.task.runner_tf` and updated the `__init__` files to make it callable. Rationale is to avoid any breaking changes for tutorials or upstream applications that still relied on the low-level TF taskrunner. This can be removed entirely in a future release as needed
  - Also updated the train and validation task names to `train_validation` and `task_validation` to be consistent with the torch taskrunner
- Created a new `tf_cnn_mnist` workspace and updated the `torch_cnn_histology` workspace to run on the new `TensorFlowTaskRunner` using the `src/dataloader.py` and `src/taskrunner.py` convention.
  - update to `TensorFlow v2.15.1` (latest TensorFlow to not use `Keras v3.x` by default)
- Minor `tf_3dunet_brats` to use new `TensorFlowTaskRunner` (did not make changes to `src` files because I did not have Brats3D dataset to verify a large update
- Minor updates to `tf_2dunet` to run on archived `TensorFlowTaskRunner_v1` 

**Future work**
- Consolidation step still needed:
  - Migrate all `tf_2d_unet` from `TensorFlowTaskRunner_v1`  to new `TensorFlowTaskRunner` 
  - Migrate all keras workspaces from `KerasTaskRunner` to new `TensorFlowTaskRunner` and remove/archive KerasTaskRunner
-  Look into updated `TensorFlowTaskRunner` to run on `TF v2.16+` with `Keras 3.x` (this may need some large changes to weight handling that will likely not have backwards compatibility)


openfl-workspace/tf_2dunet/src/tf_2dunet.py
MasterSkepticista, line 1: Can we rename this to `TensorFlowTaskRunnerLegacy` instead, or a dedicated legacy task runner module? `v1` might cause confusion with `TF 1.x` API. Open to other ideas though.

openfl-workspace/tf_3dunet_brats/src/tf_3dunet_model.py
```diff
@@ -5,14 +5,14 @@
 
 import tensorflow as tf
 
-from openfl.federated import KerasTaskRunner
+from openfl.federated import TensorFlowTaskRunner
 from .define_model import build_model
```
MasterSkepticista: Though this is not a change coming from this PR, we should migrate to using absolute imports as-encountered. Relevant discussion [here](https://github.com/securefederatedai/openfl/pull/980#discussion_r1628970037).

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -17,40 +19,47 @@ class TensorFlowCNN(KerasTaskRunner):
...
-                     validation_smoothing=1.0,
-                     **kwargs):
-        """Create the TensorFlow CNN Histology model.
+        self.model.summary(print_fn=self.logger.info)
+
+        self.logger.info(f'Train Set Size : {self.get_train_data_size()}')
+        self.logger.info(f'Valid Set Size : {self.get_valid_data_size()}')
+
+    def build_model(self,
+                    input_shape,
```
MasterSkepticista, line 37: Missing argument type hints

openfl/federated/__init__.py
```diff
@@ -3,14 +3,15 @@
...
 
-import pkgutil
+import importlib.util
 from .plan import Plan  # NOQA
 from .task import TaskRunner  # NOQA
 from .data import DataLoader  # NOQA
 
-if pkgutil.find_loader('tensorflow'):
-    from .task import TensorFlowTaskRunner, KerasTaskRunner, FederatedModel  # NOQA
+if importlib.util.find_spec('tensorflow'):
```
MasterSkepticista, line 11: How about a simple `try...except` block to check imports?

openfl-workspace/tf_cnn_histology/src/dataloader.py
```diff
@@ -11,6 +12,33 @@
 logger = getLogger(__name__)
 
 
+class TensorFlowHistologyInMemory(TensorFlowDataLoader):
```
MasterSkepticista: General recommendation on naming: The user would know these tutorials are for TensorFlow and Histology dataset. It helps to name functions/classes that are generic and expressive to what they do. Redundant naming increases cognitive overhead for the user.

In this case, `HistologyDataLoader` does a better job IMO.

openfl-workspace/tf_cnn_histology/src/dataloader.py
```diff
@@ -11,6 +12,33 @@
...
+
+        self.X_train = X_train
+        self.y_train = y_train
+        self.X_valid = X_valid
+        self.y_valid = y_valid
+
+        self.num_classes = num_classes
+
+
 def one_hot(labels, classes):
```
MasterSkepticista: Type hints and docstring.

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -3,12 +3,14 @@
...
+import numpy as np
 import tensorflow as tf
 
-from openfl.federated import KerasTaskRunner
+from openfl.utilities import Metric
+from openfl.federated import TensorFlowTaskRunner
 
 
-class TensorFlowCNN(KerasTaskRunner):
+class TensorFlowCNN(TensorFlowTaskRunner):
```
MasterSkepticista: `SimpleCNN`?

Also class docstrings explaining attributes and expected usage.

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -17,40 +19,47 @@ class TensorFlowCNN(KerasTaskRunner):
     """
 
     def __init__(self, **kwargs):
-        """Initialize.
+        """
+        Initialize.
 
         Args:
             **kwargs: Additional parameters to pass to the function
```
MasterSkepticista: Ambiguous. Mention `**kwargs` are forwarded to `TensorFlowTaskRunner` base class (per my understanding)

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -17,40 +19,47 @@ class TensorFlowCNN(KerasTaskRunner):
...
+                    input_shape,
+                    num_classes,
+                    **kwargs):
+        """
+        Define the model architecture.
 
         Args:
-            training_smoothing (float): (Default=32.0)
-            validation_smoothing (float): (Default=1.0)
+            input_shape (numpy.ndarray): The shape of the data
```
MasterSkepticista: Shapes are usually not `np.ndarray`s. I think you mean a `tuple` or `list` instance, in which case, type hint could be `input_shape (List[int])`

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -17,40 +19,47 @@ class TensorFlowCNN(KerasTaskRunner):
...
+            num_classes (int): The number of classes of the dataset
             **kwargs: Additional parameters to pass to the function
 
+        Returns:
+            keras.src.engine.functional.Functional
+
         """
         print(tf.config.threading.get_intra_op_parallelism_threads())
         print(tf.config.threading.get_inter_op_parallelism_threads())
```
MasterSkepticista: These two lines are neither relevant to the example nor OpenFL. Can be removed to reduce cognitive overhead.

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -17,40 +19,47 @@ class TensorFlowCNN(KerasTaskRunner):
     """
 
     def __init__(self, **kwargs):
-        """Initialize.
+        """
```
MasterSkepticista: These docstrings can be removed once class attributes' docstring is added.

openfl-workspace/tf_cnn_mnist/plan/data.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2020-2021 Intel Corporation
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+# collaborator_name,data_directory_path
+one,1
```
MasterSkepticista, line 5: General: This has been conventionally done for quite some time in OpenFL, but we should transition to using more expressive names and paths to reduce users' cognitive overhead.

openfl-workspace/tf_cnn_mnist/plan/plan.yaml
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) 2020-2024 Intel Corporation
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+aggregator :
+  defaults : plan/defaults/aggregator.yaml
+  template : openfl.component.Aggregator
+  settings :
+    init_state_path : save/tf_cnn_mnist_init.pbuf
+    best_state_path : save/tf_cnn_mnist_best.pbuf
+    last_state_path : save/tf_cnn_mnist_last.pbuf
```
MasterSkepticista: Redundant naming. It is OK to save it as `init_model.pbuf, best_model.pbuf, last_model.pbuf`.

These are saved within workspace directory IIRC, so the user already knows which experiment it belongs to

openfl-workspace/tf_cnn_mnist/src/__init__.py
MasterSkepticista, line 1: `__init__.py` can be kept empty.

openfl-workspace/tf_cnn_mnist/src/dataloader.py
```diff
@@ -0,0 +1,143 @@
...
+
+    logger.info(f'MNIST > X_train Shape : {X_train.shape}')
+    logger.info(f'MNIST > y_train Shape : {y_train.shape}')
+    logger.info(f'MNIST > Train Samples : {X_train.shape[0]}')
+    logger.info(f'MNIST > Valid Samples : {X_valid.shape[0]}')
+
+    if categorical:
+        # convert class vectors to binary class matrices
+        y_train = one_hot(y_train, num_classes)
+        y_valid = one_hot(y_valid, num_classes)
```
MasterSkepticista: Save lines!

```python
if one_hot:
  # Convert sparse to one-hot.
  y_train = np.eye(num_classes)[y_train]
  y_test = np.eye(num_classes)[y_test]
```

MasterSkepticista (changes requested):

MasterSkepticista:
Most of my comments are nitpicks around naming/formatting and/or structure. Please disposition as you find relevant.
Also, do we need TF v1 (`tf.Session` API) task runners? These are ancient APIs that none of the community uses. Does OpenFL have users on this legacy TF API?

psfoley:
> Most of my comments are nitpicks around naming/formatting and/or structure. Please disposition as you find relevant. Also, do we need TF v1 (`tf.Session` API) task runners? These are ancient APIs that none of the community uses. Does OpenFL have users on this legacy TF API?

@MasterSkepticista I agree the TF v1 / Legacy task runners can be removed at this point.

kminhta:
Thanks @MasterSkepticista @psfoley !
I will get back to addressing these comments once I'm done with some other pressing tasks. I appreciate the review

> > Most of my comments are nitpicks around naming/formatting and/or structure. Please disposition as you find relevant. Also, do we need TF v1 (`tf.Session` API) task runners? These are ancient APIs that none of the community uses. Does OpenFL have users on this legacy TF API?
> 
> @MasterSkepticista I agree the TF v1 / Legacy task runners can be removed at this point.

Great, I am also in agreement with removing the old TF runner. Actually, one reason why I did not propose it directly in this PR was because the tf2dunet workspaces uses the old TF runner. I can update it, but since it uses the BraTS dataset, I will have to get approval so I can test it

openfl-workspace/tf_cnn_mnist/plan/data.yaml
kminhta: agreed. actually, based on my understanding, the initial collaborator_name,data_directory_path in the data.yaml file is mainly for the aggregator to initialize with one of the data shards, since it is expected that the collaborators will update the path (and their name). With Patrick's new PR that allows users to specify input shape during plan initialization, this is actually not as necessary (I believe). Leaving it here for this PR though. Can do a sweep through and remove as needed later

openfl/federated/__init__.py
kminhta: That would look cleaner. Let's address in a separate PR, though. I actually only updated this because a warning kept popping up for me that pkgutil was being discontinued. Could even discard this commit if we don't want it

openfl-workspace/tf_2dunet/src/tf_2dunet.py
kminhta: v1 removed entirely and all tensorflow workspaces updated to v2

openfl-workspace/tf_3dunet_brats/src/tf_3dunet_model.py
kminhta: relative imports removed

openfl-workspace/tf_cnn_histology/src/dataloader.py
kminhta: Agreed! fixed in tf workspaces. We will need to go through all the other workspaces in a separate PR

openfl-workspace/tf_cnn_mnist/plan/plan.yaml
kminhta: Thanks! Fixed for TF workspaces. Will need a separate PR for other workspaces outside the scope of this PR

openfl-workspace/tf_cnn_mnist/src/dataloader.py
kminhta: Good catch, fixed.

openfl-workspace/tf_cnn_histology/src/taskrunner.py
kminhta: Agreed. This was copied over from an older code. Removed

openfl-workspace/tf_cnn_histology/src/taskrunner.py
kminhta: removed the docstring from the init and added docstring for class attributes

openfl-workspace/tf_cnn_histology/src/taskrunner.py
kminhta: thanks, fixed this

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
+        #  defined) then the model must be recompiled.
+        model_metrics_names = self.model.metrics_names
+
+        # TODO if there are new metrics in the flplan that were not included
+        #  in the originally
+        #  compiled model, that behavior is not currently handled.
+        for param in metrics:
+            if param not in model_metrics_names:
+                raise ValueError(
+                    f'KerasTaskRunner does not support specifying new metrics. '
```
teoparvanov, line 163: nit - I suppose you mean `TensorFlowTaskRunner` here?

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
-            _, s = self.validate_batch(X, y)
-            score += s * weight
+        vals = self.model.evaluate(
+            self.data_loader.get_valid_loader(batch_size),
+            verbose=1
+        )
+        model_metrics_names = self.model.metrics_names
+        if type(vals) is not list:
+            vals = [vals]
+        ret_dict = dict(zip(model_metrics_names, vals))
```
teoparvanov, line 204: I suppose TF guarantees that the metric names and values are in the same order, so we can safely zip together the two lists?

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
+
+        # TODO if there are new metrics in the flplan that were not included in
+        #  the originally compiled model, that behavior is not currently
+        #  handled.
+        for param in param_metrics:
+            if param not in model_metrics_names:
+                raise ValueError(
+                    f'KerasTaskRunner does not support specifying new metrics. '
```
teoparvanov, line 212: same question as a above - shouldn't this rather say `TensorFlowTaskRunner` ?

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
+        dict
+            The weight name list
+        """
+        if with_opt_vars:
+            weight_names = [weight.name for weight in obj.variables]
 
-        Args:
-            with_opt_vars (bool): Specify if we also want to get the variables
-             of the optimizer
+        weight_names = [weight.name for weight in obj.weights]
```
teoparvanov, line 259: For improved readability, I'd suggest to add an explicit `else:` block, as in
```python
if with_opt_vars:
   ...
else:
   ...
```

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
+        return weight_names
 
-        Returns:
-            dict: The weight dictionary {<tensor_name>: <value>}
+    @staticmethod
+    def _get_weights_dict(obj, suffix='', with_opt_vars=False):
+        """
+        Get the dictionary of weights.
 
+        Parameters
```
teoparvanov, line 267: Could you also document the `suffix` parameter? (its meaning isn't very obvious for the inexperienced OpenFL reader)

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
+            tensor_dict: the tensor dictionary
+            with_opt_vars (bool): True = include the optimizer's status.
+        """
+        if with_opt_vars is False:
+            # It is possible to pass in opt variables from the input tensor dict
+            # This will make sure that the correct layers are updated
+            model_weight_names = [weight.name for weight in self.model.weights]
+            model_weights_dict = {
+                name: tensor_dict[name] for name in model_weight_names
+            }
```
teoparvanov, line 358: this snippet seems identical with the one in the `else:` block, so why not move it before the `if` statement:
```python
model_weight_names = [weight.name for weight in self.model.weights]
model_weights_dict = {
    name: tensor_dict[name] for name in model_weight_names
}
```

openfl/federated/task/runner_tf.py
```diff
@@ -177,180 +133,350 @@ def train_batches(self, col_name, round_num, input_tensor_dict,
...
+
+            opt_weight_names = [
+                weight.name for weight in self.model.optimizer.variables
+            ]
+            opt_weights_dict = {
+                name: tensor_dict[name] for name in opt_weight_names
+            }
+            self._set_weights_dict(self.model, model_weights_dict)
```
teoparvanov, line 374: I guess this statement can also be moved outside of the `if` statement as it's executed in both cases:
```python
self._set_weights_dict(self.model, model_weights_dict)
```

openfl-workspace/tf_cnn_histology/src/dataloader.py
```diff
@@ -120,7 +134,7 @@ def load_histology_shard(shard_num, collaborator_count, categorical=True,
 
     if categorical:
         # convert class vectors to binary class matrices
-        y_train = one_hot(y_train, num_classes)
-        y_valid = one_hot(y_valid, num_classes)
+        y_train = np.eye(num_classes)[y_train]
+        y_valid = np.eye(num_classes)[y_valid]
```
teoparvanov, line 138: would this be equivalent to the following syntax (a bit more explicit IMHO):
```python
from keras.utils import to_categorical
y_train = to_categorical(y_train, num_classes)
y_valid = to_categorical(y_valid, num_classes)
```

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -96,13 +96,30 @@ def create_model(self,
...
+            list: Metric objects containing the computed metrics
+        """
+
+        history = self.model.fit(batch_generator,
+                                 verbose=1,
+                                 **kwargs)
+        results = []
+        for metric in metrics:
+            value = np.mean([history.history[metric]])
+            results.append(Metric(name=metric, value=np.array(value)))
```
teoparvanov, line 124: Out of curiosity, why do we take the mean value, as opposed to the latest one?

If we consider accuracy for instance, wouldn't we be interested in the accuracy at the end of the training (which would certainly be higher than the mean value)?

openfl-workspace/tf_cnn_mnist/src/dataloader.py
```diff
@@ -0,0 +1,129 @@
...
+
+    logger.info(f'MNIST > X_train Shape : {X_train.shape}')
+    logger.info(f'MNIST > y_train Shape : {y_train.shape}')
+    logger.info(f'MNIST > Train Samples : {X_train.shape[0]}')
+    logger.info(f'MNIST > Valid Samples : {X_valid.shape[0]}')
+
+    if categorical:
+        # convert class vectors to binary class matrices
+        y_train = np.eye(num_classes)[y_train]
+        y_valid = np.eye(num_classes)[y_valid]
```
teoparvanov, line 127: same here - would it make sense to use `to_categorical` from `keras.utils` instead?

openfl-workspace/tf_cnn_histology/src/taskrunner.py
```diff
@@ -96,13 +96,30 @@ def create_model(self,
             metrics=[tf.keras.metrics.SparseCategoricalAccuracy()],
         )
 
-        self.tvars = model.layers
-        print(f'layer names: {[var.name for var in self.tvars]}')
+        return model
 
-        self.opt_vars = self.optimizer.variables()
-        print(f'optimizer vars: {self.opt_vars}')
+    def train_(self, batch_generator, metrics: list = None, **kwargs):
```
teoparvanov, line 101: In this particular case, could we consider not even overriding `_train()`? It looks like in this instance, it practically does the same as the base method (seemingly differing in terms of validity checks)?

teoparvanov (commented):
Looks great, @kta-intel !

Please consider my review as a learning exercise of OpenFL and TF/Keras, so do take my comments and questions from that perspective 😊

tanwarsh:
Thank you @kta-intel. This was completed by https://github.com/securefederatedai/openfl/pull/1174.

PR #996, opened 2024-07-04
Updates to straggler handling functionality

### LINKED ISSUE #981: Potential issues observed in straggler handling policy
**Describe the bug**
Unexpected behavior observed when simulating the different straggler handling scenarios.

**Scenario:**
- Any one collaborator manually terminated during the execution of the experiment.
- Parameters for CutoffTimeBasedStragglerHandling policy:
      Cutoff time: 10 seconds
      Minimum reporting: 1
      Total collaborators: 2

**Expected behavior:**
- After the cutoff time (10 seconds) has elapsed aggregator should move to 2nd round as it has received results from minimum required collaborators (1 collaborator)

**Observed behavior:**
- Aggregator keeps waiting for the results from terminated collaborator even though Cutofftime has expired and Minimum reporting collaborators have reported the results

**To Reproduce**
1. Go to: openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
2. Change: straggler_cutoff_time parameter default value from np.inf to 10 in CutoffTimeBasedStragglerHandling class constructor.
3. Tutorial: Modify openfl-tutorials/interactive_api/PyTorch_LinearRegression tutorial to run with 2 collaborators and run the experiment
4. While the experiment is in progress terminate any one collaborator before it sends the results of the particular round.
5. See error: Aggregator will not start next round and will keep waiting for the terminated collaborators to send results.

**Desktop (please complete the following information):**
 - OS: Ubuntu (WSL)
 - OS Version: 22.04.4 LTS
 - Browser: Chrome
 - Browser Version: 125.0.6422.113 (Official Build) (64-bit)

[straggler-handling-policy-logs.zip](https://github.com/user-attachments/files/15511074/straggler-handling-policy-logs.zip)

ParthMandaliya:
This PR to resolves issue #981
 
**Summary of changes**
1. StragglerHandlingFunction: Added an interface method start_policy
2. CutoffTimeBasedStragglerHandling: start_policy method implements a timer to wait for cutoff-time and then call provided call callback method.
3. Aggregator: 
    - sendlocaltaskresults: update collaborators_done to keep track of collaborators that have finished ALL tasks
    - _straggler_cutoff_time_elapsed: call back function that is called after cutoff time has elapsed and applies the straggler policy
 
**Verification performed**

Tested various scenarios for time based cutoff straggler handling policy
All Director based workflow and Aggregator based workflow test-cases


openfl/component/aggregator/aggregator.py
```diff
@@ -331,11 +336,64 @@ def get_tasks(self, collaborator_name):
...
+    def _straggler_cutoff_time_elapsed(self) -> None:
+        """
+        This method is called by the straggler handling policy when cutoff timer is elapsed.
+        It applies straggler handling policy and ends the round early.
+
+        Returns:
+            None
+        """
+        # Stop timer from restarting until next round
+        self.straggler_handling_policy.is_timer_expired = True
```
teoparvanov: It would be cleaner if the policy object itself sets `is_timer_expired` to `True` when appropriate. It doesn't seem safe to leave this responsibility with the callback function, as the developer may not be aware of it. This also breaks the encapsulation of the policy object.

openfl/component/straggler_handling_functions/straggler_handling_function.py
```diff
@@ -5,15 +5,49 @@
 
 from abc import ABC
 from abc import abstractmethod
+from typing import Callable
 
 
 class StragglerHandlingFunction(ABC):
```
teoparvanov: This is no longer a pure functional interface - I'd suggest to rename it to `StragglerHandlingPolicy`

openfl/component/aggregator/aggregator.py
```diff
@@ -565,6 +623,49 @@ def send_local_task_results(self, collaborator_name, round_number, task_name,
...
+
+        # NOTE: Only for %age based straggler handling policy
+        # If minimum required collaborators have reported results
+        # ignore results from all other collaborators and add them in stragglers list
+        if (
+            not hasattr(self.straggler_handling_policy, "round_start_time") and
+            self.straggler_handling_policy.straggler_cutoff_check(
+                len(self.collaborators_done), self.authorized_cols
+            )
+        ):
```
teoparvanov: Regardless of the actual policy implementation, shouldn't we always add the lagging collaborators to the `stragglers` list (as soon as the `straggler_cutoff_check` triggers)?

openfl/component/straggler_handling_functions/straggler_handling_function.py
```diff
@@ -5,15 +5,49 @@
...
+        Args:
+            None
+
+        Returns:
+            None
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def straggler_cutoff_check(self, **kwargs) -> bool:
```
teoparvanov: Right now it's not clear what kind of inputs this method expects. Why not define it a bit more explicitly (while keeping the `**kwargs` argument for additional flexibility), as in:
```python
def straggler_cutoff_check(self, num_collaborators_done, all_collaborators, **kwargs) -> bool:
```
or even
```python
def straggler_cutoff_check(self, num_collaborators_done, num_all_collaborators, **kwargs) -> bool:
```
(for those policies, we most likely only need the list lengths, rather than the actual collaborator names)

teoparvanov (commented):
I think the code in this PR would work, but it introduces some tech debt - esp. with regards to the identified encapsulation "leaks" of the straggler policy objects. I've suggested some improvements in this first review pass, but it may be necessary to further revise the straggler policy interface.

openfl/component/aggregator/aggregator.py
ParthMandaliya: Accepted, the policy object now manages the flag.
NOTE: is_timer_expired is renamed to __is_policy_applied_for_round, and the flag is managed within the class.

openfl/component/straggler_handling_functions/straggler_handling_function.py
ParthMandaliya: The class is renamed as suggested.
DISCUSS: In our opinion name of the file and sub-package should be renamed. If this is acceptable then this can be done in a follow-up PR.

openfl/component/aggregator/aggregator.py
ParthMandaliya: The condition is now made generic for all straggler-handling policies.

openfl/component/straggler_handling_functions/straggler_handling_function.py
ParthMandaliya: Method definition is changed to "def straggler_cutoff_check(self, num_collaborators_done: int, num_all_collaborators: int, **kwargs) -> bool".

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -4,30 +4,130 @@
...
+        # time,
+        # then disregard the policy and wait for ALL collaborators to report
+        # results.
+        elif (
+            self.straggler_cutoff_time == np.inf
+            or (self.__straggler_time_expired() and self.__is_policy_applied_for_round)
+        ):
+            return num_all_collaborators == num_collaborators_done
+        # Something has gone, unhandled scenario, raising error.
+        raise ValueError("Unhandled scenario")
```
teoparvanov: for debugging purposes of the "unhandled" scenario, it would be useful to also include values of the parameters from the `if/else` statements above:
```python
self.__straggler_time_expired()
self.__is_policy_applied_for_round
self.straggler_cutoff_time
```

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -4,30 +4,130 @@
...
+class CutoffTimeBasedStragglerHandling(StragglerHandlingPolicy):
     def __init__(
         self,
         round_start_time=None,
         straggler_cutoff_time=np.inf,
         minimum_reporting=1,
         **kwargs
     ):
+        if minimum_reporting <= 0:
+            raise ValueError(f"minimum_reporting cannot be {minimum_reporting}")
```
teoparvanov: I'd add a clarification that `minimum_reporting` is expected to be a positive integer

openfl/component/straggler_handling_functions/percentage_based_straggler_handling.py
```diff
@@ -2,23 +2,51 @@
...
-class PercentageBasedStragglerHandling(StragglerHandlingFunction):
+class PercentageBasedStragglerHandling(StragglerHandlingPolicy):
     def __init__(
         self,
         percent_collaborators_needed=1.0,
         minimum_reporting=1,
         **kwargs
     ):
+        if minimum_reporting <= 0:
+            raise ValueError(f"minimum_reporting cannot be {minimum_reporting}")
```
teoparvanov: same here - let's explicitly state that a positive integer is expected

openfl/component/aggregator/aggregator.py
```diff
@@ -90,6 +87,10 @@ def __init__(self,
         self.write_logs = write_logs
         self.log_metric_callback = log_metric_callback
 
+        self.straggler_handling_policy = (
+            straggler_handling_policy or CutoffTimeBasedStragglerHandling()
```
teoparvanov: I suggest to explicitly set the `minimum_reporting` here to 50% of `len(authorized_cols)` when instantiating `CutoffTimeBasedStragglerHandling`. This is to account for federations with more than 2 collaborators.

teoparvanov (approved):
LGTM, thanks for improving the encapsulation of the policy objects! It does help with code readability, but it also increases our confidence about correctness.

PS: I've left a couple of final comments from my perspective. IMHO the remaining debt can be addressed in subsequent PRs.

openfl/component/straggler_handling_functions/straggler_handling_function.py
teoparvanov: Makes sense - as discussed on our sync meeting, it's ok to continue the clean-up in a separate PR.

openfl/component/aggregator/aggregator.py
teoparvanov: Awesome, thanks for fixing this!

openfl/component/straggler_handling_functions/straggler_handling_function.py
teoparvanov: Thanks, this is much better 😊

openfl/component/aggregator/aggregator.py
ParthMandaliya: Default parameter of `CutoffTimeBasedStragglerHandling()` ensure that the policy is disabled as cutoff time is set to infinite. Setting `minimum_reporting` to another value shall have no effect.

openfl/component/straggler_handling_functions/percentage_based_straggler_handling.py
ParthMandaliya: Changed message to `"minimum_reporting must be >0"`.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ParthMandaliya: Changed message to `"minimum_reporting must be >0"`.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ParthMandaliya: As you suggested, I have added 3 following statements in error message:

```
self.__straggler_time_expired()
self.__is_policy_applied_for_round
self.straggler_cutoff_time
```

openfl/component/aggregator/aggregator.py
teoparvanov: I understand, thanks for the clarification!

openfl/component/aggregator/aggregator.py
```diff
@@ -632,7 +669,22 @@ def send_local_task_results(
...
+            len(self.collaborators_done), len(self.authorized_cols)
+        ):
+            self.stragglers = [
+                collab_name
+                for collab_name in self.authorized_cols
+                if collab_name not in self.collaborators_done
+            ]
+            if len(self.stragglers) != 0:
+                self.logger.warning(f"Identified straggler collaborators: {self.stragglers}")
+            self._end_of_round_check()
```
teoparvanov: This code is an almost exact duplicate of lines 419-434. Would it make sense to extract it in a `self._end_of_round_with_stragglers_check()` method?

openfl/component/aggregator/aggregator.py
```diff
@@ -839,10 +876,11 @@ def _compute_validation_related_task_metrics(self, task_name):
...
             task_name, self.round_number
         )
-        # leave out stragglers for the round
+        # Leave out straggler for the round even if they've paritally
+        # completed given tasks
         collaborators_for_task = []
         for c in all_collaborators_for_task:
-            if self._collaborator_task_completed(c, task_name, self.round_number):
+            if c in self.collaborators_done:
                 collaborators_for_task.append(c)
```
teoparvanov: I'd suggest a more concise statement, using a list comprehension:
```python
collaborators_for_task = [c for c in all_collaborators_for_task if c in self.collaborators_done]
```

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -30,11 +31,100 @@ def __init__(
...
+        straggler_cutoff_time then return True, otherwise False.
+
+        Args:
+            num_collaborators_done: int
+                Number of collaborators finished.
+            num_all_collaborators: int
+                Total number of collaborators.
+
+        Returns:
+            bool
```
teoparvanov:
```python
"""
...
Returns:
        bool: True if the straggler cutoff conditions are met, False otherwise.
"""
```

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -30,11 +31,100 @@ def __init__(
...
+                self.logger.info(
+                    f"{num_collaborators_done} collaborators reported results within "
+                    "cutoff time. Applying cutoff policy and proceeding with end of round."
+                )
+                return True
+            self.logger.info(
+                "Disregarded straggler handling policy and waiting for minimum "
+                f"{self.minimum_reporting} collaborator(s) to report results."
+            )
+            return False
```
teoparvanov: I suggest to put this case under an explicit `else:` block

tests/openfl/component/aggregator/test_aggregator.py
```diff
@@ -171,80 +171,3 @@ def test_collaborator_task_completed_true(agg):
...
-    col2 = 'two'
-    agg.assigner.get_all_tasks_for_round = mock.Mock(return_value=[task_name])
-    agg.assigner.get_collaborators_for_task = mock.Mock(return_value=[col1, col2])
-    agg.collaborator_tasks_results = {
-        TaskResultKey(task_name, col1, round_num): 1,
-        TaskResultKey(task_name, col2, round_num): 1
-    }
-    is_round_done = agg._is_round_done()
-
-    assert is_round_done is True
```
teoparvanov, line 250: Shouldn't these tests be compatible (i.e. passing) with the "happy" scenario of the `CutoffTimeBasedStragglerHandling` policy?

teoparvanov (commented):
I have some additional questions and comments after the latest update.

openfl/component/aggregator/aggregator.py
ishant162: Consolidated duplicate code by updating it into the `_end_of_round_with_stragglers_check()` method.

openfl/component/aggregator/aggregator.py
ishant162: Updated as per your suggestion. Thanks!

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: Updated. Thanks!

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: Updated. Thanks!

tests/openfl/component/aggregator/test_aggregator.py
ishant162: Removed the unused `_is_round_done()` and `_is_task_done()` methods from the Aggregator, as they were identified as dead code. Additionally, updated the test case **tests/openfl/component/aggregator/test_aggregator.py** to remove any references to these removed methods.

openfl/component/aggregator/aggregator.py
```diff
@@ -140,6 +137,10 @@ def __init__(
         self.write_logs = write_logs
         self.log_metric_callback = log_metric_callback
 
+        self.straggler_handling_policy = (
+            straggler_handling_policy or CutoffTimeBasedStragglerHandling()
```
MasterSkepticista: If the intention is for `CutoffTimeBasedStragglerHandling` to be default, move this to `__init__`. User should not have to dig into the code to identify the default.

openfl/component/aggregator/aggregator.py
```diff
@@ -177,6 +178,10 @@ def __init__(
 
         self.collaborator_task_weight = {}  # {TaskResultKey: data_size}
 
+        # maintain a list of collaborators that have completed task and
+        # reported results in a given round
+        self.collaborators_done = []
+
```
MasterSkepticista, line 185: If this variable is only used by straggler policies, it should be stored in the policy object. Intent is to not spill a particular component's needs into other component's code. 

Example APIs: `self.straggler_policy.reset()` or `self.straggler_policy.mark_as_done(...)`

openfl/component/aggregator/aggregator.py
MasterSkepticista, line 1: Class attribute docstring is out of line with the actual arguments. Can you please fix that as well?

openfl/component/__init__.py
```diff
@@ -14,5 +14,5 @@
     PercentageBasedStragglerHandling,
 )
 from openfl.component.straggler_handling_functions.straggler_handling_function import (
```
MasterSkepticista, line 16: Even to a developer, this naming scheme is unintuitive. How can we expect users to remember?
Please eliminate redundancy in naming modules.

```python
openfl.component.straggler_policy.CutOffPolicy
# This is self-documenting, and easy to remember than

openfl.component.straggler_handling_functions.straggler_handling_function.CutoffTimeBasedStragglerHandling
```

If this also means merging the three policy modules into one file, consider doing so. The code complexity of any policy does not justify having separate modules.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -3,18 +3,19 @@
...
 from openfl.component.straggler_handling_functions.straggler_handling_function import (
-    StragglerHandlingFunction,
+    StragglerHandlingPolicy,
 )
 
 
-class CutoffTimeBasedStragglerHandling(StragglerHandlingFunction):
-    """Cutoff time based Straggler Handling function."""
-
+class CutoffTimeBasedStragglerHandling(StragglerHandlingPolicy):
```
MasterSkepticista, line 18: Missing docstring (explaining usage and behavior, not the arguments).

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -30,11 +31,101 @@ def __init__(
                 collaborators that should report. Defaults to 1.
             **kwargs: Variable length argument list.
         """
+        if minimum_reporting <= 0:
+            raise ValueError("minimum_reporting must be >0")
```
MasterSkepticista, line 37: Shouldn't the default behavior be to accept all reportings and remain synchronous, i.e. `inf`?

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -30,11 +31,101 @@ def __init__(
...
+
         self.round_start_time = round_start_time
         self.straggler_cutoff_time = straggler_cutoff_time
         self.minimum_reporting = minimum_reporting
+        self.logger = getLogger(__name__)
```
MasterSkepticista, line 42: I have seen this used in many places throughout OpenFL. 
Any good reason to create per-object loggers? Can we not use the method below?

```python
import logging
...

logging.info(...)
```

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -30,11 +31,101 @@ def __init__(
...
+                "is set to np.inf."
+            )
+
+    def reset_policy_for_round(self) -> None:
+        """
+        Reset timer for the next round.
+        """
+        if hasattr(self, "timer"):
+            self.timer.cancel()
+            delattr(self, "timer")
```
MasterSkepticista, line 56: I think this is a dangerous design choice to use `delattr` in classes. It is deleting attributes that other methods may call with no protection or knowledge thereof.

Why can't we use a simple `time.monotonic()` to store values in a variable and assign it to `0`, or `-1` to achieve the same goal?

[Here (internal)](https://github.com/intel-sandbox/orion-jax/blob/27fd37734b0cd98fbb56bf1fb20e6f94a9a3d071/orion/utils.py#L339) is an example stopwatch implementation for reference. Tweak as you like.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
```diff
@@ -30,11 +31,101 @@ def __init__(
...
+                    f"{self.minimum_reporting} collaborator(s) to report results."
+                )
+                return False
+
+        # Something has gone wrong, unhandled scenario, raising error.
+        raise ValueError(
+            "Unhandled scenario"
+            f"{self.__straggler_time_expired()=}"
+            f"{self.straggler_cutoff_time=}"
+        )
```
MasterSkepticista: This is an extremely unhelpful error message. At the very least, it should mention what was expected and what happened.

openfl/utilities/optimizers/numpy/adam_optimizer.py
MasterSkepticista, line 1: Is this a breaking change? It seems unrelated to the PR.

openfl/component/straggler_handling_functions/straggler_handling_function.py
```diff
@@ -7,15 +7,48 @@
 from abc import ABC, abstractmethod
 
 
-class StragglerHandlingFunction(ABC):
+class StragglerHandlingPolicy(ABC):
```
MasterSkepticista, line 10: Eliminate redundancy, for a user would already know what is being imported by the time they type the module path.

Suggested: `openfl.component.straggler_policy.BasePolicy`
(+ prev comment on merging all classes to single module as it is simple logic)

openfl/component/straggler_handling_functions/percentage_based_straggler_handling.py
MasterSkepticista, line 1: I think percentage based cutoff is using an incorrect logic.

Percentage-based cutoff should be a specialization of time-based cutoff. 

If all collaborators take 1 second to finish each task, give or take, the policy should not go ahead with a x% cutoff "just-because" some collaborators won the race to call RPC. In this case there are no stragglers per-se.

Ideally, this policy should give a time budget to each collaborator, and then implement a percentage cutoff if the time budget is violated at each participant. What do you think?

MasterSkepticista (changes requested):

openfl/component/aggregator/aggregator.py
ishant162: Accepted. Code Updated

openfl/component/aggregator/aggregator.py
ishant162: It is not only used by straggler policies but also to identify collaborators who have completed all their tasks in the 
`_compute_validation_related_task_metrics()` function.

openfl/component/aggregator/aggregator.py
ishant162: This change was introduced in seperate PR: https://github.com/securefederatedai/openfl/pull/1003. 
Unsure whether the changes to class attribute were intentional. Suggested to be discussed seperately as they do not seem to be related to straggler handling.

openfl/component/__init__.py
ishant162: Teo had alternate naming suggestions https://github.com/securefederatedai/openfl/pull/996#discussion_r1689369484.
Recommended that naming and other suggestions to be taken up in a seperate follow-up PR.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: Updated

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: The check ensures that the `minimum_reporting` value is greater than zero, as atleast one collaborator is required to end the round.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: There is no specific reason to follow this approach, as both options are used in OpenFL.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: Accepted. code updated

openfl/utilities/optimizers/numpy/adam_optimizer.py
ishant162: Looks like an intermediate merge. Modified as per main line

openfl/component/straggler_handling_functions/straggler_handling_function.py
ishant162: This can be handled along with https://github.com/securefederatedai/openfl/pull/996#discussion_r1751767679  recommended to be done as a seperate PR.

openfl/component/straggler_handling_functions/percentage_based_straggler_handling.py
ishant162: Our understanding is that Percentage-based policy is not a subset of time-based policy, it is rather provided as an alternative to time-based straggler handling policy. 
Due to this, suggestion of treating it as a specialization of cutoff policy cannot be considered
"Ideally, this policy should give a time budget to each collaborator, and then implement a percentage cutoff if the time budget is violated at each participant." - this seems quite similar to CutOffTimebased policy where Aggregator waits for a certain time for Collaborators to report before proceeding with early round end.

openfl/component/aggregator/aggregator.py
```diff
@@ -177,6 +178,13 @@ def __init__(
...
+        # Initialize a mutex lock for thread safety
+        self.mutex = Lock()
```
teoparvanov: nit - a mutex is a lower-level construct (used for the implementation of the Lock object under the hood). Therefore, I suggest to preserve the high-level abstraction here, and simply write:
```python
self.lock = Lock()
```

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
ishant162: Threading.timer is used to implement a cutoff timer that manages the CutOffTimeBasedPolicy. 
 
While trying to incorporate the feedback and suggestion to avoid deleting the timer attribute that was discussed yesterday, we realized that there is a need to delete the threading.timer object every round. Since the start method can be called only once per object (see [threading — Thread-based parallelism — Python 3.12.6 documentation](https://docs.python.org/3/library/threading.html#threading.Thread.start)), therefore there is a need to delete the old object at end of round and instantiate a new object every round
 
Further our understanding is that since the attribute is only used within the CutOffTimeBasedPolicy in start_policy() at the start of the round and reset_policy_for_round() at the end of round, therefore it is relatively safe to be used within the policy (if required for some more reason)

openfl/component/aggregator/aggregator.py
ishant162: Updated as per your suggestion. Thanks!

teoparvanov (approved):

openfl/component/aggregator/aggregator.py
ishant162: The code change was reverted due to failing Pytest and code coverage checks. Additional modifications are required and will be addressed in a separate follow-up PR

psfoley: @MasterSkepticista can this be marked as resolved?

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
psfoley: @ishant162 I would also recommend `time.monotonic()` to track the time elapsed - it will be far more readable for the next person who wants to touch / understand this code.

psfoley (changes requested):

openfl/component/aggregator/aggregator.py
psfoley: I spoke with @ishant162 about this offline. This will be resolved in a future PR.

openfl/component/straggler_handling_functions/cutoff_time_based_straggler_handling.py
psfoley: Discussed offline. `threading.Timer()` is necessary to trigger the check to determine whether the time has elapsed in the case that the final remaining collaborator in the middle of a long running task and others have stopped responding. Deletion of timer attribute is intended to free up memory (may not be necessary IMO). In 1.7 I see several opportunities for optimization here, but this code is well tested and addresses multiple issues in develop, so I'm marking this as resolved.

psfoley (approved):

PR #1001, opened 2024-07-10
Correct the url in the installation doc

yanmxa:
Signed-off-by: myan <myan@redhat.com>


MasterSkepticista:
Thanks @yanmxa!

Would appreciate if you could extend your fix to all files in this repository. I did a quick grep on the repository - same fix can be applied on all instances.

yanmxa:
> Thanks @yanmxa!
> 
> Would appreciate if you could extend your fix to all files in this repository. I did a quick grep on the repository - same fix can be applied on all instances.

Sure! Done.

MasterSkepticista (approved):

PR #717, opened 2023-01-30
[CI] Fix dockerization test

itrushkin:
## Issue
The output of CI dockerization jobs stops at the collaborator certification. Aggregator and collaborator containers do not start.

## Fix 
Change `ProcessPoolExecutor.submit` call signature.

PR #1028, opened 2024-08-13
Fix Dockerization CI test

MasterSkepticista:
Dockerization test is currently broken but does not appear as such on CI, as the script `tests/github/dockerization_test.py` does not check the exit code of participants.

This PR:
* Cleans up the base dockerfile: removes command pipes as flagged by coverity. Also all licensing downloads are taken care of by `./license.sh`, keeping the Dockerfile self-documenting, lean, and hackable by users for their own deployments.
* Migrates whole dockerization test to action file. Idea is to have it self-documenting that also informs users _how_ to use a dockerized workspace, as compared to the python test file.
* Improves debuggability: Each stage is easier to inspect, like in [this](https://github.com/securefederatedai/openfl/actions/runs/10372989592/job/28717021242?pr=1028) example.

This PR will close the long-standing #717.


.github/workflows/dockerization.yml
```diff
@@ -28,7 +23,75 @@ jobs:
...
+      run: |
+        cd example_workspace
+        fx workspace certify
+
+    - name: Create signed cert for collaborator
+      run: |
+        cd example_workspace
+        fx collaborator create -d 1 -n charlie --silent
+        fx collaborator generate-cert-request -n charlie --silent
+        fx collaborator certify --request-pkg col_charlie_to_agg_cert_request.zip --silent
```
mansishr, line 48: I would advocate for running 2 collaborator containers

openfl-docker/Dockerfile.base
```diff
@@ -1,122 +1,33 @@
-# Copyright (C) 2020-2023 Intel Corporation
+# Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
+# ------------------------------------
+# OpenFL Base Image
+# ------------------------------------
+FROM ubuntu:22.04 as base
 
-# If your machine is behind a proxy, make sure you set it up in ~/.docker/config.json
```
mansishr, line 4: IMO we need to keep this line about configuring `config.json` unless you've stated it somewhere else

openfl-docker/licenses.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Clean up APT lists
+rm -rf /var/lib/apt/lists/*
+
+# Update the package list and install specified packages
+apt-get update
+apt-get install -y --no-install-recommends --fix-missing \
+    openssh-server=* \
+    curl=* \
+    ca-certificates=*
```
mansishr: These installs need to go in dockerfile as they aren't related to licenses?

mansishr:
Thanks @MasterSkepticista for these changes! It should be ready to merge after the comments are addressed.

openfl-docker/Dockerfile.base
```diff
@@ -1,122 +1,33 @@
...
-
-COPY --from=python-dev /usr/lib/x86_64-linux-gnu /usr/lib/x86_64-linux-gnu
-COPY --from=python-dev /usr/lib/${PYTHON} /usr/lib/${PYTHON}
-COPY --from=python-dev /usr/local/lib/${PYTHON} /usr/local/lib/${PYTHON}
-COPY --from=python-dev /usr/bin /usr/bin
-COPY --from=python-dev /usr/local/bin /usr/local/bin
-
-FROM python-base as openfl
-
+USER root
```
kminhta: I think this will continue to trigger the Coverity security issue: _The Docker container is configured to run as the root user._ Is it possible to set this to run as a non-root user? Then we can mirror that in the gramine Dockerfile

openfl-docker/Dockerfile.base
MasterSkepticista: Done.

.github/workflows/dockerization.yml
MasterSkepticista: The (now-deprecated) python test was based on a single collaborator. I will limit the scope of this PR as fixing the CI test.
But this is a fair concern. I have added a `TODO` in the action file to be addressed in later PRs.

openfl-docker/licenses.sh
MasterSkepticista: Removed.

openfl-docker/Dockerfile.base
MasterSkepticista: Done.

PR #1030, opened 2024-08-21
Remove TensorDB utils

MasterSkepticista:
These are not used throughout the framework. `TensorDB.get_tensor_from_cache` is the recommended method to use.


psfoley:
These methods are used in the [FeTS Challenge code](https://github.com/FeTS-AI/Challenge/blob/main/Task_1/FeTS_Challenge.py#L411-L413y) for storing custom information at the aggregator.

PR #1054, opened 2024-09-20
added no-browser (headless) option to run jupyter notebooks on VM.

tanwarsh:
While running Jupyter Notebooks from a virtual machine FX start tutorial throws error
```
W 2024-09-20 06:06:05.536 ServerApp] No web browser found: Error('could not locate runnable browser').
```
In order to access notebooks we need to run it in headless mode.
Added no browser option to start the server without browser.

```
fx tutorial start --ip="0.0.0.0" --port=8080 
fx tutorial start --ip="0.0.0.0" --port=8080 --no-browser
```

Result:
```
[I 2024-09-20 05:59:07.189 ServerApp] Serving notebooks from local directory: /home/openfl/openfl-tutorials
[I 2024-09-20 05:59:07.189 ServerApp] Jupyter Server 2.14.2 is running at:
[I 2024-09-20 05:59:07.190 ServerApp] http://VM-IP:8080/lab?token=<token>
[I 2024-09-20 05:59:07.190 ServerApp]     http://127.0.0.1:8080/lab?token=<token>
```


openfl/interface/tutorial.py
```diff
@@ -42,12 +42,20 @@ def tutorial(context):
     type=IntRange(1, 65535),
     help="The port the Jupyter Lab server will listen on",
 )
-def start(ip, port):
+@option(
+    "-no-browser",
+    "--no-browser",
+    required=False,
+    type=bool,
+    help="The Jupyter Lab server will without opening a browser",
```
rahulga1: typo

rahulga1 (approved):

MasterSkepticista:
@psfoley I have a different opinion on this. `fx tutorial start [OPTIONS]` is a pure wrapper over `jupyter lab [OPTIONS]`. I think we should eliminate this command group entirely. I don't see a good reason for users to remember this command, given how trivial it is to install/run notebook servers. We can document this in our install guide. WDYT?

@tanwarsh If the decision is to keep this command, it would be better to simply forward all keyword arguments to `jupyter lab` subcommand that gets called. That way users know `fx tutorial start` is simply a pass-through.

tanwarsh:
@MasterSkepticista  I agree. `fx tutorial start [OPTIONS] ` is a pure wrapper and If we do not want to keep this then we can remove this command group or else we can forward all keyword arguments to the subcommand if we want to keep it.

MasterSkepticista:
@tanwarsh I suggest we proceed by deprecating `fx tutorial` command group in code (removing all instances from documentation, examples etc.). It can be removed completely in a subsequent PR. I will let @psfoley comment if we intend to keep it.

teoparvanov (approved):
If I may - I actually agree and support both of you, @MasterSkepticista and @tanwarsh 😊 
My suggestion would be for a two-step approach:

1. We merge this PR as an incremental improvement which is valuable in itself
2. A second (future) PR removes the `fx tutorial` command group, and updates all relevant documentation, to help users migrate to the `jupyter lab` native approach.

theakshaypant (approved):

PR #1070, opened 2024-09-27
enable installation of requirements through `fx plan initialize`

kminhta:
This PR: 
- enables installation/re-installation of packages listed under `requirements.txt`  when `fx plan initialize` is invoked. This is done through a newly added `--install_reqs` flag that accepts a Boolean input (Default: `True`). This is important for creating custom workspaces. `fx plan initialize` correctly captures new py scripts and updates to the `plan.yaml`; however it is also expected that a user will introduce new packages required by their custom experiment. Currently, a user would have to run an additional `pip install -r -requirements.txt` in order for new packages to be installed or else the initialization will fail due to missing packages.

_Test_

To test this:
- Create a new workspace using `fx workspace create --template torch_cnn_mnist --prefix my_workspace`
- Uninstall `torch` and `torchvision` from your env
- Run `fx plan initialize`

Expected behavior:
- `torch` and `torchvision` will be installed (as defined in `requirements.txt`) and initialization will proceed as normal


teoparvanov (approved):

bsenthilr (approved):
LGTM

PR #1076, opened 2024-10-02
Provide a standard way to get `data_loader` in util contexts

jkk-intel:
## Problem Statement

There isn't a way to easily instantiate the `TaskRunner` object because it requires a functioning `data_loader` to be present during initialization. This forces non-workload util commands such as `fx plan initialize` or `fx model save` to also have data to be present. In short, in order to invoke any util function in `task_runner` which has nothing to do with the data or training (but rather, operation around before/after taining), one is still burdened with providing the initial data.

The PR aims to provide a standard way for util commands to gain access to very useful `task_runner` functions without having full collaborator data present in the workspace context.

## Getting Task Runner in Util Context

`get_dataloader(plan, prefer_minimal=True)` has been added to be able to instantiate data loader without data.

```python
# no need to provide or collaborator or data info, if you already know `input_shape`
data_loader = get_dataloader(plan, prefer_minimal=True, input_shape=my_input_shape)
task_runner = plan.get_task_runner(data_loader=data_loader)

# do task runner util stuff (non-training related)
task_runner.my_util_function()
```

## Change Log

- Added `openfl/utilities/dataloading.py` to provide a way to get
data_loader from plan object. `get_dataloader` function will accept
the plan object and options like `prefer_minimal` and `input_shape`
to provide either full `DataLoader` with training capability or a light
`MockDataLoader` with input_shape to make sure we can make
task_runner instance without full data context.

- Updated `openfl/interface/plan.py` and replaced the data_loader
fetch logic to use the new `get_dataloader`

- Updated `openfl/interface/model.py` and replaced the data_loader
fetch logic to use the new `get_dataloader`. This will make the fx
command `fx model save` to be able to save the model without
any initial data being present in the workspace. Added a new param
`--input-shape` to be able to instantiate the dataloader and taskrunner
when common sources don't provide the info.

### DIFF
--- a/openfl/utilities/dataloading.py
+++ b/openfl/utilities/dataloading.py
@@ -1,0 +1,87 @@
+# Copyright 2020-2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+import os
+import zipfile
+from typing import Union
+
+from openfl.federated import Plan
+from openfl.federated.data.loader import DataLoader
+from openfl.utilities.mocks import MockDataLoader
+
+
+def get_dataloader(
+    plan: Plan,
+    prefer_minimal: bool = False,
+    input_shape: Union[list, dict] = None,
+    collaborator_index: int = 0,
+) -> DataLoader:
+    """Get dataloader instance from plan
+
+    NOTE: if `prefer_minimal` is False, cwd must be the workspace directory
+    because we need to construct dataloader from actual collaborator data path
+    with actual data present.
+
+    Args:
+        plan (Plan):
+            plan object linked with the dataloader
+        prefer_minimal (bool ?):
+            prefer to use MockDataLoader which can be used to more easily
+            instantiate task_runner without any initial data.
+            Default to `False`.
+        input_shape (list | dict ?):
+            input_shape denoted by list notation `[a,b,c, ...]` or in case
+            of multihead models, dict object with individual layer keys such
+            as `{"input_0": [a,b,...], "output_1": [x,y,z, ...]}`
+            Defaults to `None`.
+        collaborator_index (int ?):
+            which collaborator should be used for initializing dataloader
+            among collaborators specified in plan/data.yaml.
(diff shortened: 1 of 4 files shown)


openfl/utilities/dataloading.py
```diff
@@ -0,0 +1,45 @@
...
+       'input_shape' in plan.config['data_loader']:
+        input_shape = plan.config['data_loader']['input_shape']
+
+    if prefer_minimal and input_shape:
+        # input_shape known, we can use MockDataLoader
+        data_loader: DataLoader = MockDataLoader(input_shape)
+    else:
+        # input_shape not known, load from data of first collab
+        # note: this is very odd flow
+        # TODO: need to review this way of getting data_loader
```
bsenthilr: one edge case if when user provides data config and doesn't provide collaborator yaml since the PR doesn't mandate it can break this logic

bsenthilr: the rest of the updates looks good. Can you please review this edge case

jkk-intel: Updated and handled both `data.yaml` not present case and also actual data path given by the selected collaborator not existing in the current workspace.

openfl/utilities/mocks.py
```diff
@@ -13,3 +13,9 @@ def __init__(self, feature_shape):
 
     def get_feature_shape(self):
         return self.feature_shape
+
```
bsenthilr, line 16: not related to this PR, but will it be cleaner to make 
MockDataLoader(DataLoader) and default other overrides for typing it better?

kminhta (approved):
LGTM, thanks @jkk-intel This will help unify the `fx plan initialize` and `fx model save` experience

openfl/utilities/mocks.py
jkk-intel: We probably need more discussion how to move forward with `data_loader` in context of util operations without data present in the workspace.

Issue #1056, opened 2024-09-20
Local build of Docker image gives permissions issue in the pip install step


ishaileshpant:
Local build of Docker image throws permission denied / file node found issue on "/home/openfl/" file(s) limiting local docker build / test and local trivy scans

Possibly an issue with "COPY" command on src folder to docker build context for "openfl" user home folder
![Screenshot 2024-09-19 140429](https://github.com/user-attachments/assets/75a8a666-6abc-45f0-940e-71177d3a1a23)

theakshaypant:
Closed with #1078

PR #1078, opened 2024-10-03
Fixes issue #1056 - related to docker src copy permissions in build

ishaileshpant:
- Fix docker permissions issue on src copy command to default user/group

### DIFF
--- a/openfl-docker/Dockerfile.base
+++ b/openfl-docker/Dockerfile.base
@@ -28,7 +28,7 @@ USER openfl
 
 # Install OpenFL.
 WORKDIR /home/openfl
-COPY . .
+COPY --chown=openfl:default . .
 ENV PATH=/home/openfl/.local/bin:$PATH
 RUN pip install --no-cache-dir -U pip setuptools wheel && \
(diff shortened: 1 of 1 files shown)


teoparvanov (approved):

noopurintel:
@ishaileshpant - Could you also add `name` and `tag` to the generated image - any default value which is agreed upon? Both are `none` right now.

image

noopurintel (approved):

theakshaypant:
> @ishaileshpant - Could you also add `name` and `tag` to the generated image - any default value which is agreed upon? Both are `none` right now.
> 
> image

@noopurintel the name and tag for the image need to be provided in the docker build command. The [Makefile](https://github.com/securefederatedai/openfl/blob/develop/openfl-docker/Makefile) uses `openfl:latest` (the latest tag is inferred as none is provided in the build command) as the naming convention.

noopurintel:
@teoparvanov - can this be merged now?

teoparvanov:
> @teoparvanov - can this be merged now?

Hi @noopurintel , sure, looks good! Let me merge it for you.

PR #1094, opened 2024-10-22
Logging fix - No of rounds saved vs started

noopurintel:
In addition, enabling manual trigger of pytest coverage workflow.

**Before** - Round number `r+1` is shown as "Saved" and "Started" both.
round_no_issue

**After** - Existing round number `r` is "Saved" and `r+1` is "Started", which is expected.
round_no_fixed


tanwarsh (approved):

ishaileshpant (approved):
LGTM

teoparvanov (approved):

Issue #73, opened 2021-05-07
FX command - make it more independent


rstoki:
**Is your feature request related to a problem? Please describe.**
During the initialization of the federation (federated environment), the `fx` command is very useful. However, it performs some "additional tasks", that are typically not required (or, in the future may be problematic), and need to be 'rolled-back' manually. 

A list of non-necessary actions:
1) Do **not call** `pip install -r requirements.txt``inside  `fx workspace create`
   - the command `fx workspace create` creates workspace from a template, but it also calls `pip install -r requirements.txt` inside. 
For example, for template `tf_2dunet`, it installs TensorFlow 2.3.1, which is non-current version as of 05/2021 -- the current version is 2.4.1. So typically, there is a big chance it will roll-back already installed (and working) tensorflow version in the user's python environment to the previous version, which may not be working for him. Moreover, if the user want to change/modify his model and/or supply his own pretrained one.  

2) do **not check data folders** inside `fx plan initialize` 
   - the command `fx plan initialize` take into consideration also data paths set in the `<workspace>\plan\data.yaml` file. But since the `fx plan initialize` is called on aggregator, and the data folders for individual clients are on a completely different computers, it must not be assumed they can be accessible from the aggregator. 

**Describe the solution you'd like**
1) Do not call `pip ... ` from `fx` tool 
2) Do not check existence of **data paths** since the setup is performed on the Aggregator (which, by definition, do not have access to any data)

rahulga1:
With recent changes, both of the ask have been taken care of.

PR #1096, opened 2024-10-22
Refactor `dockerize` and `export`

MasterSkepticista:
This PR:

- Fuses base image build into `fx workspace dockerize`.
- Refactors `fx workspace export`. This picks accumulated best practices from OpenFL-Security, like applying reasonable assumptions and eliminating the need for user intervention.
- Removes dead code.
- Removes use of Docker Python API for build. This makes it easier for post-hoc build debugging.
- Improves readability.
- Uses generic username to avoid confusion with `openfl` repo name.

Command:
```shell
Usage: fx workspace dockerize [OPTIONS]

  Package current workspace as a Docker image.

Options:
  --save           Export the docker image as <workspace_name>.tar file.
  --rebuild        If set, rebuilds docker images with `--no-cache` option.
  --revision TEXT  Optional, version of OpenFL source code to build base image from. If unspecified, default value in
                   `Dockerfile.base` will be used, typically the latest stable release. Format:
                   <OPENFL_GIT_URL>@<COMMIT_ID/BRANCH>
  --help           Show this message and exit.
```

Example usage:
```shell
# To build base image + workspace image
$my_workspace> fx workspace dockerize

# To export workspace image as tarball
$my_workspace> fx workspace dockerize --save

# To specify custom OpenFL build target for base image
$my_workspace> fx workspace dockerize --revision https://github.com/MasterSkepticista/openfl.git@v1.6
```


openfl-docker/Dockerfile.base
```diff
@@ -13,28 +13,27 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
 # Install OpenFL.
-WORKDIR /home/openfl
-COPY --chown=openfl:default . .
-ENV PATH=/home/openfl/.local/bin:$PATH
-RUN pip install --no-cache-dir -U pip setuptools wheel && \
-    pip install --no-cache-dir -e .
-
-# Download thirdparty licenses.
-RUN INSTALL_SOURCES=yes /home/openfl/openfl-docker/licenses.sh
+RUN git clone --depth=1 --branch=develop https://github.com/securefederatedai/openfl.git && \
```
ishaileshpant: This might limit if someone wants to build/test on their fork / branch as it will always checkout upstream develop

Suggestion: better to take that as param to base file and supply upstream and develop branch as defaults

teoparvanov: AFAIU, in the original Dockerfile, the container image was built based on the current folder (supposedly the openfl repo root) via the `COPY --chown=openfl:default . .` instruction.

As @ishaileshpant points out, why do we need to change this approach?

openfl-docker/Dockerfile.base
```diff
@@ -13,28 +13,27 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
+    apt-get purge -y linux-libc-dev && \
     rm -rf /var/lib/apt/lists/*
-RUN apt-get purge -y linux-libc-dev
 
 # Create an unprivileged user.
 RUN groupadd -g 1001 default && \
-    useradd -m -u 1001 -g default openfl
-USER openfl
+    useradd -m -u 1001 -g default user
+USER user
```
teoparvanov, line 29: What's the advantage of switching from `openfl` to `user` ?

openfl-docker/Dockerfile.base
MasterSkepticista: Agree. This is editable in OpenFL-Security for the same reason you mention. But I don't see this option being widely used here. I initially tested an editable version, but then retracted for multiple reasons:

1. The added complexity of forwarding URLs (`fx -> build-arg -> docker build`) is not worth the benefit gained.
2. It is trivial to change the URL by simply editing the base Dockerfile.
3. Dockerfile (and the associated `fx` command code) becomes less readable and adds maintenance cost.
4. Users typically edit their workspace source code, and base images are intended to be stable (unless they edit OpenFL source code, which we should not require them to, for it must be stable and bug-free).

That said, if there is a high _user_ consensus on making URLs configurable via `fx` command flags, that can be addressed in a future PR.

openfl/interface/workspace.py
```diff
@@ -423,89 +340,132 @@ def _get_dir_hash(path):
...
+    """
+    Exports the OpenFL workspace (in current directory)
+    to an archive.
+
+    \b
+    The archive contains the following files/dirs copied as-is:
+        - `src`: All experiment source code.
+        - `plan`: The FL plan directory.
+        - `save`: Model initial weights.
+        - `requirements.txt`: Package list required for the experiment.
```
teoparvanov, line 358: AFAIK, at the moment when the model owner exports the workspace, he is in possession of all the public certs (both client and server side). Why couldn't we include them all here for convenience?

PS: as the certs weren't included even in the original export implementation, this comment could be addressed in a separate PR

openfl-docker/Dockerfile.base
ishaileshpant: May be not via fx command flags but atleast in the Dockerfile.base itself they can be exposed as build-args and them defaulted to upstream - will be beneficial to openfl devs and not just users - every code changes that has to be validated in container, in current state will mandate dockerfile change as well instead of supplying the fork/branch to validate in PR as build args - does that make sense?

teoparvanov (commented):
Hi @MasterSkepticista, I have a couple of questions from my first read-throgh:

openfl-docker/Dockerfile.base
theakshaypant: As a compromise between this implementation and the pre-existing one, perhaps we can retain the `--base-image` flag in `fx workspace dockerize` to use a pre-built base image so that the user can build from their local/remote branch if they require or proceed with the latest develop code if the param is not used. 
This leaves the user with the option to modify the Dockerfile.base as per their own volition.

openfl-docker/Dockerfile.base
MasterSkepticista: Eliminates confusion between the cloned repo `openfl` and `openfl` user which has a home directory named `/home/$USER`

openfl/interface/workspace.py
MasterSkepticista: Exporting a workspace happens before (or perhaps regardless of) possession of certificates. It is the equivalent of "building a container image of all the computation that has to be done". This image is common for all participants.

Certificates and datasets mounted as volumes, since they vary for each participant.

teoparvanov: Actually, given that they are public information, and known at the time `fx workspace export` is invoked, the entirety of the federation's PKI certificates (agg, col, CA cert chain, etc.) can be bundled together and shared among all participants. This is what the Certificate Server effectively does in OpenFL-Security, right?

Anyhow, this discussion goes beyond the scope of the PR - so feel free to ignore me for now haha 😄

openfl-docker/Dockerfile.base
teoparvanov: In this case I'd suggest preserving your suggested approach, but using a less generic name, such as `openfl-user`. WDYT?

ishaileshpant: or make it configurable - that way whoever run can get all under their home dir in the container as well and no permissions / user-mapping issues as well from host->container

openfl-docker/Dockerfile.base
teoparvanov: I agree with @MasterSkepticista's simplification proposal. However, for the sake of stability, I suggest to "peg" to a specific stable tag (such as `v1.6`), instead of the latest `develop` branch, which could regularly contain breaking changes.

When you think about it, a regular user of OpenFL would not likely want to build their FL workspace on the "bleeding edge" of the framework code. If they do (or decide to use a custom branch), they can always change the Dockerfile, as @MasterSkepticista pointed out.

openfl-docker/Dockerfile.base
ishaileshpant: defaulted to generic name like @teoparvanov suggested

openfl/interface/workspace.py
MasterSkepticista: This is a good idea. I'm in for this, especially if there is potential to cut down steps!

openfl-docker/Dockerfile.base
MasterSkepticista: > AFAIU, in the original Dockerfile, the container image was built based on the current folder (supposedly the openfl repo root) via the `COPY --chown=openfl:default . .` instruction.
> 
> As @ishaileshpant points out, why do we need to change this approach?

My initial approach was to retain copying the OpenFL source code. I retracted it for following reasons:
1. Ideally we'd want a user to be able to run `fx workspace dockerize` anywhere a workspace is. Building a base image with copied source requires specifying build context. This would require adding another flag in `dockerize` to point to the source code. In other words, intention was to keep `fx workspace dockerize` portable, regardless of where the workspace is located on the file system. What if we don't want to provide this flag? We can do this via `SITEPACKS`.
2. Now `SITEPACKS` value depends on what _kind_ of the build it is. Example: we fetch dockerfile using `SITEPACKS/openfl-docker/Dockerfile.X`. Now two ugly cases emerge:
    - If user had installed OpenFL via `pip install -e .`, then `SITEPACKS` points to the (intended) OpenFL root. This heavily bloats the docker build context with venv and workspace data. There are ways around it ofcourse, but again, not worth fighting that complexity.
    - If user had installed OpenFL via `pip install .`, then `SITEPACKS` is `.../lib/python/site-packages`. This will pass _all_ packages into build context of base image, and will fail because it does not contain OpenFL installation code (setup.py etc.).
3. User will anyways have access to internet because build requires installation of pip packages. They can simply point to a fork of their choice and avoid all other debug hassle.

openfl-docker/Dockerfile.base
MasterSkepticista: Docker allows specifying `--user` with `docker run`. I think that covers?

MasterSkepticista:
@theakshaypant @ishaileshpant @teoparvanov 
https://github.com/securefederatedai/openfl/pull/1096/commits/13cd2983a2aa31ad72f08c179727efd569bac2e4 addresses version pinning and configurability. Please review.

openfl-docker/Dockerfile.base
teoparvanov: Yep, sounds good.

openfl-docker/Dockerfile.base
ishaileshpant: For anyone working on core framework features/bugs that needs to be tested / validated in container env will have to change the Dockerfile to their fork/branch to test/validate and in the PR will have to revert that back to the upstream - whereas, imo, keeping it configurable at the Dockerfile.base level will make it more flexible with no such additional overhead of maintenance as defaults can still be develop on upstream

.github/workflows/dockerization.yml
```diff
@@ -24,16 +24,12 @@ jobs:
...
-      run: |
-        docker build -t openfl -f openfl-docker/Dockerfile.base .
-
     - name: Create workspace image
       run: |
         fx workspace create --prefix example_workspace --template keras_cnn_mnist
         cd example_workspace
         fx plan initialize -a localhost
-        fx workspace dockerize --base_image openfl
+        fx workspace dockerize --save --revision https://github.com/securefederatedai/openfl.git@develop
```
teoparvanov: As we discussed offline, this wouldn't test the dockerization of the topic branch undergoing CI, so a potential breakage could go unnoticed...

After a little research, I came across the following way to "introspect" the branch and the commit hash:
```yaml
jobs:
...
    steps:
      - name: Get current branch
        id: get_branch
        run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV

      - name: Get current commit
        id: get_commit
        run: echo "COMMIT_SHA=${GITHUB_SHA}" >> $GITHUB_ENV

      - name: Display branch and commit
        run: |
          echo "Current branch: ${{ env.BRANCH_NAME }}"
          echo "Current commit: ${{ env.COMMIT_SHA }}"
```

openfl-docker/Dockerfile.base
```diff
@@ -13,28 +14,26 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
+WORKDIR /home/user
+ENV PATH=/home/user/.local/bin:$PATH
 
 # Install OpenFL.
-WORKDIR /home/openfl
-COPY --chown=openfl:default . .
-ENV PATH=/home/openfl/.local/bin:$PATH
+ARG OPENFL_REVISION=https://github.com/securefederatedai/openfl.git@ad53a0fdbefde8bc18404b6ea4d48344956b96ee
```
teoparvanov: Can we be more explicit, and reference the tag instead, as in:
```bash
ARG OPENFL_REVISION=https://github.com/securefederatedai/openfl.git@v1.6
```

openfl-docker/Dockerfile.base
theakshaypant: Haven't used `--user` before personally, hence the comment; I hope having `WORKDIR /home/user` in Dockerfile would not affect the container, permissions or otherwise, if any other user is specified while running.

MasterSkepticista: To an extent this is inevitable. AFAIK, a `USER` defined in the container should match the `--user` that gets passed at runtime (+ same uid/gid) to avoid permission issues.

Since any guess on our part would be just as good (or bad), we should leave it up to the users to edit their Dockerfile and match their deployments.

.github/workflows/dockerization.yml
MasterSkepticista: Fair. 

Most PRs come from a different upstream (i.e. user forks). Is there a way to extract upstream?
Format: `https://upstream.git@<branch|commit_sha>`

openfl-docker/Dockerfile.base
MasterSkepticista: Fixed https://github.com/securefederatedai/openfl/pull/1096/commits/21edded270d3bc9769fb27f22198d144b3ff1b6d

.github/workflows/dockerization.yml
theakshaypant: As @teoparvanov  mentioned, I think the pre-existing `GITHUB_REF` env variable will be helpful as mentioned in GitHub [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables). 

GITHUB_REF | The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>, for pull requests it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1.
-- | --

teoparvanov: @MasterSkepticista, as suggested by Aayush, could you try with `GITHUB_HEAD_REF` for the branch, and `GITHUB_SHA` for the commit hash?

MasterSkepticista:
@theakshaypant @ishaileshpant @teoparvanov
WDYT?

openfl-docker/Dockerfile.workspace
```diff
@@ -8,13 +8,11 @@ FROM ${BASE_IMAGE}
...
 
-USER openfl
+USER user
 ARG WORKSPACE_NAME
 COPY ${WORKSPACE_NAME}.zip .
-
-WORKDIR /home/openfl
 RUN fx workspace import --archive ${WORKSPACE_NAME}.zip && \
-    pip install -r ${WORKSPACE_NAME}/requirements.txt
+    pip install --no-cache-dir -r ${WORKSPACE_NAME}/requirements.txt
```
theakshaypant, line 15: Is the pip install necessary given that it is [already done by import sub-command](https://github.com/securefederatedai/openfl/blob/develop/openfl/interface/workspace.py#L253-L263)?

theakshaypant (approved):
LGTM.

openfl-docker/Dockerfile.workspace
MasterSkepticista: Ideally, import should not install packages (#73). A separate install is explicit and allows users to supply additional flags (or even use different package managers).

Implicit pip install will be removed in a future PR.

teoparvanov (approved):
Awesome improvements, thank you @MasterSkepticista !

Issue #1087, opened 2024-10-14
tensor_db get_aggregated_tensor duplicates code, returns tuple in corner case


msheller:
**Describe the bug**
Calling get_aggregated_tensor on a tensor that has already been aggregated returns a tuple of (tensor, {}).

**Expected behavior**
I would expect that get_aggregated_tensor would call get_tensor_from_cache first, then aggregate if no tensor is found, thus using the same logic as the get_tensor_from_cache method (which doesn't return a tuple).

**Additional context**
Unclear why we use the method names "get_tensor_from_cache" and "cache_tensor" instead of "get_tensor" and "add_tensor". Legacy names, maybe?

tanwarsh:
@msheller, This is an intended behaviour as aggregator calling `get_aggregated_tensor` on a tensor that has already been aggregated should not occur in normal conditions. Is there a special use case which requires aggregation more then once?

In my opinion, instead of throwing an error that causes the experiment to stop abruptly, we should handle this scenario gracefully by stating the reason and marking the experiment as unsuccessful.

or we can still allow experiment to continue for which the changes are done in https://github.com/securefederatedai/openfl/pull/1121.
@rahulga1 @teoparvanov @psfoley @MasterSkepticista @theakshaypant what are your thoughts on this?

MasterSkepticista:
Instead of maintaining internal call state, and/or raising exceptions, the function should obey [idempotence](https://en.wikipedia.org/wiki/Idempotence) - returning the same result on every call.

Unless there are situations where calling any function more than once is violation of a state in a state machine (which I do not believe to be the case for calling `get_aggregated_tensor`), idempotence is the way to go.

tanwarsh:
> idempotence

Ya, changes in https://github.com/securefederatedai/openfl/pull/1121 will make it idempotent which is suggested in my previous comment.

teoparvanov:
Suggesting to close this issue, as it has been addressed in OpenFL [v1.6](https://github.com/securefederatedai/openfl/releases/tag/v1.6) as part of PR https://github.com/securefederatedai/openfl/pull/996/

PR #1121, opened 2024-11-05
Fix for tensor_db get_aggregated_tensor returns tuple in corner case 1087

tanwarsh:
Describe the bug
Calling get_aggregated_tensor on a tensor that has already been aggregated returns a tuple of (tensor, {}).

For more information please refer https://github.com/securefederatedai/openfl/issues/1087

Reproduced by  calling get_aggregated_tensor twice from aggregator while running experiment with 2 collaborators.

Before:
Screenshot 2024-11-05 at 4 54 49 PM

After:
Screenshot 2024-11-05 at 5 02 02 PM


rahulga1 (approved):

scngupta:
My **impression** was that this scenario: aggregator calling get_aggregated_tensor on a tensor that has already been aggregated should not occur in **normal conditions.** For e.g. this may happen when there is an undesired behavior due to which aggregator is performing the aggregation twice. In this case I found that the existing functionality of returning a tuple + subsequent exception actually helps catch undesired behavior in the system 

Can you please clarify if there are use cases where this is an intended behavior?

scngupta-dsp:
My **impression** was that this scenario: aggregator calling get_aggregated_tensor on a tensor that has already been aggregated should not occur in **normal conditions**. For e.g. this may happen when there is an undesired behavior due to which aggregator is performing the aggregation twice. In this case I found that the existing functionality of returning a tuple + subsequent exception actually helps catch undesired behavior in the system

Can you please clarify if there are use cases where this is an intended behavior ?

nisha987 (approved):

MasterSkepticista:
@tanwarsh Could you provide a MWE for this bug?

teoparvanov:
Closing this PR following https://github.com/securefederatedai/openfl/issues/1087#issuecomment-2478561800.

PR #1130, opened 2024-11-08
Flake8 linter fixes + updating PR lint rules

teoparvanov:
Here we update the `ubuntu.yml` workflow to use the same (OpenFL-specific) lint rules as the `lint.yml` workflow that runs as part of the PR pipeline.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -16,14 +16,12 @@ jobs:
       uses: actions/setup-python@v3
       with:
         python-version: "3.8"
-    - name: Install dependencies
+    - name: Install linters
       run: |
         python -m pip install --upgrade pip
         pip install -r linters-requirements.txt 
-        pip install .
-    - name: Lint with flake8
-      run: |
-        flake8 --show-source
+    - name: Lint with OpenFL-specific rules
+      run: bash shell/lint.sh
 
   pytest-coverage: # from pytest_coverage.yml
     needs: lint
--- a/.github/workflows/docker-bench-security.yml
+++ b/.github/workflows/docker-bench-security.yml
@@ -3,12 +3,14 @@ name: Docker Bench for Security
 on:
   pull_request:
     branches: [ develop ]
+    types: [opened, synchronize, reopened, ready_for_review]
 
 permissions:
   contents: read
 
 jobs:
   build:
+    if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     timeout-minutes: 10
 
--- a/.github/workflows/dockerization.yml
+++ b/.github/workflows/dockerization.yml
@@ -4,12 +4,14 @@ name: Dockerization
 on:
   pull_request:
     branches: [ develop ]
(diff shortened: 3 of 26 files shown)


MasterSkepticista:
@teoparvanov Those do not appear in Lint CI because there are exceptions to `flake8` provided via a config file in [shell/lint.sh](https://github.com/securefederatedai/openfl/blob/d4c97b3f67cf27ee56253a480265b96594ac2a1e/shell/lint.sh#L13), and it only covers source code, no examples/tutorials/templates.

I think we need to remove all `flake8` checks elsewhere. Any exceptions should be captured in `setup.cfg`.

teoparvanov:
> @teoparvanov Those do not appear in Lint CI because there are exceptions to `flake8` provided via a config file in [shell/lint.sh](https://github.com/securefederatedai/openfl/blob/d4c97b3f67cf27ee56253a480265b96594ac2a1e/shell/lint.sh#L13), and it only covers source code, no examples/tutorials/templates.
> 
> I think we need to remove all `flake8` checks elsewhere. Any exceptions should be captured in `setup.cfg`.

Thanks for pointing this out, @MasterSkepticista , I wasn't aware of the `setup.cfg` file. Let me take a look and I'll update the PR accordingly.

.github/workflows/ubuntu.yml
```diff
@@ -16,14 +16,12 @@ jobs:
...
+    - name: Install linters
       run: |
         python -m pip install --upgrade pip
         pip install -r linters-requirements.txt 
-        pip install .
-    - name: Lint with flake8
-      run: |
-        flake8 --show-source
+    - name: Lint with OpenFL-specific rules
+      run: bash shell/lint.sh
```
MasterSkepticista, line 24: @teoparvanov Not sure I understood why do we need a flake8 test in Ubuntu workflow?
I do see that it existed before albeit without correct args, but I don't think we need it here in the first place (it is already covered in lint test of every PR)

MasterSkepticista (commented):
JFYI: For files in `openfl-workspace`, this formatting was done to silence previous flake8 errors, but it won't make any difference in the future since `./shell/lint.sh` only covers `openfl/**` code.

You may choose to keep these changes or only keep `ubuntu.yml` and `lint.yml` changes.

.github/workflows/ubuntu.yml
teoparvanov: Hi @MasterSkepticista, in terms of integration, I think there is value in doing testing post-merge, just to confirm that the latest code is stable (in case multiple PRs interfered with each other functionally). I agree this is less relevant in the linting case, as any regressions would be caught at the PR level. Still, I'd prefer to keep the existing pattern, just aligning it with the rules defined for the PR workflow.

teoparvanov: PS: as you suggested, I'm dropping the workspace code fixes which are anyway not needed, given the settings in `setup.cfg`

MasterSkepticista (approved):

PR #1137, opened 2024-11-12
Set FQDN as env variable for pytest coverage

noopurintel:
**Problem**  - Pytest coverage is failing as part of Ubuntu and Windows nightly jobs. https://github.com/securefederatedai/openfl/actions/runs/11788591035/attempts/1

**Reason** - FQDN is missing in the environment section of above workflow yaml files.

After adding FQDN as localhost, it is working fine - https://github.com/payalcha/openfl/actions/runs/11792791416

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -26,6 +26,9 @@ jobs:
   pytest-coverage: # from pytest_coverage.yml
     needs: lint
     runs-on: ubuntu-latest
+    env:
+      # A workaround for long FQDN names provided by GitHub actions.
+      FQDN: "localhost"
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python 3.8
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -7,6 +7,10 @@ on:
 permissions:
   contents: read
 
+env:
+  # A workaround for long FQDN names provided by GitHub actions.
+  FQDN: "localhost"
+
 jobs:
   lint: # from lint.yml
     runs-on: ubuntu-latest
@@ -26,9 +30,6 @@ jobs:
   pytest-coverage: # from pytest_coverage.yml
     needs: lint
     runs-on: ubuntu-latest
-    env:
-      # A workaround for long FQDN names provided by GitHub actions.
-      FQDN: "localhost"
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python 3.8
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -7,12 +7,13 @@ on:
 permissions:
   contents: read
 
+env:
+  # A workaround for long FQDN names provided by GitHub actions.
+  FQDN: "localhost"
+
 jobs:
   pytest-coverage: # from pytest_coverage.yml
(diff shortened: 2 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -10,6 +10,9 @@ permissions:
 jobs:
   pytest-coverage: # from pytest_coverage.yml
     runs-on: windows-latest
+    env:
+      # A workaround for long FQDN names provided by GitHub actions.
+      FQDN: "localhost"
```
teoparvanov: I suggest making this a global env setting for all jobs in the `windows.yml` workflow.

.github/workflows/ubuntu.yml
```diff
@@ -26,6 +26,9 @@ jobs:
   pytest-coverage: # from pytest_coverage.yml
     needs: lint
     runs-on: ubuntu-latest
+    env:
+      # A workaround for long FQDN names provided by GitHub actions.
+      FQDN: "localhost"
```
teoparvanov: Same here, let's make it a global setting for `ubuntu.yml`.

rajithkrishnegowda:
ubuntu.yaml pipeline was failed because of 
FAILED tests/openfl/interface/test_aggregator_api.py::test_aggregator_generate_cert_request - ValueError: Attribute's length must be >= 1 and <= 64, but it was 65
======= 1 failed, 273 passed, 1 skipped, 2 warnings in 65.13s (0:01:05) ======== 

If we rerun pipeline(https://github.com/securefederatedai/openfl/actions/runs/11788591035) will be passed not sure if FQDN will help.

.github/workflows/windows.yml
noopurintel: Done.

.github/workflows/ubuntu.yml
noopurintel: Done.

teoparvanov:
> ubuntu.yaml pipeline was failed because of FAILED tests/openfl/interface/test_aggregator_api.py::test_aggregator_generate_cert_request - ValueError: Attribute's length must be >= 1 and <= 64, but it was 65 ======= 1 failed, 273 passed, 1 skipped, 2 warnings in 65.13s (0:01:05) ========
> 
> If we rerun it pipeline will be passed not sure if FQDN will help.

It will, as the error is due to the excessively long FQDNs of GHA runners. The suggestion here is to use localhost as the (aggregator) FQDN instead.

noopurintel:
@rajithkrishnegowda  - This error actually comes when we do not pass the FQDN and system picks the default domain name of given runner machine.
For Github runners - this is one example where I encountered the same error - `fv-az1986-223.m0gpccztuiae5imzg12qvibqrh.bx.internal.cloudapp.net` . It's clearly crossing the limit of 64 char.

Thus we have to explicitly provide the FQDN. It's already present in `pytest_coverage.yaml` file.

rajithkrishnegowda (approved):

teoparvanov (approved):
LGTM, thanks @noopurintel !

noopurintel:
Triggered jobs after making FQDN as global env:

1. Ubuntu - https://github.com/noopurintel/openfl/actions/runs/11793213643
2. Windows - https://github.com/noopurintel/openfl/actions/runs/11793214547

PR #1140, opened 2024-11-12
Gramine-SGX Container TEE support

MasterSkepticista:
## Overview
This PR adds **Gramine-based TEE support for Task Runner API** via the `fx workspace dockerize` command.
All images are built SGX-ready, with signing key automatically generated if a user does not explicitly provide one. This implies zero change in the command from a user's perspective, with a path forward to move to TDX/COCO.

Running enclaves does require attaching specific volumes and devices as Gramine requires. This is covered in the documentation.

Getting-started guide: [README.md](https://github.com/securefederatedai/openfl/blob/e4139ae0c2246b54ca297e7a690b20154fa8617b/openfl-docker/README.md)

### Usage (only new changes are shown):
```shell
Usage: fx workspace dockerize [OPTIONS]

  Package current workspace as a TEE-ready Docker image.

Options:
  ...
  --enclave-key TEXT  Path to an enclave signing key. If not provided, a key will be auto-generated in the workspace.
                      Note that this command builds a TEE-ready image, key is NOT packaged along with the image. You
                      have the flexibility to not run inside a TEE later.
  ...
```

### Tests
- Added Docker `gramine-direct` CI test (SGX is not supported on CI yet).
- Tested with `gramine-direct` and `gramine-sgx` on SGX-supported VMs.


openfl-docker/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+OpenFL can be used within a container for simulating Federated Learning experiments, or to deploy real-world experiments within Trusted Execution Environments (TEEs).
+
+## Base Image
+
+To develop or simulate experiments within a container, build the base image (or pull one from docker hub).
+
+```shell
+# Pull latest stable base image
+$> docker pull intel/openfl
```
teoparvanov, line 11: Is this image automatically uploaded on each OpenFL release?

openfl-docker/gramine_app/fx.manifest.template
```diff
@@ -0,0 +1,73 @@
...
+# List of allowed files that SGX enclave does NOT verify with signatures.
+# One should be conservative as to which files are allowed, these can be modified by enclave.
+sgx.allowed_files = [
+  "file:{{ workspace_root }}/save",
+  "file:{{ workspace_root }}/logs",
+  "file:{{ workspace_root }}/cert",
+  "file:{{ workspace_root }}/data",
+  "file:{{ workspace_root }}/plan/cols.yaml",
+  "file:{{ workspace_root }}/plan/data.yaml",
+  "file:{{ workspace_root }}/plan/plan.yaml",
```
teoparvanov, line 72: Given that these are direct configurations that directly determine the behavior of the enclave, shouldn't they be part of the trusted file list?

openfl-docker/gramine_app/fx.manifest.template
```diff
@@ -0,0 +1,73 @@
...
+# SGX configuration
+sgx.debug = false
+sgx.enclave_size = "16G"
+sgx.preheat_enclave = false
+sgx.remote_attestation = "dcap"
+sgx.max_threads = 512
+
+# List of trusted files, that are hashed and signed by the enclave.
+# If these files change after signing of an enclave, application cannot run.
+sgx.trusted_files = [
```
teoparvanov, line 47: I believe the openfl package (incl. `fx` commands) should be explicitly included here

teoparvanov (approved):
Great work, thank you @MasterSkepticista ! I really like the approach where the image is TEE-ready by default, but you can also run it in non-TEE environments. I just have a couple of minor comments/questions, but overall I believe this PR is in very good shape!

PS: with this, OpenFL is back in the TEE game 🚀

openfl-docker/README.md
MasterSkepticista: I think the scripts [here](https://github.com/securefederatedai/openfl/tree/develop/scripts) are used to publish latest wheels and images.
Following 1.7, OpenFL base image will come bundled with Gramine (and ability to build SGX-ready workspaces).

openfl-docker/gramine_app/fx.manifest.template
MasterSkepticista: Trusted list includes hashed read-only files that are known during the enclave measurement. Allowed files are those that cannot be known (sp. collaborator datasets, certificates etc.) during enclave build or could be modified during runtime (model checkpoints, logs).

openfl-docker/gramine_app/fx.manifest.template
MasterSkepticista: Current manifest includes `VENV_ROOT` which contains all python binaries (incl. `fx`) and packages' source code (this is also the approach taken internally).

openfl-docker/gramine_app/fx.manifest.template
teoparvanov: OK, makes sense, indeed.

openfl-docker/README.md
teoparvanov: Indeed, the `intel/openfl` image seems relatively up to date. The one I just pulled is marked as created "27 hours ago" - likely by some automated workflow. I guess I'm a little worried because those scripts are really old (3-4 years), and I don't understand the publishing workflow. But maybe you have better confidence in this process.

Just in case, have you tested the graminization workflow with this image, rather than the one built locally?

MasterSkepticista: > Just in case, have you tested the graminization workflow with this image, rather than the one built locally?

I'm not sure of the publishing workflow either. We should revisit this.

Dockerize with Gramine won't work with any older commits/tags/releases for graminized workspaces. Part of the process is copying manifest/makefiles from the source code (which means it must be at least the latest commit that this PR points to).

I agree that this is a little painful as we have to wait for next major release until users can benefit from this. Best would be if users always git clone the tip of develop.

kminhta (approved):
LGTM, nice work @MasterSkepticista !

PR #1141, opened 2024-11-12
[FeTS][GaNDLF] skip loading `train.csv` during inference mode

kminhta:
This PR is in support of updating [FeTS AI Challenge code base](https://github.com/FeTS-AI/Challenge/pull/201).

Specifically, this PR introduces a conditional check to handle inference mode in the data loader by adjusting the assignment of `self.train_csv`. If `data_path` contains a reference to `inference`, then `self.train_csv` will not be assigned a path. Otherwise, `self.train_csv` will load in `train.csv` from the specified `data_path`

This is important since training data is not present during inference/testing, so there needed to be a way to bypass this assignment


teoparvanov (approved):
LGTM!

PR #1145, opened 2024-11-13
Workflow-api based notebook tutorial for tensotflow.keras mnist based model with FedProx

yontyon:


rahulga1:
@kta-intel do we want to have modified FedProxOptimizer code part of tutorial?
Or should we modified it in the framework code itself. I know it will be reworked part of Keras 3 support in OpenFL.

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "14821d97",
+   "metadata": {},
+   "source": [
+    "# Workflow Interface Tutorial: Keras MINST"
```
MasterSkepticista: Rename: `Workflow API Tutorial: Keras MNIST`

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "f7f98600",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install tensorflow\n",
```
MasterSkepticista: New tensorflow does not support Keras APIs without a legacy flag. Suggest taking a look at [TensorFlow + Keras 2 backwards compatibility](https://keras.io/getting_started/) and modify as necessary.

My recommendation is to fully transition this tutorial to Keras 3 APIs instead of using deprecated Keras 2.

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+    "\n",
+    "    major_version = int(tf.__version__.split(\".\")[0])\n",
+    "    if major_version >= 2:\n",
+    "        from tensorflow.python.util import _pywrap_util_port\n",
+    "        print(\"Intel-optimizations (DNNL) enabled:\",\n",
+    "              _pywrap_util_port.IsMklEnabled())\n",
+    "    else:\n",
+    "        print(\"Intel-optimizations (DNNL) enabled:\")\n",
+    "\n",
+    "test_intel_tensorflow()"
```
MasterSkepticista: Not required. Unnecessary cognitive overhead + all tensorflow binaries are Intel-optimized by default.

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+    "inference(flflow.model, (test_images, test_labels), 32)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "a61a876d",
+   "metadata": {},
+   "source": [
+    "Now that the flow is complete, let's dig into some of the information captured along the way"
```
MasterSkepticista: Not sure the purpose of all cells after this comment. Suggest adding outputs, if relevant, or remove those cells altogether.

MasterSkepticista (changes requested):

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+    "# Workflow Interface Tutorial: Keras MINST"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "bd059520",
+   "metadata": {},
+   "source": [
+    "This tutorial demonstrates how to train a model using TensorFlow Keras on the MNIST dataset, leveraging the new FedAI Workflow Interface. The Workflow Interface provides a novel way to compose federated learning experiments with OpenFL, enabling researchers to handle non-IID data and perform federated averaging with optimizations like FedProx. Through this tutorial, you will learn how to set up the federated learning environment, define the flow, and execute the training process across multiple collaborators.\n"
```
teoparvanov: FedAI -> OpenFL

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
teoparvanov: @MasterSkepticista, I agree with your suggestion in general. However, in practice TaskRunner API currently supports up to Keras 2. Note that there are already preparations for a step-by-step upgrade to Keras 3. In the meantime, if we want the tutorial notebooks to be compatible with the upcoming `FederatedRuntime` (which uses TaskRunner API under the hood), we should stick with Keras 2 for now.

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+    "test_intel_tensorflow()"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "7237eac4",
+   "metadata": {},
+   "source": [
+    "The FedAvgWeights method is used in our federated learning architecture to aggregate model weights from multiple collaborators. In this approach, each collaborator trains a local model on its own data and then sends the model weights to a central aggregator. The aggregator runs the FedAvgWeights method to calculate the aggregated model weights by averaging the weights from all collaborators. "
```
teoparvanov: a couple of nits: 
* "its own data" -> "their local data"
* FedAvgWeights -> FedAvg

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+    "    \n",
+    "# # Setup participants\n",
+    "aggregator = Aggregator()\n",
+    "# aggregator.private_attributes = {}\n",
+    "collaborators = []\n",
+    "collaborator_names = ['Portland', 'Seattle', 'Chandler','Bangalore']\n",
+    "for idx, collaborator_name in enumerate(collaborator_names):\n",
+    "    collaborators.append(\n",
+    "        Collaborator(\n",
+    "            name=collaborator_name, num_cpus=0, num_gpus=0.3,\n",
```
teoparvanov: `num_cpus=1, num_gpus=0` ?

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import tensorflow as tf\n",
+    "import tensorflow.keras as keras\n",
+    "from tensorflow.python.ops import standard_ops\n",
+    "\n",
+    "\n",
+    "@keras.utils.register_keras_serializable()\n",
+    "class FedProxOptimizer(keras.optimizers.legacy.Optimizer):\n",
```
teoparvanov: Instead of implementing it inline, I suggest to update and use the `FedProxOptimizer` class from the `utilities.optimizers.keras` package.

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+      "_________________________________________________________________\n",
+      " 10/563 [..............................] - ETA: 6s - loss: 2.3115 - accuracy: 0.0938"
+     ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/home/yoni/github/openfl/.venv/lib/python3.8/site-packages/tensorflow/python/data/ops/structured_function.py:265: UserWarning: Even though the `tf.config.experimental_run_functions_eagerly` option is set, this option does not apply to tf.data functions. To force eager execution of tf.data functions, please use `tf.data.experimental.enable_debug_mode()`.\n",
+      "  warnings.warn(\n"
```
teoparvanov: Is it possible to silence or hide this warning? (especially if it mentions your name as in `/home/yoni/github/openfl` 😅 )

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
```diff
@@ -0,0 +1,1304 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "426f2395",
+   "metadata": {},
+   "source": [
+    "# Congratulations!\n",
+    "Now that you've completed your first workflow interface quickstart notebook, see some of the more advanced things you can do in our [other tutorials](broken_link), including:\n",
```
teoparvanov: the `other tutorials` link is broken

teoparvanov (approved):
LGTM, with a couple of comments and suggestions!

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
kminhta: An alternative option here would be to implement FedProx via a custom loss function instead of at the optimizer level. 

However, as far as I know, keras FedProx is not used in any "live" tutorials or workflows, so it would be safe to upgrade the package directly too. Only caveat is that we will eventually need to update it for Keras3, sooner or later

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
kminhta: If we decide to temporarily stick with Keras 2, we also will need to pin `tensorflow` to a compatible version or install `tf-keras` to use legacy code. 

However, since this tutorial is set to use `LocalRuntime` by default, I'm more inclined to say we should upgrade this and FedProx to Keras 3 API and then let the `Task Runner`/`FederatedRunTime` catch up. Which it should in a reasonable time. 

Upgrading FedProx to Keras3 is not something that would naturally happen by updating the Task Runner, so it would be worth the effort to include it in this PR if @yontyon as the bandwidth and/or is still interested, or otherwise an adjacent PR
 
Open to thoughts and suggestions

kminhta:
> @kta-intel do we want to have modified FedProxOptimizer code part of tutorial? Or should we modified it in the framework code itself. I know it will be reworked part of Keras 3 support in OpenFL.

In my opinion, I recommend not modifying FedProx inline for this tutorial. I think it would be better to update the framework code itself. I mentioned it in one of the review threads because I missed your message (sorry), but FedProx actually exists outside the taskrunner, so the efforts to upgrade the taskrunner to Keras3 would not capture FedProx. It would be more impactful to get that upgraded at the framework level

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: @kta-intel , I've updated the inline implementation to keras3 API. However, I don't think we should replace the original implementation before upgrading the whole project to keras3 (to avoid confusion and versions mixed in the code source). WDYT?
Please review the keras3 implementation as well

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: see my comment above.

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
teoparvanov: @yontyon, I agree with @kta-intel, so I'd suggest at least assessing how big would be the change of upgrading the [Keras FedProxOptimizer](https://github.com/securefederatedai/openfl/blob/develop/openfl/utilities/optimizers/keras/fedprox.py) at the framework level. If non-trivial, it can be done in a follow-up PR too.

openfl-tutorials/experimental/workflow/404_Keras_MNIST_with_FedProx.ipynb
```diff
@@ -0,0 +1,553 @@
...
+    "# What is it?"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "a7989e72",
+   "metadata": {},
+   "source": [
+    "The workflow interface is a new way of composing federated learning expermients with OpenFL. It was borne through conversations with researchers and existing users who had novel use cases that didn't quite fit the standard horizontal federated learning paradigm. "
```
eyalrund: It was **borne** through ==> born

yontyon: fixed

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: @kta-intel , @teoparvanov , I moved the new implementation to the openfl optimizers package. At the moment, the only usage of the Keras based fedprox implementation is in this tutorial so no additional changes are required,

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
yontyon: see above

openfl-tutorials/experimental/Workflow_FedProx_Keras_MNIST.ipynb
teoparvanov: Awesome, thanks @yontyon ! As pointed out by @kta-intel , this approach has a positive impact on the overall framework.

teoparvanov (approved):
This is great, thank you @yontyon !

openfl/utilities/optimizers/keras/fedprox.py
```diff
@@ -24,123 +23,51 @@ class FedProxOptimizer(keras.optimizers.Optimizer):
...
+        """Update step given gradient and the associated model variable."""
+        lr_t = tf.cast(learning_rate, variable.dtype)
+        mu_t = tf.cast(self.mu, variable.dtype)
+        gradient_t = tf.cast(gradient, variable.dtype)
+        vstar = self.vstars[self._get_variable_index(variable)]
+
+        self.assign_sub(variable, lr_t * (gradient_t + mu_t * (variable - vstar)))
 
     def get_config(self):
-        """Return the config of the optimizer.
```
kminhta, line 131: We should keep the docstring (and modify it if needed) for documentation purposes

openfl/utilities/optimizers/keras/fedprox.py
```diff
@@ -24,123 +23,51 @@ class FedProxOptimizer(keras.optimizers.Optimizer):
...
+        super().build(variables)
+        self.vstars = []
+        for variable in variables:
+            self.vstars.append(
+                self.add_variable_from_reference(
+                    reference_variable=variable, name="vstar"
+                )
+            )
+
+    def update_step(self, gradient, variable, learning_rate):
```
kminhta, line 56: Can you add a little more detail to this docstring (i.e. args, and "return"). Especially around proximal terms (mu and vstar) Good for documentation and will help if we need to update or debug in the future

kminhta (approved):
Very well done @yontyon ! The Keras 3 implementation is so much more elegant and readable. I just have some minor comments and a non-blocking question.

**Not a blocker, but a general question**: in your efforts, did you consider (or have any insights around) implementation of FedProx at the loss level? I know that OpenFL classically implemented FedProx and other aggregation algorithms at the optimizer level, but I am realizing that this reduces flexibility in training. If we incorporate a proximal term on the loss function itself, we enable the ability to use other optimizers (SGD, Adam, etc.) for updating gradients. There are at least a couple frameworks that implement it this way, which is why I bring it up. It probably makes no different for most applications, but I wonder if it's worth considering that route as we expand functionality to more complex models. Plus, it's simply easier to implement in most cases, which means it'll be easier to maintain.

In a future PR, of course :smile:

PR #1147, opened 2024-11-13
Update trivy version

gbikkiintel:


.github/workflows/trivy.yml
```diff
@@ -23,11 +23,13 @@ jobs:
...
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.24.0
+        uses: aquasecurity/trivy-action@0.28.0
         with:
           image-ref: 'docker.io/securefederatedai/openfl:${{ github.sha }}'
           format: 'sarif'
           output: 'trivy-results.sarif'
+        env:
+          TRIVY_DB_REPOSITORY: 'ghcr.io/aquasecurity/trivy-db,public.ecr.aws/aquasecurity/trivy-db'
```
ishaileshpant, line 32: IF the order matters in terms of failure, then I think keeping ecr ahead of ghcr would yield lesser failures due to higher limit

gbikkiintel: Agree, however, I put it ghcr.io/aquasecurity/trivy-db first so,  we get updates quickly after all public.ecr.aws/aquasecurity/trivy-db is a mirror and there could be some latencies!
Also, failover is super quick!

teoparvanov (approved):

ishaileshpant (approved):
LGTM

PR #1148, opened 2024-11-13
Task Runner tests: Adds non-TLS handling

noopurintel:
This PR aims to add non-TLS handling for Task runner tests.

In the existing workflow **Task Runner E2E**, I have added one more job to run with torch_cnn_mnist and python 3.10.
Existing job with TLS enabled will continue to run as it is.

Successful test run - https://github.com/noopurintel/openfl/actions/runs/11838637571


.github/workflows/task_runner_e2e_wo_mtls.yml
```diff
@@ -0,0 +1,93 @@
...
+  contents: read
+
+# Environment variables common for all the jobs
+env:
+  NUM_ROUNDS: ${{ inputs.num_rounds || '5' }}
+  NUM_COLLABORATORS: ${{ inputs.num_collaborators || '2' }}
+
+jobs:
+  test_run:
+    name: tr_non_tls
```
teoparvanov: There are actually several possible combinations: non-TLS, TLS (server auth only), and mTLS (client & server auth). As you cover those cases in future PRs, it could make sense to include those in the matrix, rather than creating a job per case.

tests/end_to_end/models/participants.py
```diff
@@ -132,13 +133,13 @@ def modify_plan(self, new_rounds=None, num_collaborators=None):
 
         data["aggregator"]["settings"]["rounds_to_train"] = int(self.rounds_to_train)
         data["data_loader"]["settings"]["collaborator_count"] = int(self.num_collaborators)
+        data["network"]["settings"]["disable_client_auth"] = True if disable_client_auth else False
```
teoparvanov: or simply `data["network"]["settings"]["disable_client_auth"] = disable_client_auth` ?

tests/end_to_end/models/participants.py
```diff
@@ -132,13 +133,13 @@ def modify_plan(self, new_rounds=None, num_collaborators=None):
...
+        data["network"]["settings"]["tls"] = False if disable_tls else True
```
teoparvanov: similarly, could we simply write `data["network"]["settings"]["tls"] = not disable_tls` instead ?

teoparvanov (approved):
LGTM!

.github/workflows/task_runner_e2e_wo_mtls.yml
noopurintel: I have kept the non-TLS as well in the same workflow now under a separate job.

For non-TLS, the matrix will have single combination of torch_cnn_mnist and python 3.10 for now. If required, this can be extended to other models and python versions in future.

tests/end_to_end/models/participants.py
noopurintel: Done.

tests/end_to_end/models/participants.py
noopurintel: Done.

Issue #1119, opened 2024-11-05
Some of important links are broken in GOVERNANCE.md


grib0ed0v:
_[Candidate for Good First Issue]_

In [GOVERNANCE.MD](https://github.com/securefederatedai/openfl/blob/develop/GOVERNANCE.md) some of important links are broken. 

For example: 
The current list of TSC members is [here](https://github.com/securefederatedai/openfl/TSC).

Easy to fix, since the file exists and you just need to correct the link: 
The current list of TSC members is [here](https://github.com/securefederatedai/openfl/blob/develop/TSC.md).

But, maybe, it's a good idea not just fix this, but also update other `.md` docs as well.

MasterSkepticista:
Thanks for pointing this out @grib0ed0v. We'll check other impacted files.

rajithkrishnegowda:
fixed in this [https://github.com/securefederatedai/openfl/pull/1175](PR)

rajithkrishnegowda:
fixed in PR  [https://github.com/securefederatedai/openfl/pull/1175]

PR #1150, opened 2024-11-15
Updating the main README.md

teoparvanov:
Some of the resources and terminology having become outdated, I propose a refresh of the OpenFL main README!

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,16 +35,16 @@ For more installation options check out the [online documentation](https://openf
 ## Getting Started
 
 
-OpenFL enables data scientists to set up a federated learning experiment following one of the workflows:
+OpenFL enables data scientists to set up a federated learning experiment following one of the interfaces:
 
-- [Aggregator-based Workflow](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html):
+- [Task Runner API](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html):
 Define an experiment and distribute it manually. All participants can verify model code and [FL plan](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html#federated-learning-plan-fl-plan-settings) prior to execution. The federation is terminated when the experiment is finished
 
-- [Workflow Interface](https://openfl.readthedocs.io/en/latest/about/features_index/workflowinterface.html) ([*experimental*](https://openfl.readthedocs.io/en/latest/developer_guide/experimental_features.html)):
+- [Workflow API](https://openfl.readthedocs.io/en/latest/about/features_index/workflowinterface.html) ([*experimental*](https://openfl.readthedocs.io/en/latest/developer_guide/experimental_features.html)):
 Create complex experiments that extend beyond traditional horizontal federated learning. See the [experimental tutorials](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/) to learn how to coordinate [aggregator validation after collaborator model training](https://github.com/securefederatedai/openfl/tree/develop/openfl-tutorials/experimental/102_Aggregator_Validation.ipynb), [perform global differentially private federated learning](https://github.com/psfoley/openfl/tree/experimental-workflow-interface/openfl-tutorials/experimental/Global_DP), measure the amount of private information embedded in a model after collaborator training with [privacy meter](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/Privacy_Meter/readme.md), or [add a watermark to a federated model](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/301_MNIST_Watermarking.ipynb).
 
 The quickest way to test OpenFL is to follow our [tutorials](https://github.com/securefederatedai/openfl/tree/develop/openfl-tutorials). </br>
-Read the [blog post](https://towardsdatascience.com/go-federated-with-openfl-8bc145a5ead1) explaining steps to train a model with OpenFL. </br>
+Read the [blog post](https://medium.com/openfl/from-centralized-machine-learning-to-federated-learning-with-openfl-b3e61da52432) explaining steps to train a model with OpenFL. </br>
 Check out the [online documentation](https://openfl.readthedocs.io/en/latest/index.html) to launch your first federation.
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,16 +35,16 @@ For more installation options check out the [online documentation](https://openf
 ## Getting Started
 
 
-OpenFL enables data scientists to set up a federated learning experiment following one of the workflows:
+OpenFL enables data scientists to set up a federated learning experiment following one of the interfaces:
```
MasterSkepticista: "OpenFL supports two APIs to set up a Federated Learning experiment:"?

We may want to reserve "interfaces" for something later :)

README.md
```diff
@@ -35,16 +35,16 @@ For more installation options check out the [online documentation](https://openf
...
 
-- [Aggregator-based Workflow](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html):
+- [Task Runner API](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html):
 Define an experiment and distribute it manually. All participants can verify model code and [FL plan](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html#federated-learning-plan-fl-plan-settings) prior to execution. The federation is terminated when the experiment is finished
```
MasterSkepticista: "This is a low-level API, meant for enterprise-ready real world FL experiments with support for TLS-based encryption, authentication and TEE-ready execution using SGX."

README.md
```diff
@@ -35,16 +35,16 @@ For more installation options check out the [online documentation](https://openf
...
+- [Workflow API](https://openfl.readthedocs.io/en/latest/about/features_index/workflowinterface.html) ([*experimental*](https://openfl.readthedocs.io/en/latest/developer_guide/experimental_features.html)):
 Create complex experiments that extend beyond traditional horizontal federated learning. See the [experimental tutorials](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/) to learn how to coordinate [aggregator validation after collaborator model training](https://github.com/securefederatedai/openfl/tree/develop/openfl-tutorials/experimental/102_Aggregator_Validation.ipynb), [perform global differentially private federated learning](https://github.com/psfoley/openfl/tree/experimental-workflow-interface/openfl-tutorials/experimental/Global_DP), measure the amount of private information embedded in a model after collaborator training with [privacy meter](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/Privacy_Meter/readme.md), or [add a watermark to a federated model](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/301_MNIST_Watermarking.ipynb).
```
MasterSkepticista: Addendum early in this line: "This is a high-level API, targeted mainly for simulating FL experiments at scale, using local compute infrastructure with low-boilerplate. Support for real-world, enterprise-ready FL experiments using this API is on the way."

README.md
```diff
@@ -53,7 +53,7 @@ Check out the [online documentation](https://openfl.readthedocs.io/en/latest/ind
 - Ubuntu Linux 18.04+
 - Python 3.7+ (recommended to use with [Virtualenv](https://virtualenv.pypa.io/en/latest/)).
 
-OpenFL supports training with TensorFlow 2+ or PyTorch 1.3+ which should be installed separately. User can extend the list of supported Deep Learning frameworks if needed.
+OpenFL supports training with TensorFlow 2+ or PyTorch 2+ which should be installed separately. Users can extend the list of supported Machine Learning frameworks if needed.
```
MasterSkepticista: "OpenFL supports popular NumPy-based ML frameworks like TensorFlow, PyTorch and Jax."

I don't think we should mention minimum versions. That is up to the user to choose.

README.md
```diff
@@ -53,7 +53,7 @@ Check out the [online documentation](https://openfl.readthedocs.io/en/latest/ind
 - Ubuntu Linux 18.04+
 - Python 3.7+ (recommended to use with [Virtualenv](https://virtualenv.pypa.io/en/latest/)).
```
MasterSkepticista: We should remove both these recommendations. Linux and Windows are both supported. As far as Python goes, 3.8 is our current minimum (which reminds me we should move to python3.10 **minimum** already - that is 2y old)

Recommendation for `virtualenv` also seems unnecessary now...

README.md
```diff
@@ -35,16 +35,16 @@ For more installation options check out the [online documentation](https://openf
...
 
 The quickest way to test OpenFL is to follow our [tutorials](https://github.com/securefederatedai/openfl/tree/develop/openfl-tutorials). </br>
-Read the [blog post](https://towardsdatascience.com/go-federated-with-openfl-8bc145a5ead1) explaining steps to train a model with OpenFL. </br>
+Read the [blog post](https://medium.com/openfl/from-centralized-machine-learning-to-federated-learning-with-openfl-b3e61da52432) explaining steps to train a model with OpenFL. </br>
 Check out the [online documentation](https://openfl.readthedocs.io/en/latest/index.html) to launch your first federation.
```
MasterSkepticista: This are too many. Can we just keep two (the new medium blog and a pointer to official docs)?

MasterSkepticista (approved):
Thanks for picking this up. I've added my suggestions as quotes so you can copy/paste/edit.

P.S: `Support` section has an outdated 2022 line. Please remove that too.

MasterSkepticista:
@teoparvanov Please also sneak this in if possible: #1119

README.md
```diff
@@ -53,7 +53,7 @@ Check out the [online documentation](https://openfl.readthedocs.io/en/latest/ind
 - Ubuntu Linux 18.04+
 - Python 3.7+ (recommended to use with [Virtualenv](https://virtualenv.pypa.io/en/latest/)).
 
-OpenFL supports training with TensorFlow 2+ or PyTorch 1.3+ which should be installed separately. User can extend the list of supported Deep Learning frameworks if needed.
+OpenFL supports training with TensorFlow 2+ or PyTorch 2+ which should be installed separately. Users can extend the list of supported Machine Learning frameworks if needed.
 
 ## Project Overview
 ### What is Federated Learning
```
kminhta, line 54: The `### Supported Aggregation Algorithms` table could be reworked. I think the aggregation algorithms are generally framework agnostic since the weights have been reduced to layer-based np arrays by the time it gets to this stage. Also, the docs all point to the same page. maybe we just keep it at `| algorithm name | paper |` ?

kminhta (approved):
Definitely time for a rework! Thanks a lot for taking this on @teoparvanov 
I'm sure this will be an ongoing effort, but this is an excellent start to freshen things up

README.md
teoparvanov: I agree in principle with your suggestion. I'll rephrase it a little bit, so it's true now, but also "future-proof". Let me know whether the updated phrasing makes sense.

teoparvanov:
> @teoparvanov Please also sneak this in if possible: #1119

I suggest to treat this separately as it requires general rework, based on a more strategic perspective. It's on my TODO list already 😉!

README.md
teoparvanov: Thank for pointing this out, @kta-intel. For now I propose a light-weight approach where I just remove the final column which is indeed redundant. In follow-up PRs we can also update the core content of this paragraph.

README.md
teoparvanov: I've added this, albeit a little rephrased.

PR #1151, opened 2024-11-15
Enable federated XGBoost using bootstrap aggregation in Task Runner

kminhta:
This PR enables a TaskRunner-based federated XGBoost using the bootstrap aggregation

Specifically this PR:
- creates an `xgb_higgs` task runner workspace to train on the higgs dataset [[ref](https://archive.ics.uci.edu/dataset/280/higgs)] with all required code (i.e. `src/taskrunner.py`, `src/dataloader.py `,`plan/*.yaml`, etc.
- adds a `tasks_xgb.yaml` to enable new `FedBaggingXGBoost` aggregation when running xgb training workloads
- adds `delta_updates`  parameter to Aggregator in order to bypass delta updating (for deep learning models getting weight deltas makes sense since the model size should stay relatively consistent, for tree-based algorithms, this makes less sense because more trees are added over time)
   - `delta_updates` is set to `true` by default to preserve normal behavior. xgboost taskrunner explicitly sets it to `false` to bypass it
- introduces new `loader_xgb.py` as the backend / superclass to `src/dataloader.py`
- introduces new `runner_xgb.py` as the backend / superclass to `src/taskrunner.py`
- introduces new federated boostrap algorithm for xgboost in `aggregation_function.fed_bagging` which bags the latest trees to a global model, consistent with currently accept federated xgboost algorithms in the industry


openfl-workspace/workspace/plan/defaults/tasks_xgb.yaml
MasterSkepticista, line 1: General question: Have you used specific formatters for `yaml` files?

openfl-workspace/xgb_higgs/.workspace
```diff
@@ -0,0 +1 @@
+current_plan_name: default
```
MasterSkepticista: Where is this used?

openfl-workspace/xgb_higgs/plan/cols.yaml
MasterSkepticista, line 1: It's a good idea to mention `DO NOT EDIT` for these files. As these are intended to be populated by `fx` commands.

openfl-workspace/xgb_higgs/src/dataloader.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def load_Higgs(data_path, **kwargs):
+    train_data = pd.read_csv(os.path.join(data_path, 'train.csv'), header=None)
+    X_train = train_data.iloc[:, 1:].values
+    y_train = train_data.iloc[:, 0].values
+
+    valid_data = pd.read_csv(os.path.join(data_path, 'valid.csv'), header=None)
+    X_valid = valid_data.iloc[:, 1:].values
+    y_valid = valid_data.iloc[:, 0].values
```
MasterSkepticista, line 58: Helps to add a docstring that briefs structure of the CSV.

openfl-workspace/xgb_higgs/src/setup_data.py
```diff
@@ -0,0 +1,94 @@
...
+
+logger = getLogger(__name__)
+
+"""HIGGS Dataset."""
+
+URL = "https://archive.ics.uci.edu/ml/machine-learning-databases/00280/HIGGS.csv.gz"
+FILENAME = "HIGGS.csv.gz"
+CSV_FILENAME = "HIGGS.csv"
+CSV_SHA384 = 'b8b82e11a78b81601381420878ad42ba557291f394a88dc5293e4077c8363c87429639b120e299a2a9939c1f943b6a63'
+DEFAULT_PATH = path.join(path.expanduser('~'), '.openfl', 'data')
```
MasterSkepticista: I suggest creating within same directory, instead of a hidden folder. This folder may not be guaranteed to mount in an enclave (one less thing to worry about if a user ever wants to do that)

openfl-workspace/xgb_higgs/src/taskrunner.py
```diff
@@ -0,0 +1,56 @@
...
+from sklearn.metrics import accuracy_score
+
+
+class XGBoostRunner(XGBoostTaskRunner):
+    """
+    Simple CNN for classification.
+
+    PyTorchTaskRunner inherits from nn.module, so you can define your model
+    in the same way that you would for PyTorch
+    """
```
MasterSkepticista: :)

openfl/component/aggregator/aggregator.py
```diff
@@ -186,6 +187,8 @@ def __init__(
         # Initialize a lock for thread safety
         self.lock = Lock()
 
+        self.delta_updates = delta_updates
+
```
MasterSkepticista: Boolean state variables (and function arguments) should be named as `self.use_delta_updates` - improves readability.

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
+# Copyright 2020-2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""XGBoostTaskRunner module."""
+
+# from copy import deepcopy
+# from typing import Iterator, Tuple
```
MasterSkepticista: Can be removed

openfl/federated/task/runner_xgb.py
MasterSkepticista, line 1: Many LOC seem to duplicate from other runners. Is there a possibility to inherit or abstract better?

openfl-workspace/xgb_higgs/plan/defaults
```diff
@@ -0,0 +1 @@
+../../workspace/plan/defaults
```
MasterSkepticista: I was under the impression that all plans inherit from OpenFL defaults without needing to mention the path. Is this required?

openfl-workspace/xgb_higgs/plan/data.yaml
MasterSkepticista, line 1: `DO NOT EDIT` comment

MasterSkepticista (changes requested):

openfl-workspace/xgb_higgs/src/setup_data.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Split the dataset into features and labels
+    X = higgs_data.iloc[:, 1:].values
+    y = higgs_data.iloc[:, 0].values
+
+    # Split the dataset into training and testing sets
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)
+
+    # Combine X and y for train and test sets
+    train_data = pd.DataFrame(data=np.column_stack((y_train, X_train)))
```
teoparvanov, line 78: It seems a bit counter-intuitive to put the labels (`y_train`) in the first column. Is this the expected layout for XGBoost training and validation?

openfl-workspace/xgb_higgs/src/setup_data.py
```diff
@@ -0,0 +1,94 @@
...
+    for i in range(collaborators):
+        dst = f'data/{i+1}'
+        makedirs(dst, exist_ok=True)
+
+        # Split the training data for the current collaborator
+        split_train_data = train_data.iloc[i::collaborators]
+        split_train_data.to_csv(path.join(dst, 'train.csv'), index=False, header=False)
+
+        # Copy the test data for the current collaborator
+        test_data.to_csv(path.join(dst, 'valid.csv'), index=False, header=False)
```
teoparvanov: So, if I understand correctly, the test data is the same for all collaborators?

openfl/federated/data/loader_xgb.py
```diff
@@ -0,0 +1,168 @@
+from math import ceil
+
+import numpy as np
+import xgboost as xgb
+
+
+class XGBoostDataLoader:
```
teoparvanov: Is there a reason not to inherit from the abstract `DataLoader`? You seem to follow this interface anyway.

openfl/federated/data/loader_xgb.py
```diff
@@ -0,0 +1,168 @@
...
+            np.random.seed(self.random_seed)
+
+        idxs = np.random.permutation(np.arange(X.shape[0]))
+
+        # compute the number of batches
+        if num_batches is None:
+            num_batches = ceil(X.shape[0] / batch_size)
+
+        # build the generator and return it
+        return self._batch_generator(X, y, idxs, batch_size, num_batches)
```
teoparvanov: It would improve readability if you moved the private utility methods at the bottom of the module.

PS: not a blocker for this PR, but as also pointed out by @MasterSkepticista, it's unfortunate that those utilities have to be copy/pasted across all data loader base classes for the supported ML frameworks. If there's already a utility module that those classes can import, such methods should go there. Otherwise we should create one. WDYT?

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
...
+        self.global_model = None
+        self.required_tensorkeys_for_function = {}
+        self.training_round_completed = False
+
+    def rebuild_model(self, input_tensor_dict):
+        """
+        Rebuilds the model using the provided input tensor dictionary.
+
+        This method checks if the 'local_tree' key in the input tensor dictionary is either a non-empty numpy array
+        or a non-None value. If this condition is met, it updates the internal tensor dictionary with the provided input.
```
teoparvanov: Just curious, what type could be this "non-None" value, other than a numpy array?

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
...
+        # optimizer state tensor names to request from the aggregator because
+        # these are only created after training occurs. A work around could
+        # involve doing a single epoch of training on random data to get the
+        # optimizer names, and then throwing away the model.
+        if self.opt_treatment == "CONTINUE_GLOBAL":
+            self.initialize_tensorkeys_for_functions()
+
+        # This will signal that the optimizer values are now present,
+        # and can be loaded when the model is rebuilt
+        self.training_round_completed = True
```
teoparvanov: Given that XGBoost doesn't have discrete optimizer objects, could we do away with this code, and the verbose TODO?

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def initialize_tensorkeys_for_functions(self, with_opt_vars=False):
+        """Set the required tensors for all publicly accessible task methods.
+
+        By default, this is just all of the layers and optimizer of the model.
+        Custom tensors should be added to this function.
+
+        Args:
+            with_opt_vars (bool): Flag to check if optimizer variables are
+                included. Defaults to False.
```
teoparvanov: This is N/A for XGBoost, right?

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
...
+            round_num (int): What round is it.
+            input_tensor_dict (dict): Required input tensors (for model).
+            **kwargs: Additional parameters.
+
+        Returns:
+            global_output_dict (dict):  Tensors to send back to the aggregator.
+            local_output_dict (dict):   Tensors to maintain in the local
+                TensorDB.
+        """
+        loader = self.data_loader.get_valid_dmatrix()
```
teoparvanov: I suggest to rename `loader` to simply `data`

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
...
+        if self.bst is None:
+            self.rebuild_model(input_tensor_dict)
+
+        # if self.bst is still None after rebuilding, then there was no initial global model, so set metric to 0
+        if self.bst is None:
+            # for first round agg validation, there is no model so set metric to 0
+            # TODO: this is not robust, especially if using a loss metric
+            metric = Metric(name="accuracy", value=np.array(0))
+        else:
+            metric = self.validate_(loader)
```
teoparvanov: ... which would make this `metric = self.validate_(data)`

openfl/federated/task/runner_xgb.py
```diff
@@ -0,0 +1,382 @@
...
+            input_tensor_dict (dict): Required input tensors (for model).
+            **kwargs: Additional parameters.
+
+        Returns:
+            global_output_dict (dict):  Tensors to send back to the aggregator.
+            local_output_dict (dict):   Tensors to maintain in the local
+                TensorDB.
+        """
+        self.rebuild_model(input_tensor_dict)
+        loader = self.data_loader.get_train_dmatrix()
```
teoparvanov: Same as above, suggesting to rename `loader` to `data`, so that it reads more naturally.

openfl/interface/aggregation_functions/fed_bagging.py
```diff
@@ -0,0 +1,130 @@
...
+
+            for local_tensor in local_tensors:
+                local_tree_bytearray = bytearray(local_tensor.tensor.astype(np.uint8).tobytes())
+                local_trees = json.loads(local_tree_bytearray)
+                global_model = append_trees(global_model, local_trees)
+
+        global_model_json = json.dumps(global_model)
+        global_model_bytes = global_model_json.encode("utf-8")
+
+        return np.frombuffer(global_model_bytes, dtype=np.uint8).astype(np.float32)
```
teoparvanov: I have to admit I'm a bit lost here... What is of type `np.uint8`, and what is of type `np.float32`? Don't we risk loss of precision in the transformation?

teoparvanov (commented):
Awesome work, thanks @kta-intel! I have a couple of questions and comments, but overall the PR looks in excellent shape for such a sizeable new feature.

PS: is there an easy way to add a CI job that covers at least the "happy path" of an XGBoost-based federation?

openfl/interface/aggregation_functions/fed_bagging.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+    for new_tree in range(num_local_trees):
+        local_trees[new_tree]["id"] = num_global_trees + new_tree
+        global_model["learner"]["gradient_booster"]["model"]["trees"].append(local_trees[new_tree])
+        global_model["learner"]["gradient_booster"]["model"]["tree_info"].append(0)
+
+    return global_model
+
+
+class FedBaggingXGBoost(AggregationFunction):
```
teoparvanov, line 62: When someone contributes an aggregation function, are there any facilities for (unit) testing it? Or are such functions only testable via at least a local E2E run of an FL workspace?

PS: I'm asking because AFAIK we also expect a couple of such contributions from various research organizations down the road, so we should probably set some expectations in this regard.

openfl/federated/task/runner_xgb.py
kminhta: Right, good catch. Fixed

openfl/federated/task/runner_xgb.py
kminhta: Yes, you're right. I've removed this bulky portion of code

openfl/federated/data/loader_xgb.py
kminhta: As an alternative, I've removed the batching and dataloader methods since they're not relevant to XGBoost

I do think it would be good to include a utility module for methods that are reused on multiple places. I won't include it in this PR, but it should be added to the backlog since it could clean up a lot of the code in other loaders/runners

openfl/federated/data/loader_xgb.py
kminhta: Oh, I must've accidentally deleted it when changing the class name. Thanks

openfl-workspace/xgb_higgs/src/setup_data.py
kminhta: That is a mistake on my end! Good catch. It should also split the test data for the collaborators

openfl-workspace/xgb_higgs/src/setup_data.py
kminhta: it's the way the higgs dataset is formatted, so I just kept it when splitting when data too (for consistency)

teoparvanov (approved):

openfl/federated/data/loader_xgb.py
teoparvanov: Sounds good, thanks!

openfl/interface/aggregation_functions/fed_bagging.py
kminhta: There is some inefficiency happening here that can probably be optimized. Before getting sent to `fed_bagging.py`, each collaborator trains their own decision tree. XGBoost models are accessed with `.save_raw()`, which returns a bytearray (is a sequence of bytes (`uint8`)). Almost all of OpenFL's pipelines will cast any data getting sent across the federation to `np.float32` before converting to bytes (ex. [pipeline.py#L72](https://github.com/securefederatedai/openfl/blob/develop/openfl/pipelines/pipeline.py#L72)). In theory, the XGBoost model bytearray can just be converted to bytes and sent directly, but this would require changes in the pipeline to relax their conditionals or else it'll through an error since it's expecting `np.float32`. In addition, models are stored as np arrays in the TensorDB and the aggregated model/values get converted to an np array (see: [tensor_db.py#L247](https://github.com/securefederatedai/openfl/blob/develop/openfl/databases/tensor_db.py#L247)). So a lot of these initial steps were in an effort to preserve the primary infrastructure (tensordb and pipelines). To answer your question though, the bytearray cannot be directly converted to `np.float32` and therefore is converted to np.uint8 before being casted as `np.float32`. There is a risk of precision loss, but this would happen during the pipeline anyway, so doing it in the runner allows us to check for precision loss and raise a warning (I added a `check_precision_loss` function to do before sending across the network).

In the future, I think we should aim to relax and/or otherwise address the conversion of data throughout the entire OpenFL work flow. I'll admit I need to dive a bit deeper into this myself since XGBoost was the first instance of having to deal with these types of conversions.

openfl/federated/task/runner_xgb.py
kminhta: This is vestigial. I removed it. Originally, I was trying to send metadata for the fedbagging algorithm as a dictionary alongside the model itself (within the same key), but that ended up not working out

kminhta:
> PS: is there an easy way to add a CI job that covers at least the "happy path" of an XGBoost-based federation?

Thanks for the review Teo! I'm not sure what you mean by this, would this just be a toy sanity check CI?

openfl/interface/aggregation_functions/fed_bagging.py
kminhta: This has been a question I've been musing on in the back of my mind for a while. I don't think there are unit tests in place and seems only testable via local E2E run, but I am wondering if there is a basic test a user can run with some dummy tensors to make sure that there are no bugs in the logic, since they presumably would know what the result would be (or have their own working implementation from their initial development). Open to suggestions though

openfl-workspace/xgb_higgs/.workspace
kminhta: thanks, this has been removed. copied over from a template workspace

openfl-workspace/xgb_higgs/plan/defaults
kminhta: I just tested it, your impression seems accurate. I removed this. It was included in other workspaces so I pulled it in by default

openfl/federated/task/runner_xgb.py
kminhta: Yeah, I've noticed that while working with the runners. There probably is a more elegant way to inherit / abstract methods to save lines of code. For now though, the base taskrunner is very skeletal with most methods not implemented. I think for now, this is the way the methods are implemented in the xgboost runner is the best we'll have it. We should put in the backlog a way to clean up the taskrunner so that each child class only write necessary framework specific logic and minimize duplication

openfl-workspace/xgb_higgs/src/dataloader.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    train_data = pd.read_csv(os.path.join(data_path, 'train.csv'), header=None)
+    X_train = train_data.iloc[:, 1:].values
+    y_train = train_data.iloc[:, 0].values
+
+    valid_data = pd.read_csv(os.path.join(data_path, 'valid.csv'), header=None)
+    X_valid = valid_data.iloc[:, 1:].values
+    y_valid = valid_data.iloc[:, 0].values
+
+    return X_train, y_train, X_valid, y_valid
```
noopurintel, line 60: Any specific reason for keeping an uppercase **X** in `X_train` and `X_valid` ? Same goes for load_**H**iggs. 
Wouldn't all lowercase function and variable names be more uniform?

teoparvanov:
> > PS: is there an easy way to add a CI job that covers at least the "happy path" of an XGBoost-based federation?
> 
> Thanks for the review Teo! I'm not sure what you mean by this, would this just be a toy sanity check CI?

Yes, I mean a very basic CI job that does an E2E run of the XGBoost workspace. Like [here](https://github.com/securefederatedai/openfl/blob/develop/.github/workflows/taskrunner.yml#L44), but using `xgb_higgs` as the template, and with a reduced test matrix (f.e. just ubuntu + python3.10). However, if the effort for this turns out to be significant, you can also consider doing it in a separate PR.

Other than that, the PR is ready to be merged IMO. Thanks, @kta-intel !

openfl/federated/task/runner_xgb.py
teoparvanov: I've added this to the backlog (we could give it a try for 1.8).

openfl/interface/aggregation_functions/fed_bagging.py
teoparvanov: I've already listed this in the backlog (post 1.7, most likely). This could include some some sort of contributor docs that outlines the process - APIs, test facilities, examples, etc. Speaking of APIs, it seems we only support the `AggregationFunction` abstraction for TaskRunner API. We should probably introduce something similar for Workflow API too. 

Some of the newly joining team members are already refactoring the FedProx-related examples to Workflow API, so they may take this up at some point in the future. I would suggest starting with a design document for a holistic approach. Your support in this regard will certainly be needed as you know best the practical circumstances and needs of external contributors of aggregation algorithms.

openfl/interface/aggregation_functions/fed_bagging.py
teoparvanov: Thanks for the detailed explanation, @kta-intel , this helps!

openfl-workspace/workspace/plan/defaults/tasks_xgb.yaml
kminhta: I copied over the `yaml` from other workspaces as a template then ran `bash shell/format.sh` in the whole repo. Is there something additional that you recommend?

openfl-workspace/xgb_higgs/src/dataloader.py
kminhta: True, I think it would be. It is mainly this way because of source convention. For `X_train` and `X_valid`, our other dataloaders used this convention, so I kept it here for consistently. For `Higgs` it is because the dataset refers to the Higgs boson (though I notice now that the dataset itself refers to HIGGS in all uppercase, so it might just be stylistic)

kminhta:
> Yes, I mean a very basic CI job that does an E2E run of the XGBoost workspace. Like [here](https://github.com/securefederatedai/openfl/blob/develop/.github/workflows/taskrunner.yml#L44), but using the `xgb_higgs` as a template, and with a reduced test matrix (f.e. just ubuntu + python3.10). However, if the effort for this turns out to be significant, you can also consider doing it in a separate PR.

I see! This is a good idea, but lets save it as a separate PR. Thanks for the review and suggestion(s) @teoparvanov!

openfl-workspace/xgb_higgs/src/dataloader.py
psfoley: @noopurintel `X_train` / `X_valid` is a common convention from other ML frameworks (see scikit-learn for an example): https://scikit-learn.org/dev/modules/generated/sklearn.model_selection.train_test_split.html#train-test-split

PR #1155, opened 2024-11-18
Memory leak Details 

payalcha:


openfl/component/aggregator/aggregator.py
```diff
@@ -966,6 +1008,15 @@ def _end_of_round_check(self):
         for task_name in all_tasks:
             self._compute_validation_related_task_metrics(task_name)
 
+        self.logger.info("*******************END OF ROUND CHECK: AGGREGATOR LOGS*******************************")
+        process = psutil.Process()
+        process_mem = round(process.memory_info().rss / (1024 ** 2),2)
+        self.logger.info(f"FEDAIQE Round: {self.round_number}")
```
rahulga1: Would suggest not to use FEDAI in the logs.

tests/end_to_end/models/participants.py
```diff
@@ -132,6 +132,8 @@ def modify_plan(self, new_rounds=None, num_collaborators=None, disable_client_au
             data = yaml.load(fp, Loader=yaml.FullLoader)
 
         data["aggregator"]["settings"]["rounds_to_train"] = int(self.rounds_to_train)
+        data["aggregator"]["settings"]["memleak_check"] = True
```
rahulga1: in my opinion, we should enable this by default.

openfl/component/collaborator/collaborator.py
```diff
@@ -588,3 +608,41 @@ def named_tensor_to_nparray(self, named_tensor):
         self.tensor_db.cache_tensor({decompressed_tensor_key: decompressed_nparray})
 
         return decompressed_nparray
+
+    def get_memory_usage(self, round_number, metric_origin):
```
rahulga1, line 613: It is good to have the overall memory usage, but if we can target the processes/containers specific to openfl, it will give more clear picture.

payalcha: This is the limitation currently. We will discuss this.

openfl-workspace/torch_cnn_mnist/plan/plan.yaml
```diff
@@ -9,6 +9,7 @@ aggregator :
     best_state_path     : save/torch_cnn_mnist_best.pbuf
     last_state_path     : save/torch_cnn_mnist_last.pbuf
     rounds_to_train     : 10
+    memleak_check : true
```
teoparvanov: I suggest calling the flag `log_memory_usage` instead.

openfl/component/collaborator/collaborator.py
```diff
@@ -171,6 +174,23 @@ def run(self):
 
                 # Cleaning tensor db
                 self.tensor_db.clean_up(self.db_store_rounds)
+                if self.memleak_check:
+                    # This is the place to check the memory usage of the collaborator
+                    self.logger.info("*****************COLLABORATOR LOGS*******************************")
+                    process = psutil.Process()
+                    process_mem = round(process.memory_info().rss / (1024 ** 2),2)
+                    self.logger.info("Collaborator Round: %s", round_number)
+                    self.logger.info("Collaborator Process Mem: %s", process_mem)
```
teoparvanov, line 184: you actually already do log the memory usage at the process level here ...

openfl/component/aggregator/aggregator.py
```diff
@@ -667,6 +672,44 @@ def send_local_task_results(
 
             self._end_of_round_with_stragglers_check()
 
+    def get_memory_usage(self, round_number, metric_origin):
```
teoparvanov, line 675: maybe call this `get_host_memory_usage` ?

teoparvanov (commented):
A couple of comments, as per our offline discussion

openfl/component/collaborator/collaborator.py
payalcha: Yes adding this details in the JSON also.

openfl/component/aggregator/aggregator.py
payalcha: I added details of process as well in get_memory_usage function. Is function name still needed to be changed to get_host_memory_usage?

openfl/component/collaborator/collaborator.py
payalcha: Added process related memory as well in the result json structure.

tests/end_to_end/models/participants.py
payalcha: Memory details are not needed for all runs. Basically it will be better for performance and Memory Leak related tests I think. 
Please let me know if still needs to change this.

openfl/component/aggregator/aggregator.py
teoparvanov: Makes sense, dropping my comment.

openfl/component/collaborator/collaborator.py
```diff
@@ -171,6 +174,24 @@ def run(self):
 
                 # Cleaning tensor db
                 self.tensor_db.clean_up(self.db_store_rounds)
+                if self.log_memory_usage:
+                    # This is the place to check the memory usage of the collaborator
+                    self.logger.info("*****************COLLABORATOR LOGS*******************************")
```
teoparvanov, line 179: ` ********* COLLABORATOR MEMORY USAGE ********* `?

teoparvanov (approved):
LGTM, thanks @payalcha !

PR #1156, opened 2024-11-19
Deprecate Python Native API

tanwarsh:
- Marked the Python Native API as deprecated in all docs
- Moved the tutorials based on this API under deprecated/native_api
- Removed dependency on python native api code ([openfl-workspace/torch_llm_horovod/src/InHorovodrun.py](https://github.com/securefederatedai/openfl/compare/develop...tanwarsh:openfl:19-Nov-24?expand=1#diff-869591ea31812a4fc5392bb3f06fe5c0df76339e3d67a560f6f06108f0482203)).


rahulga1 (approved):

rajithkrishnegowda (approved):

openfl-workspace/torch_llm_horovod/src/InHorovodrun.py
```diff
@@ -50,7 +50,7 @@ def get_args():
 
 def main():
     logger = getLogger(__name__)
-    fx.setup_logging(level="INFO", log_file=None)
+    setup_logging
```
theakshaypant: Do the 2 functions have the exact same functionality?
Also is the `()` not required here?

theakshaypant (approved):
LGTM, just a small question.

openfl-workspace/torch_llm_horovod/src/InHorovodrun.py
tanwarsh: Thanks for pointing this out. I missed adding (). and yes the functionality is same for both function apart from importing tensorflow in native api which is not required here.

openfl-workspace/torch_llm_horovod/src/InHorovodrun.py
```diff
@@ -9,7 +9,7 @@
 
 import horovod.torch as hvd
 
-import openfl.native as fx
```
teoparvanov, line 12: Why is this change of `InHorovodrun.py` needed, given that at this stage we're just deprecating the Python API?

teoparvanov (approved):
LGTM!

PR #1158, opened 2024-11-19
Add pytorch MNIST Workflow tutorial with FedProx

ynonflumintel:
This branch introduces a tutorial for setting up a neural network training federation using pytorch, the MNIST dataset using the openfl workflow API.


rahulga1:
@ynonflumintel  I think this is not the correct file in this commit, I dont see workflow api getting used in this file. 
can you please confirm?

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,632 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    " #Get the current values of the plan. Each of these can be overridden\n",
+    "import json\n",
+    "print(json.dumps(fx.get_plan(), indent=4, sort_keys=True))"
```
teoparvanov: AFAIK, there's no notion of FL plan in Workflow API. Maybe consider dropping this cell?

ynonflumintel: I pushed with some uncommitted changes, it's fixed now

ynonflumintel:
> @ynonflumintel I think this is not the correct file in this commit, I dont see workflow api getting used in this file. can you please confirm?

I pushed the wrong version of the file before, please take a look now

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,343 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Federated FedProx PyTorch MNIST Workflow API Tutorial\n",
+    "This notebook mimics Federated_FedProx_PyTorch_MNIST_Tutorial using the workflow API.\n"
```
teoparvanov: I don't think we need to reference the deprecated (and soon to be deleted legacy notebook). Instead, I'd suggest providing a short overview of the new notebook (FedProx algo, Workflow API, LocalRuntime - scalable to a federated setting in the future).

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,343 @@
...
+    "        self.model = model\n",
+    "        self.optimizer = optimizer\n",
+    "        self.rounds = rounds\n",
+    "        self.loss = 0.\n",
+    "\n",
+    "    @aggregator\n",
+    "    def start(self):\n",
+    "        print(f'Performing initialization for model')\n",
+    "        self.collaborators = self.runtime.collaborators\n",
+    "        self.private = 10\n",
```
teoparvanov: The `private` field isn't really needed here. It's there in the 101 tutorial to illustrate how certain attributes can be explicitly filtered out when switching to the next step of the flow.

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,343 @@
...
+    "        self.current_round = 0\n",
+    "        self.next(self.aggregated_model_validation, foreach='collaborators', exclude=['private'])\n",
+    "\n",
+    "    def compute_accuracy(self, data_loader):\n",
```
teoparvanov: For improved readability of the federated flow spec, I suggest to keep utility methods towards the bottom.

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
teoparvanov, line 1: Let's move this notebook together with the other experimental/workflow-based ones. Also, consider prefixing it as `105_`, to follow the "curriculum module" numbering pattern. Alternatively, we could consider prefixing it as `401_`, and renaming the non-IID FedProx notebook as `402_`. WDYT?

teoparvanov (approved):
Looks great, thanks @ynonflumintel ! I just have a couple of small suggestions:

teoparvanov (commented):
PS: don't forget to sign off your commits for a successful CI run (DCO job):
```shell
git commit -s -m "..."
```

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,343 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Federated FedProx PyTorch MNIST Workflow API Tutorial\n",
```
MasterSkepticista: ~Federated~ FedProx PyTorch MNIST Tutorial using Workflow API

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,343 @@
...
+    "import torch.utils.data\n",
+    "import torchvision\n",
+    "import torchvision.transforms as transforms\n",
+    "\n",
+    "from openfl.utilities.optimizers.torch.fedprox import FedProxAdam\n",
+    "\n",
+    "from openfl.experimental.interface import FLSpec, Aggregator, Collaborator\n",
+    "from openfl.experimental.runtime import LocalRuntime\n",
+    "from openfl.experimental.placement import aggregator, collaborator\n",
+    "\n",
```
MasterSkepticista: Suggest splitting the cell here. Helps partition the context. The functions in this cell should be close to where they are actually called.

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
```diff
@@ -0,0 +1,343 @@
...
+    "        self.model.train()\n",
+    "        self.optimizer = get_optimizer(self.model)\n",
+    "        for batch_idx, (data, target) in enumerate(self.train_loader):\n",
+    "            self.optimizer.zero_grad()\n",
+    "            output = self.model(data)\n",
+    "            loss = F.cross_entropy(output, target)\n",
+    "            loss.backward()\n",
+    "            self.optimizer.step()\n",
+    "\n",
+    "            if (len(data) * batch_idx) / len(self.train_loader.dataset) >= log_threshold:\n",
```
MasterSkepticista: Instead of this complicated logic of calculating "quarter" of samples used, it is OK to print every 10 or 100 steps. We need to be cautious of all the cognitive overhead we put on the user.

MasterSkepticista (changes requested):
General comment on the entire notebook: Include output of all the cells. New users should have visibility into the cells and evaluate for themselves if the notebook fulfils their goal of visit. Not showing output, can put-off newcomers because they end goal is not "visible".

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: I'm not using Fedprox as the aggregation algorithm here. Is it implemented somewhere and can be integrated here?

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: So if you remove the original and only keep this notebook, don't you want it here?

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: you need them defined before you use them so that you can use "run above" in the notebook and know it includes all of the dependencies

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: I don't mind adding a prefix to the file name, which one do you want?
Where do you want this file to be moved to?
What does IID stand for?

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
teoparvanov: It's still available in he [101_MNIST](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/101_MNIST.ipynb) notebook, which is not going anywhere :)

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: done

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: done

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
teoparvanov: I'd suggest `401_`, but currently this prefix is used for the "non-IID" notebook. You can rename the non-IID one to `402_`, and use the `401_` prefix for your notebook. 

Regarding IID, it means "Independent and Identically Distributed", i.e. all elements of the data set are sampled from the same statistical distribution. Non-IID means that dataset elements come from different distributions, which could be an obstacle for the convergence of the "vanilla" FedAvg. Enter FedProx, which offers better convergence for non-IID datasets (among other benefits related to dataset size and compute power heterogeneity among collaborators). You can read more about it [here](https://arxiv.org/pdf/1812.06127.pdf).

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: I ran experiments where there were a couple of thousand steps (small chunks), it's too spammy IMO

ynonflumintel:
> General comment on the entire notebook: Include output of all the cells. New users should have visibility into the cells and evaluate for themselves if the notebook fulfils their goal of visit. Not showing output, can put-off newcomers because they end goal is not "visible".

Thought it would be too verbose and increase the file size significantly, but I added them as per your request

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: please be specific, which file do you want me to name what - I'm getting a little lost

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
ynonflumintel: Oh, okay, done

openfl-tutorials/Federated_FedProx_PyTorch_MNIST_Workflow_Tutorial.ipynb
teoparvanov: Actually FedProx is not a "pure" aggregation algorithm, but a combination of the FedProx optimizer (which adds the so called "proximal" element to the loss function), and FedAvg for the aggregation part.

teoparvanov (approved):

PR #1159, opened 2024-11-19
Memory Leak Logs

payalcha:
Reviewed PR - https://github.com/securefederatedai/openfl/pull/1155 (Had to close as some unsigned commits were there)

**Openfl component change**
Add memory leak logs for aggregator and collaborator, if **log_memory_usage** flag can be set for ubuntu.

Changes - 
1. aggregator.py - Function get_memory_usage added , which collects process as well as system log.
2. collaborator.py - Function get_memory_usage added , which collects process as well as system log.
**Output -** 
agg_mem_details.json - details of aggregator in json format
<collaborator-name>_mem_details.json - details of aggregator in json format

Structure -
    {
        "round_number": 0,
        "metric_origin": "collaborator1",
        "process_memory": 1155.19,
        "virtual_memory": {
        },
        "swap_memory": {
        }
    }

**Test changes**

1. log_memory_usage flag is added in conftest.py and passed to test when set.
2. **Correction** -
federation_helper.py - _verify_completion_for_participant - added timeout on the basis of num_rounds.


rahulga1 (approved):

teoparvanov (approved):

MasterSkepticista (changes requested):
@payalcha Per our last discussion, I'm not in favor of this PR for the following reasons:
1. There is evidence of memory growth from internal experiments, which means we need to identify the source of the growth/leak to disposition it. If the investigation suggests changes to `openfl` to **fix** the issue, that is a great PR to have.
2. In a situation where the leak/growth is due to an underlying framework (pandas, tf, torch etc.), or it falls outside of `openfl` purview, it needs to be tracked in the respective framework repository.

That said, if the overall decision is to still add this telemetry functionality, I suggest you use decorators over functions that you want to log memory for, within your CI tests. All helper functions can live within `tests/`. I don't think framework source code (anything within `openfl/`) is the right place for it, because users are neither required nor encouraged to use it in their usual experiments.

openfl/component/aggregator/aggregator.py
```diff
@@ -969,6 +972,11 @@ def _end_of_round_check(self):
         for task_name in all_tasks:
             self._compute_validation_related_task_metrics(task_name)
 
+        if self.log_memory_usage:
+            # This is the place to check the memory usage of the aggregator
+            memory_detail = get_memory_usage(self.logger, self.round_number, "aggregator")
```
MasterSkepticista: Please avoid passing logger objects. I am not sure why it helps. Each module (.py) file can define its own logger.

openfl/utilities/logs.py
```diff
@@ -57,3 +58,51 @@ def setup_loggers(log_level=logging.INFO):
     formatter = logging.Formatter("[%(asctime)s][%(name)s][%(levelname)s] - %(message)s")
     handler.setFormatter(formatter)
     root.addHandler(handler)
+
+
+def get_memory_usage(logger, round_number, metric_origin):
```
MasterSkepticista: Add type hints.

openfl/component/collaborator/collaborator.py
```diff
@@ -171,6 +174,14 @@ def run(self):
...
                 # Cleaning tensor db
                 self.tensor_db.clean_up(self.db_store_rounds)
+                if self.log_memory_usage:
+                    # This is the place to check the memory usage of the collaborator
+                    memory_detail = get_memory_usage(
+                        self.logger, round_number, metric_origin=self.collaborator_name
+                    )
+                    memory_details.append(memory_detail)
+        if self.log_memory_usage:
+            self.logger.info(f"Publish memory usage: {self.memory_details}")
```
MasterSkepticista: I suggest logging once at the end of experiment, or only within the function call.

openfl/utilities/logs.py
```diff
@@ -57,3 +58,51 @@ def setup_loggers(log_level=logging.INFO):
...
+    """Logs the memory usage statistics for the given round number.
+
+    This method retrieves the current virtual and swap memory usage statistics
+    using the psutil library, formats them into a dictionary, and logs the
+    information using the logger.
+
+    Args:
+        round_number (int): The current round number for which memory usage is being logged.
+    """
+    process = psutil.Process()
```
MasterSkepticista: For readability sake: `psutil.Process(os.getpid())`

openfl/utilities/logs.py
```diff
@@ -57,3 +58,51 @@ def setup_loggers(log_level=logging.INFO):
...
     handler.setFormatter(formatter)
     root.addHandler(handler)
+
+
+def get_memory_usage(logger, round_number, metric_origin):
+    """Logs the memory usage statistics for the given round number.
+
+    This method retrieves the current virtual and swap memory usage statistics
+    using the psutil library, formats them into a dictionary, and logs the
+    information using the logger.
```
MasterSkepticista:
```python
"""Logs memory usage details of the current python process."""
```

openfl/utilities/logs.py
MasterSkepticista: Instead of passing `round_number` and `metric_origin`, pass a single argument called `msg`, which is a string that gets prefixed on every print statement made by the function. The goal of functions is to have reusability and keep them generic without compromising readability.

Also suggest renaming the function to `log_memory_usage(msg: str)`.

This is my recommended signature:

```python
def log_memory_usage(msg: str = "") -> dict:
  """Logs memory usage details of the current process.
  
  Args:
    msg: Optional prefix message for logging.

  Returns:
    A dict of memory usage details.
  """
  pass
```

openfl/utilities/logs.py
```diff
@@ -57,3 +58,51 @@ def setup_loggers(log_level=logging.INFO):
...
+        },
+        "swap_memory": {
+            "total": round(swap_memory.total / (1024**2), 2),
+            "used": round(swap_memory.used / (1024**2), 2),
+            "free": round(swap_memory.free / (1024**2), 2),
+            "percent": swap_memory.percent,
+        },
+    }
+    logger.info(
+        f"**************** End of round check: {metric_origin} Memory Logs ******************"
```
MasterSkepticista: Not required, function may be called outside of round checks.

MasterSkepticista (changes requested):

openfl/utilities/logs.py
payalcha: Code change done.

openfl/component/aggregator/aggregator.py
payalcha: As discussed removed the logger.

openfl/component/collaborator/collaborator.py
payalcha: Done. Now details will get publish at the end of run.

openfl/component/collaborator/collaborator.py
```diff
@@ -171,6 +174,13 @@ def run(self):
 
                 # Cleaning tensor db
                 self.tensor_db.clean_up(self.db_store_rounds)
+                if self.log_memory_usage:
+                    # This is the place to check the memory usage of the collaborator
+                    memory_detail = get_memory_usage()
+                    memory_detail["round_number"] = round_number
```
MasterSkepticista, line 180: Think you forgot to add entity name?

openfl/component/aggregator/aggregator.py
```diff
@@ -969,6 +972,12 @@ def _end_of_round_check(self):
         for task_name in all_tasks:
             self._compute_validation_related_task_metrics(task_name)
 
+        if self.log_memory_usage:
+            # This is the place to check the memory usage of the aggregator
+            memory_detail = get_memory_usage()
+            memory_detail["round_number"] = self.round_number
```
MasterSkepticista, line 978: Entity name?

MasterSkepticista (approved):

openfl/component/collaborator/collaborator.py
payalcha: Added.

openfl/component/aggregator/aggregator.py
payalcha: Added

PR #1174, opened 2024-11-26
Updated Keras and TensorFlow Task Runner and related workspaces.

tanwarsh:
Related issue https://github.com/securefederatedai/openfl/issues/973.

Changes in this PR:
1. Updated Tensorflow and Keras to Tensorflow v2.18.0 and Keras 3
2. Updated openfl/federated/task/runner_kr.py
3. Standardizing the workspace.
4. Removed workspaces.
6. Updated workspaces.
    1. keras_cnn_mnist
    2. keras_nlp 
    3. keras_2dunet


tanwarsh:
I will add code changes for the remaining Keras workspace - Keras_cnn_with_compression and keras_nlp.

@teoparvanov @kta-intel @psfoley, I was facing an issue with optimizer weights due to resetting the weights before saving them to the tensor database and using them to rebuild the model, which resulted in no improvement in accuracy. I have skipped the rebuild (and hence resetting the model optimizer weights) for the first round. This can be further discussed, and we can proceed with these changes once all the pipelines are working and the workspace changes are also completed.

.github/workflows/ubuntu.yml
```diff
@@ -32,10 +32,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
```
teoparvanov: Is this a requirement of the new versions of TF/Keras?

openfl/federated/task/runner_keras.py
```diff
@@ -60,7 +60,7 @@ def rebuild_model(self, round_num, input_tensor_dict, validation=False):
         """
         if self.opt_treatment == "RESET":
             self.reset_opt_vars()
-            self.set_tensor_dict(input_tensor_dict, with_opt_vars=False)
+            self.set_tensor_dict(input_tensor_dict, with_opt_vars=True)
```
teoparvanov: Intuitively, if the opt treatment is "RESET", shouldn't we have `with_opt_vars=False` ?

openfl/federated/task/runner_keras.py
```diff
@@ -281,7 +280,12 @@ def _get_weights_names(obj):
         Returns:
             weight_names (list): The weight name list.
         """
-        weight_names = [weight.name for weight in obj.weights]
+        if isinstance(obj, ke.optimizers.Optimizer):
+            weight_names = [weight.name for weight in obj.variables]
+        else:
+            weight_names = [
+                layer.name + "/" + weight.name for layer in obj.layers for weight in layer.weights
+            ]
```
teoparvanov, line 287: Which other optimizers would trigger the `else` block?

openfl/federated/task/runner_keras.py
```diff
@@ -298,10 +302,22 @@ def _get_weights_dict(obj, suffix=""):
...
+        if isinstance(obj, ke.optimizers.Optimizer):
+            weight_names = [weight.name for weight in obj.variables]
+            weights_dict = {
+                weight_names[i] + suffix: weight.numpy()
+                for i, weight in enumerate(copy.deepcopy(obj.variables))
+            }
+        else:
+            weight_names = [
+                layer.name + "/" + weight.name for layer in obj.layers for weight in layer.weights
+            ]
```
teoparvanov: This code is repeated several times - could you extract it in a separate function?

openfl-workspace/keras_nlp/requirements.txt
teoparvanov, line 1: I don't think the `keras_nlp` is covered by the CI. Have you tested it locally after upgrading to Keras 3?

teoparvanov (commented):
LGTM overall. A couple of questions came up during my read-through:

.github/workflows/ubuntu.yml
tanwarsh: yes, tensorflow and keras are not compatible with python < 3.9

openfl/federated/task/runner_keras.py
tanwarsh: This is done as when optmizers weights are reset. Leraning rate is zero and model accuracy remiains around ~0.125.

openfl/federated/task/runner_keras.py
tanwarsh: else is for model weights and if is for Optimizer weights

openfl-workspace/keras_nlp/requirements.txt
tanwarsh: Yes, I have tested Keras nlp and Kears cnn mnist both locally.

openfl/federated/task/runner_keras.py
tanwarsh: removed duplicate code.

.github/workflows/pki.yml
```diff
@@ -22,10 +22,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
```
payalcha: Please change the task_runner_e2e.yaml workflow as well. It runs nightly so not failed for the PR.

.github/workflows/double_ws_export.yml
```diff
@@ -22,10 +22,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
```
payalcha: If 3.8 not supported anymore, change the README documents as well for the same.

openfl-workspace/keras_cnn_mnist/src/taskrunner.py
```diff
@@ -3,11 +3,10 @@
 
 """You may copy this file as the starting point of your own model."""
 
-import tensorflow.keras as ke
-from tensorflow.keras import Sequential
-from tensorflow.keras.layers import Conv2D
-from tensorflow.keras.layers import Dense
-from tensorflow.keras.layers import Flatten
+from keras.models import Sequential
```
payalcha, line 6: Tensorflow Keras is still evident and we see many examples in the ML community for the same.
Shouldn't we add separate workspace for Keras 3.0 for now and remove this eventually. 

Or we are saying that we are not going to support tensorflow keras anymore.

openfl/federated/task/runner_keras.py
```diff
@@ -17,8 +18,8 @@
 
 with catch_warnings():
     simplefilter(action="ignore")
+    import keras as ke
     import tensorflow as tf
```
payalcha: Ideally tensorflow must not be used with Keras as Keras implementation must be free of any backend while running.
Please find the alternative for the same.

.github/workflows/pki.yml
tanwarsh: done

openfl/federated/task/runner_keras.py
```diff
@@ -348,46 +361,23 @@ def set_tensor_dict(self, tensor_dict, with_opt_vars):
...
+            opt_weight_names = self._get_weights_names(self.model.optimizer)
             opt_weights_dict = {name: tensor_dict[name] for name in opt_weight_names}
             self._set_weights_dict(self.model, model_weights_dict)
             self._set_weights_dict(self.model.optimizer, opt_weights_dict)
 
     def reset_opt_vars(self):
         """Resets the optimizer variables."""
-        for var in self.model.optimizer.variables():
+        for var in self.model.optimizer.variables:
             var.assign(tf.zeros_like(var))
```
kminhta: Re: learning rate resetting to 0
This line seems to be the cause of that issue. It's a legacy line so I am not sure the context behind it, but for some reason this line will reset every variable of the optimizer (including learning rate) to 0, which is why the model is not learning when `self.opt_treatment == "RESET"`. The torch task runner actually seems to just pass through this method entirely ([see](https://github.com/securefederatedai/openfl/blob/develop/openfl/federated/task/runner_pt.py#L453))

FYI, we should also get rid of any `tf` calls or imports to remove any dependency on tensorflow in the keras task runner

openfl/federated/task/runner_keras.py
kminhta: Once `reset_opt_vars()` is fixed, let's set this back to `with_opt_vars=False` since the intention is to not rebuild the model with the global optimizer (and "reset" it back to the local optimizer)

openfl/federated/task/runner_keras.py
kminhta: we should get rid of this import. I think the only place where `tf` is called is in `reset_opt_vars()`

openfl-workspace/keras_cnn_mnist/src/taskrunner.py
kminhta: You're right that we should not remove support for tf-keras. We will eventually update the existing tf workspaces to tf-keras and deprecate native tf support since keras has been adopted as the "official" high-level API for tf. 

The goal in this PR is to have an entirely native keras task runner and accompanying workspace since we will eventually want to support jax-keras and torch-keras as well.

kminhta:
> @teoparvanov @kta-intel @psfoley, I was facing an issue with optimizer weights due to resetting the weights before saving them to the tensor database and using them to rebuild the model, which resulted in no improvement in accuracy. I have skipped the rebuild (and hence resetting the model optimizer weights) for the first round. This can be further discussed, and we can proceed with these changes once all the pipelines are working and the workspace changes are also completed.

Despite the name, I think `rebuild_model()` is actually a necessary step in the first round since the flow begins with the aggregator sending out a global model, so the collaborators do need to rebuild their initially compiled models with the global model's weights. We should be able to bring it back once the comments around `reset_opt_vars()` are addressed

openfl/federated/task/runner_keras.py
```diff
@@ -17,8 +18,8 @@
 
 with catch_warnings():
     simplefilter(action="ignore")
+    import keras as ke
```
psfoley: I realize this is carried over from prior examples, but `ke` is not a standard abbreviation for `keras`: https://keras.io/examples/vision/image_classification_with_vision_transformer/

While all of the keras code is being updated, I would recommend changing this to the following to make the code slightly easier to read for developers:

```suggestion
    import keras
```

openfl/federated/task/runner_keras.py
```diff
@@ -98,7 +98,8 @@ def train(
             raise KeyError("metrics must be defined")
 
         # rebuild model with updated weights
-        self.rebuild_model(round_num, input_tensor_dict)
+        if round_num > 0:
```
psfoley: I believe this will result in the aggregator's initial model not being used for the first round. This will be problematic if a pretrained model is used as the basis for federated fine-tuning.

openfl/federated/task/runner_keras.py
```diff
@@ -220,23 +222,20 @@ def validate(self, col_name, round_num, input_tensor_dict, **kwargs):
         else:
             batch_size = 1
 
-        self.rebuild_model(round_num, input_tensor_dict, validation=True)
+        if round_num > 0:
```
psfoley: This should also be reverted.

psfoley (dismissed):

.github/workflows/double_ws_export.yml
psfoley: Keras 3.0 drops python 3.8 support, but other frameworks (i.e. PyTorch) still support it. If we decide to remove support for Python 3.8, it should be its own PR.

tanwarsh: raised PR https://github.com/securefederatedai/openfl/pull/1189 for version change

openfl/federated/task/runner_keras.py
tanwarsh: reverted.

openfl/federated/task/runner_keras.py
tanwarsh: reverted.

openfl/federated/task/runner_keras.py
tanwarsh: done

openfl/federated/task/runner_keras.py
tanwarsh: done

openfl/federated/task/runner_keras.py
tanwarsh: done

openfl/federated/task/runner_keras.py
tanwarsh: this code is removed.

openfl-workspace/keras_cnn_mnist/src/taskrunner.py
tanwarsh: workspaces for older keras are removed/updated in this PR and  support for tf - keras will be marked deplrecated in https://github.com/securefederatedai/openfl/pull/1204 and eventually removed in future.

openfl/federated/task/runner_keras.py
tanwarsh: tensorflow is removed.

openfl-workspace/keras_cnn_mnist/plan/plan.yaml
```diff
@@ -1,13 +1,13 @@
...
 aggregator :
   defaults : plan/defaults/aggregator.yaml
   template : openfl.component.Aggregator
   settings :
-    init_state_path : save/keras_cnn_mnist_init.pbuf
-    best_state_path : save/keras_cnn_mnist_best.pbuf
-    last_state_path : save/keras_cnn_mnist_last.pbuf
+    init_state_path : save/init.pbuf
+    best_state_path : save/best.pbuf
+    last_state_path : save/last.pbuf
```
MasterSkepticista, line 10: Overall: Good that you generalized paths and module names throughout the PR.

openfl-workspace/keras_cnn_mnist/src/taskrunner.py
MasterSkepticista, line 1: Since this is Keras3, what is the default backend? 

My suggestion would be to set the backend as an env variable explicitly. That informs the user how to switch the backends. It would also be good to test all 3 backends with the same code and ensure it passes (because OpenFL is agnostic to backend)

openfl-workspace/keras_nlp/src/dataloader.py
```diff
@@ -68,7 +66,7 @@ def get_train_loader(self, batch_size: Optional[int] = None) -> Iterator[List[np
         return self._get_batch_generator(X1=self.X_train[0], X2=self.X_train[1],
                                          y=self.y_train, batch_size=batch_size)
 
-    def get_valid_loader(self, batch_size: Optional[int] = None) -> Iterator[List[np.ndarray]]:
+    def get_valid_loader(self, batch_size: Optional[int] = None):
```
MasterSkepticista: Returns a callable instead?

openfl/federated/task/runner_keras_tf.py
```diff
@@ -0,0 +1,480 @@
+# Copyright 2020-2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""
+Base classes for developing a ke.Model() Federated Learning model.
```
theakshaypant: Nit: Remove the `ke` acronym.

openfl-workspace/keras_nlp/src/taskrunner.py
MasterSkepticista, line 1: A reader cannot tell what this example is about (model/use case). Suggest adding a header (also making the license header consistent with other files) that tells what architecture, and what use case this example is for.

theakshaypant (commented):
Is openfl setup not affected by these changes?

openfl/federated/__init__.py
```diff
@@ -14,7 +14,11 @@
     from openfl.federated.data import FederatedDataSet  # NOQA
     from openfl.federated.data import KerasDataLoader, TensorFlowDataLoader
     from openfl.federated.task import FederatedModel  # NOQA
-    from openfl.federated.task import KerasTaskRunner, TensorFlowTaskRunner
+    from openfl.federated.task import (
+        KerasTaskRunner,
+        TensorFlowTaskRunner,
+        KerasTensorFlowTaskRunner,
```
MasterSkepticista: ~Rename all instances of `KerasTensorFlowTaskRunner` to `TensorFlowKerasTaskRunner`. It may feel like a nitpick, but the legacy API namespace is actually `tf.keras`.~

~Current naming may confuse the user to interpret `KerasTensorFlowTaskRunner` as "Tensorflow backend of Keras3"~

Edit: OK, I am confused now. What exactly are we supporting? I thought this was about updating to `KerasTaskRunner` (i.e. the multi-backend Keras3) and `TensorFlowKerasTaskRunner` (i.e. the legacy tf 2.x namespace). Please clarify?

MasterSkepticista:
I think this is a good place to refactor the directory structure:

```shell
openfl-workspace (in future, "examples")
 |_ keras
 |  |_ mnist
 |  |_ lstm
 |_ (other backends in future PRs)
```

Thoughts?

teoparvanov:
> I think this is a good place to refactor the directory structure:
> 
> ```shell
> openfl-workspace (in future, "examples")
>  |_ keras
>  |  |_ mnist
>  |  |_ lstm
>  |_ (other backends in future PRs)
> ```
> 
> Thoughts?

@MasterSkepticista , would it be acceptable to do this in a follow-up PR?

openfl/federated/task/runner_keras_tf.py
tanwarsh: Removed.

openfl-workspace/keras_cnn_mnist/src/taskrunner.py
tanwarsh: By default it is tensorflow. This is already planned. There will be followup PR which will restructure code to support all backends (tensorflow, JAX and Torch)

openfl/federated/__init__.py
tanwarsh: - KerasTaskRunner: Keras 2 and TensorFlow 2.13
- TensorFlowTaskRunner: Keras 2 and TensorFlow 2.13
- KerasTensorFlowTaskRunner: Updated to Keras 3 and TensorFlow 2.18

KerasTaskRunner is used by 5 workspaces prior to this PR:

- Keras_cnn_mnist
- Keras_cnn_with_compression
- Keras_nlp
- tf_3dunet_brats
- tf_cnn_histology

TensorFlowTaskRunner is used by tf_2dunet.

After this PR and https://github.com/securefederatedai/openfl/pull/1204 we will have all 6 workspaces updated with the latest Keras and TensorFlow versions. https://github.com/securefederatedai/openfl/pull/1204 will also move older TensorFlow workspace code to the deprecated folder, which will be removed in the future. Therefore, KerasTaskRunner and TensorFlowTaskRunner will be removed later.

openfl-workspace/keras_nlp/src/dataloader.py
tanwarsh: fixed.

openfl/federated/task/runner_keras_tf.py
```diff
@@ -0,0 +1,480 @@
...
+from openfl.utilities import Metric, TensorKey, change_tags
+from openfl.utilities.split import split_tensor_dict_for_holdouts
+
+with catch_warnings():
+    simplefilter(action="ignore")
+    import keras
+
+
+class KerasTensorFlowTaskRunner(TaskRunner):
+    """The base model for Keras models in the federation.
```
teoparvanov: nit - base task runner

teoparvanov (approved):
Great work, thanks @tanwarsh ! This PR brings OpenFL back into the TensorFlow game 😉

openfl/federated/__init__.py
kminhta: I might've missed some discussion. What is the rationale for keeping `KerasTaskRunner` and introducing a new `KerasTensorFlowTaskRunner` instead of just updating `KerasTaskRunner` as a multi-backend Keras3 (as @MasterSkepticista mentioned)?

If we have large enough justification to keep Keras 2 support in this release, we should just add another class to `runner_keras.py`, imo. After this and that other PR you mentioned, it would look like:
- `runner_keras.KerasTaskRunner` -> supports Keras 3 and TensorFlow 2.18
- `runner_keras.Keras2TaskRunner` -> supports Keras 2 and TensorFlow 2.13
- remove/deprecate `TensorFlowTaskRunner`

`runner_keras.KerasTaskRunner` should support anything that Keras supports (i.e. TF, keras-JAX, keras-torch, etc.), so we won't need a separate `KerasTensorFlowTaskRunner`

MasterSkepticista: > * KerasTaskRunner: Keras 2 and TensorFlow 2.13
> * TensorFlowTaskRunner: Keras 2 and TensorFlow 2.13
> * KerasTensorFlowTaskRunner: Updated to Keras 3 and TensorFlow 2.18
> 
> KerasTaskRunner is used by 5 workspaces prior to this PR:
> 
> * Keras_cnn_mnist
> * Keras_cnn_with_compression
> * Keras_nlp
> * tf_3dunet_brats
> * tf_cnn_histology
> * TensorFlowTaskRunner is used by tf_2dunet.
> 
> After this PR and #1204 we will have all 6 workspaces updated with the latest Keras and TensorFlow versions. #1204 will also move older TensorFlow workspace code to the deprecated folder, which will be removed in the future. Therefore, KerasTaskRunner and TensorFlowTaskRunner will be removed later.

I remain confused with this. Here is my proposal:

* `KerasTaskRunner`: A new Keras 3 task runner which is agnostic of any backend that "keras" provides support for, i.e., any keras 3 taskrunner workspace should work with torch/tf/jax/numpy backends.
* `TensorFlowTaskRunner`: The `tf.keras` TaskRunner for `tensorflow<=2.16,>=2.0`. This can remain in maintenance mode for about a year, because community no longer writes entire NNs in pure TensorFlow anymore. They may write using `tf.keras`, but Google will likely deprecate this fully next year.
* `PyTorchTaskRunner`: as-is.

Note: This may mean removing support (+code/examples) for `tensorflow==1.x`, which is what we refer by `TensorFlowTaskRunner` today. `tf1.x` API is so old and practically unusable, there is no harm removing it without looking back.

Cc: @kta-intel, https://github.com/securefederatedai/openfl/pull/985#issuecomment-2163817246 https://github.com/securefederatedai/openfl/pull/985#issuecomment-2163323526

MasterSkepticista: > I might've missed some discussion. What is the rationale for keeping `KerasTaskRunner` and introducing a new `KerasTensorFlowTaskRunner` instead of just updating `KerasTaskRunner` as a multi-backend Keras3 (as @MasterSkepticista mentioned)?
> 
> If we have large enough justification to keep Keras 2 support in this release, we should just add another class to `runner_keras.py`, imo. After this and that other PR you mentioned, it would look like:
> 
> * `runner_keras.KerasTaskRunner` -> supports Keras 3 and TensorFlow 2.18
> * `runner_keras.Keras2TaskRunner` -> supports Keras 2 and TensorFlow 2.13
> * remove/deprecate `TensorFlowTaskRunner`
> 
> `runner_keras.KerasTaskRunner` should support anything that Keras supports (i.e. TF, keras-JAX, keras-torch, etc.), so we won't need a separate `KerasTensorFlowTaskRunner`

I agree, Kevin. One tiny detail though: Keras 2 isn't commonly known, except for us since we closely track tf2 and keras3 naming. Users would only know `tf.keras` and `keras>=3`. So my suggestion (subject to having a consensus) would be to deprecate `TensorFlowTaskRunner` as you mentioned, since it refers to `tf1.x`, but retain the same name for Keras2 (equivalently `tf.keras`).

TL;DR: We deprecate all `tf1.x` code in OpenFL, but retain the class name `TensorFlowTaskRunner` for all `tf2.x` code. 

WDYT?

kminhta: > I agree, Kevin. One tiny detail though: Keras 2 isn't commonly known, except for us since we closely track tf2 and keras3 naming. Users would only know `tf.keras` and `keras>=3`. So my suggestion (subject to having a consensus) would be to deprecate `TensorFlowTaskRunner` as you mentioned, since it refers to `tf1.x`, but retain the same name for Keras2 (equivalently `tf.keras`).
> 
> TL;DR: We deprecate all `tf1.x` code in OpenFL, but retain the class name `TensorFlowTaskRunner` for all `tf2.x` code.
> 
> WDYT?

I like this proposal, it's a good idea. +1 from me!
Using `TensorFlowTaskRunner` as our vehicle for Keras2 support via `tf.keras` makes a lot more sense

tanwarsh:
> Is openfl setup not affected by these changes?

No this will not affect openfl setup.

tanwarsh:
@teoparvanov @MasterSkepticista @kta-intel 

As discussed I have migrated workspaces mentioned below to latest tensorflow and keras 3 versions and rest of the workspaces are removed.

1. keras_cnn_mnist
2. keras_nlp
3. keras_2dunet (renamed from tf_2dunet)

openfl/federated/__init__.py
tanwarsh: TensorFlowTaskRunner  and KerasTensorFlowTaskRunner code removed. KerasTaskRunner is update to support latest versions.

rahulga1 (approved):

teoparvanov (approved):
LGTM!

Issue #1074, opened 2024-10-01
AttributeError: module 'importlib' has no attribute 'util'


kminhta:
**Description**
When attempting to import the `util` submodule from the `importlib` module, an `AttributeError` is raised. This is problematic as `openfl` has a few `importlib.util` calls.

One example:
```
Python 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from openfl.federated import PyTorchDataLoader
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "openfl/federated/__init__.py", line 9, in <module>
    from openfl.federated.data import DataLoader  # NOQA
  File "openfl/federated/data/__init__.py", line 12, in <module>
    if importlib.util.find_spec("tensorflow") is not None:
AttributeError: module 'importlib' has no attribute 'util'
```

_Note 1_ : it does not seem to be triggered when calling fx commands through `fx` (i.e.  `fx plan initialize` does not seem to trigger the error even though the dataloader is called during this command)

_Note 2_ : I have only been able to reproduce it when I install `openfl` via `pip install .` as opposed to `pip install -e .`

**Steps to reproduce the behavior:**
1. `python3.10 -m venv venv`
2. `source venv/bin/activate`
3. `git clone https://github.com/securefederatedai/openfl.git`
4. `cd openfl`
5. `pip install .`
6. enter python console: `python`
7. from within python console: `from openfl.federated import PyTorchDataLoader`

You should see the above error

**Expected behavior**
`PyTorchDataLoader` should import with no error

**Proposed resolution**
Seems best known practice for using `importlib.util` is to import the submodule directly with `import importlib.util`
Doing this throughout the repo where `import importlib` is called should avoid this and similar issues in the future

rahulga1:
fixed in #1180

PR #1180, opened 2024-11-29
Fix for 'importlib' has no attribute 'util'

tanwarsh:
For more information please refere https://github.com/securefederatedai/openfl/issues/1074


rahulga1 (approved):

MasterSkepticista (approved):
Strange that this appears only now. Have we checked python/library documentation of this module for the reason behind sudden breakage?

PR #1187, opened 2024-12-03
Simplify setup

MasterSkepticista:
About this PR:
* https://github.com/securefederatedai/openfl/pull/1187/commits/6191d8222c8efecf2ed83b092a0b44c9ffa42881 Removes traces of `log_metric`, `log_metric_callback` and `write_metric` as they are not used throughout, and not defined as per python best practices.
* https://github.com/securefederatedai/openfl/pull/1187/commits/e17172ec53f441890d9a52380060e636fc0cc894 Populates all subpackages automatically using setuptools `find_packages`.
* https://github.com/securefederatedai/openfl/pull/1187/commits/6017900b44d3992fd4c9b30058a140dbca88d1a2 Shrinks requirement footprint significantly by removing unused packages, or those not required as a bare minimum by `openfl`.
* Interactive API CI tests have been disabled [here](https://github.com/securefederatedai/openfl/actions)


setup.py
```diff
@@ -61,113 +58,34 @@ def run(self):
...
-    long_description = f.read()
-
 setup(
     name='openfl',
     version='1.6',
-    author='The OpenFL Team',
+    author='OpenFL Team',
     description='Federated Learning for the Edge',
-    long_description=long_description,
+    long_description=open("README.md", encoding="utf-8").read(),
```
psfoley, line 66: 👍

psfoley (changes requested):
The change of removing `log_metric`, `log_metric_callback` and `write_metric` should be reverted. This would remove our ability to publish metrics to Tensorboard: https://github.com/securefederatedai/openfl/blob/7c53fb75a4ab22465c003eb06ad9862b0b76d5c8/docs/developer_guide/advanced_topics/log_metric_callback.rst#L4

Writing logs in the tensorboard format for later analysis is the default in many ([1](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml#L12),[2](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml#L12),[3](https://github.com/securefederatedai/openfl/blob/7c53fb75a4ab22465c003eb06ad9862b0b76d5c8/openfl-workspace/torch_llm_horovod/plan/plan.yaml#L13)) of the existing TaskRunner API examples. This capability should not be removed unless there is a plan to replace it.

MasterSkepticista:
Turns out I confused `log_metric` and `log.metric` in one of the previous PRs, which led to it being removed, giving an impression of dead code.

I have restored TB support and added missing function calls. With this, gramine-based runs fail, likely because python multiprocessing is not supported https://github.com/gramineproject/gramine/issues/1156.

**Edit:** Tensorboard logging is disabled for gramine CI test. This issue is known internally, with no fixes on the horizon.

setup.py
```diff
@@ -61,113 +58,35 @@ def run(self):
...
-        'openfl.utilities.optimizers.numpy',
-        'openfl.utilities.optimizers.torch',
-        'openfl-docker',
-        'openfl-tutorials',
-        'openfl-workspace',
-    ],
+    packages=find_packages(include=("openfl", "openfl.*", "openfl-docker", "openfl-workspace")),
     include_package_data=True,
+    setup_requires=['grpcio-tools>=1.56.2,<1.66.0'],  # ensure it is in-sync with `install_requires`
     install_requires=[
```
theakshaypant, line 72: Some of the minimum version requirements are explicitly mentioned due to SDL requirements. I think we should keep those.

MasterSkepticista: Could you mention which, specifically?
Edit: @theakshaypant We can revisit version pins as we get closer to release, if SDL demands.

openfl/component/aggregator/aggregator.py
MasterSkepticista, line 1: These have been reverted.

setup.py
theakshaypant: `requests>=2.32.0` is one that I can remember on top of my head; but sure, we can create separate PRs for those as required.

PR #1195, opened 2024-12-05
Introduce `callbacks` API

MasterSkepticista:
This PR introduces **callbacks**, an abstract way of running user-defined actions in TaskRunner or Workflow APIs. 

Callbacks can be used to perform actions at different stages of the Federated Learning process. To create a custom callback, a user/developer needs to subclass `openfl.callbacks.Callback` and then implement the necessary methods. This creates a fine separation between framework/user code, and provides a step-function improvement in the time taken to implement ad-hoc tasks by eliminating the time spent by a user or a developer to contextualize framework code.

Callbacks can be triggered on the aggregator and collaborator side for the following events:
* At the beginning of an experiment, e.g. for one-off setups, or state recovery
* At the beginning of every round, e.g. for inspecting aggregated models
* At the end of every round, e.g. for saving model weights
* At the end of an experiment, e.g. to serve/upload artifacts

With this abstraction, common user-facing actions can be instrumented in the experiment without having to modify the source code. Here is an example:

```python
# Define actions on Aggregator-side
aggregator = openfl.component.Aggregator(..., callbacks=[
    openfl.callbacks.MetricWriter("./logs", use_tensorboard=True), 
    openfl.callbacks.LambdaCallback(on_experiment_end=lambda: _start_model_server(exit_in_minutes=10)),
    openfl.callbacks.ModelCheckpoint("./save", task="aggregated_model_validation", monitor="loss", mode="min"),
])

# Define actions on Collaborator-side
collaborator = openfl.component.Collaborator(..., callbacks=[
    openfl.callbacks.MemoryProfiler("./logs"), 
    openfl.callbacks.LambdaCallback(
        on_experiment_start=lambda: print("Experiment started!"),
        on_round_begin=lambda round_num, _: print(f"Starting round {round_num}!"),
        on_round_end=lambda round_num, _: print(f"End of round {round_num}!"),
        on_experiment_end=lambda: print("Experiment ended successfully!"),
    ),
    openfl.callbacks.Recover("./snapshots"),  # State recovery before experiment starts, save on round end
])
```

This PR provides built-in support for: 
- [x] `MemoryProfiler`: To use, set `log_memory_usage: true` in the plan file. Both agg/col are supported.
- [x] `LambdaCallback`: for arbitrary functions
- [x] `MetricWriter`: Supports both Tensorboard and text logging. To use, set `write_logs: true` in the plan file. Both agg/col are supported.

Future:
- `ModelCheckpoint`: a new metric-configurable checkpointing logic
- `BackupAndRestore`: a new state recovery and saving technique, for all participant resiliency (coming soon)


openfl/callbacks/callback_list.py
```diff
@@ -0,0 +1,83 @@
...
+        self._add_default_callbacks(add_memory_profiler)
+
+        self.set_tensor_db(tensor_db)
+        self.set_params(params)
+
+    def set_params(self, params):
+        self.params = params
+        if params:
+            for callback in self.callbacks:
+                callback.set_params(params)
```
rahulga1, line 43: can there be params which are not applicable to all callbacks but to a specific callback?

openfl/callbacks/callback_list.py
```diff
@@ -0,0 +1,83 @@
...
+        self.tensor_db = tensor_db
+        if tensor_db:
+            for callback in self.callbacks:
+                callback.set_tensor_db(tensor_db)
+
+    def _add_default_callbacks(self, add_memory_profiler):
+        self._memory_profiler = None
+        for cb in self.callbacks:
+            if isinstance(cb, MemoryProfiler):
+                self._memory_profiler = cb
```
rahulga1, line 58: we can put break after this.

openfl/callbacks/callback_list.py
```diff
@@ -0,0 +1,83 @@
...
+
+        if add_memory_profiler and self._memory_profiler is None:
```
rahulga1, line 62: in what scenario self._memory_profiler can be not None?

openfl/component/aggregator/aggregator.py
```diff
@@ -76,10 +79,11 @@ def __init__(
         single_col_cert_common_name=None,
         compression_pipeline=None,
         db_store_rounds=1,
-        write_logs=False,
-        log_memory_usage=False,
-        log_metric_callback=None,
         initial_tensor_dict=None,
+        log_memory_usage=False,
```
rahulga1, line 82: are we using these variables write_logs and log_metric_callback?

openfl/component/collaborator/collaborator.py
```diff
@@ -123,8 +126,7 @@ def __init__(
         self.delta_updates = delta_updates
 
         self.client = client
-        # Flag can be enabled to get memory usage details for ubuntu system
-        self.log_memory_usage = log_memory_usage
+
         self.task_config = task_config
 
         self.logger = getLogger(__name__)
```
rahulga1: it will be good idea to maintain the same format across different component like either using self.logger or logger.

openfl/component/collaborator/collaborator.py
```diff
@@ -82,6 +82,8 @@ def __init__(
         compression_pipeline=None,
         db_store_rounds=1,
         log_memory_usage=False,
+        write_logs=False,
```
rahulga1, line 87: where is this being used?

rahulga1 (commented):
@payalcha @noopurintel please check if it impacts any of your test for memory related info?

openfl/callbacks/callback_list.py
MasterSkepticista: The intent is to define params during initialization that are visible to all callbacks in the list. For params that are limited to a specific callback, a user may choose to subclass and provide them during `__init__` of the respective class.

openfl/callbacks/callback_list.py
MasterSkepticista: If `log_memory_usage=False`, it leads to `add_memory_profiler=False`.

A user can either provide `MemoryProfiler` as part of the `callbacks` list (in which case the iterator checks and assigns the same if provided), or uses the boolean flag to add a new instance, when it is not part of the `callbacks` list.

openfl/component/aggregator/aggregator.py
MasterSkepticista: Both are for backwards compatibility. They are not used.

With addition of `TensorBoard` and `ModelCheckpoint` callback, the following arguments will be removed:
* `write_logs`
* `log_metric_callback`
* `{init,best,last}_state_path`
* `log_memory_usage` (once callbacks become plan-configurable)

@payalcha @noopurintel JFYR.

openfl/component/collaborator/collaborator.py
MasterSkepticista: Not used, same comment as previous.

openfl/component/collaborator/collaborator.py
MasterSkepticista: Good catch. I missed refactoring `self.logger` for collaborator. Will do

openfl/callbacks/callback_list.py
MasterSkepticista: Agree, for the code today. 

However, there will be more default callbacks coming in. Take this example of default `MemoryProfiler` and `MetricLogger`:

```python
for cb in self.callbacks:
  if isinstance(cb, MemoryProfiler):
    self._memory_profiler = cb
  elif isinstance(cb, MetricLogger):
    self._metric_logger = cb
```

A break exits the loop when we don't intend to. The downside of not having `break`, is that callbacks list will only assign the 'latest' seen callback instead of the 'first' seen callback.

openfl/callbacks/callback_list.py
```diff
@@ -0,0 +1,83 @@
...
+        callbacks: A list of `openfl.callbacks.Callback` instances.
+        add_memory_profiler: If True, adds a `MemoryProfiler` callback to the list.
+        tensor_db: Optional `TensorDB` instance of the respective participant.
+            If provided, callbacks can access TensorDB for various actions.
+        params: Additional parameters saved for use within the callbacks.
+    """
+
+    def __init__(
+        self,
+        callbacks: List[Callback],
```
theakshaypant: Nit: `typing.List` has been deprecated as of python 3.9. [Ref.](https://docs.python.org/3/library/typing.html#typing.List) 
`list` can be used directly.

openfl/callbacks/callback_list.py
```diff
@@ -0,0 +1,83 @@
...
+        add_memory_profiler=False,
+        tensor_db=None,
+        **params,
+    ):
+        super().__init__()
+        self.callbacks = _flatten(callbacks) if callbacks else []
```
theakshaypant, line 32: Could you give an example/explain how this flatten would be used given that `callbacks` arg has list typing?

openfl/component/aggregator/aggregator.py
theakshaypant, line 1: Would there be no use case for `on_experiment_begin`, `on_round_begin` or `on_experiment_end` on the aggregator side given that they are not invoked here?

theakshaypant (commented):
Had some questions related to the `callbacks` api.

openfl/component/aggregator/aggregator.py
MasterSkepticista: Good question. I don't know, yet.

Aggregator does not seem to have a "clear" point of round begin. It just starts, and collaborators start calling RPCs asynchronously. Same for experiment begin. An argument could be made to call `on_experiment_begin` during the initialization of aggregator. Lots of loose ends, I am looking for an elegant approach here. WDYT?

openfl/callbacks/callback_list.py
MasterSkepticista: Yay, 3.9 is minimum now

openfl/callbacks/callback_list.py
MasterSkepticista: It is to cover an edge case, where `callbacks` is a single `Callback` instead of a sequence of `Callback`s. `_flatten` will then create a singleton list.

I forgot to add that, it is now added in https://github.com/securefederatedai/openfl/pull/1195/commits/af0c40fe9e2a488daeddb4817b07299d8ec0f5d7

openfl/component/aggregator/aggregator.py
theakshaypant: For `on_experiment_begin`, I was thinking along the same lines, that it could be done when the aggregator is initialised. Similarly `on_experiment_end` could be denoted when the aggregator is being stopped (given that all tasks have been completed on collaborators) but one could argue the semantics on when the experiment actually begins or ends. 
Only have a non-elegant way to verify `on_round_begin` for aggregators where a flag can be used to check the first time `get_tasks` is called in each round.

MasterSkepticista: `on_experiment_{begin,end}` proposals are reasonable - during `__init__` and `stop(...)`.

Lets think more for `on_round_begin`. Can it become elegant if we take the liberty of minor refactoring?

theakshaypant: Similar to `Aggregator._end_of_round_check`, we can introduce a `_beginning_of_round_check` which is called at the start of `get_tasks` unless it `time_to_quit`. 

We can log (read "store") when a collaborator calls `get_tasks` for each round. If the round number is not logged (again, read "stored"), we know this is the first `get_tasks` for the round and hence the beginning of the round. 

The storage mentioned can be done in an in-memory dict similar to other attributes (such as `collaborator_tasks_results`) of the `Aggregator`.

psfoley:
@MasterSkepticista Are the 3 remaining unresolved items in the checklist (`TensorBoard`, `ModelCheckpoint`, `BackupAndRestore`) intended to be addressed before merge?

openfl/component/aggregator/aggregator.py
MasterSkepticista: Using this for now: https://github.com/securefederatedai/openfl/pull/1195/commits/1cffb9d4d4c39575caf69140bbe85878950ea54d

Edit: Please review once too.

MasterSkepticista:
@psfoley `ModelCheckpoint` and state backup/restore are big-ticket items. Tensorboard support is added via `MetricWriter` callback. Updated PR description for usage methods.

This PR does not introduce any new plan arguments, for backwards compatibility. Callbacks will be exposed to users in a follow-up PR. So, ready to merge.

openfl/component/aggregator/aggregator.py
```diff
@@ -1009,14 +994,14 @@ def _end_of_round_check(self):
         # resetting collaborators_done for next round
         self.collaborators_done = []
 
+        # https://github.com/securefederatedai/openfl/pull/1195#discussion_r1879479537
+        self.callbacks.on_round_begin(self.round_number)
```
theakshaypant: This would execute the callbacks an additional time at the completion of all the rounds as well, right? We can add a check against `self.rounds_to_train` to maybe avoid that or add it under the `self._time_to_quit` check just below this. 
Can't say whether running this extra callback(List) without a round starting would have any considerable impact on what is intended to be done by the callbacks.

openfl/callbacks/callback.py
```diff
@@ -0,0 +1,56 @@
...
+        self.params = None
+        self.tensor_db = None
+
+    def set_params(self, params):
+        self.params = params
+
+    def set_tensor_db(self, tensor_db):
+        self.tensor_db = tensor_db
+
+    def on_round_begin(self, round_num: int, logs=None):
```
theakshaypant, line 34: Based on my read through, I see that the `logs` named-arg is only used by `MetricWriter`. 
For extensibility, would it make more sense to have a `**kwargs` instead?
We would still need to consider how these "dynamic" named args are sent by the aggregator/collaborator but this could be maybe thought of in subsequent/future changes to the `callbacks` API.

theakshaypant (commented):
Have a couple of questions which might be out of scope for this PR but still wanted to highlight them.

openfl/component/aggregator/aggregator.py
MasterSkepticista: Thanks for pointing this out. Addressed in https://github.com/securefederatedai/openfl/pull/1195/commits/d01ec5e898ba9ddf7cbfe269f9910c6025ebb70e

openfl/callbacks/callback.py
MasterSkepticista: > For extensibility, would it make more sense to have a **kwargs instead?

Generally speaking, `**kwargs` should be avoided. To pass information that specific callbacks would need, we have `params`  that are declared at the time of instantiating `callbacks` list. This makes it clear what attributes `callbacks` will have access to, internally. Dynamic setting makes it harder to inspect and test code.

https://github.com/securefederatedai/openfl/blob/38f7c30c57aa5d92db1ecf0c35a22be6733ed30d/openfl/callbacks/callback_list.py#L18-L20

Currently, `logs` are the default because most `callbacks` generally trigger actions based on metrics - say writing `logs` to file, trigger saving a checkpoint if `logs["loss"] < 0.2`, write `logs` to tensorboard etc. 

That said, I agree that we can expand this argument list further - if more callbacks benefit from them.

psfoley (approved):

PR #1206, opened 2024-12-10
Add Fedcurv example using TaskRunner API

ynonflumintel:


openfl/federated/task/fl_model.py
```diff
@@ -75,10 +76,17 @@ def __init__(self, build_model, optimizer=None, loss_fn=None, **kwargs):
             self.runner.validate = lambda *args, **kwargs: build_model.validate(
                 self.runner, *args, **kwargs
             )
+        # TODO: not used anywhere - should it be removed?
```
teoparvanov: It's not very common indeed, but I saw one instance [here](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch_llm_horovod/src/InHorovodLLMTrainer.py#L200).

openfl/federated/task/fl_model.py
```diff
@@ -75,10 +76,17 @@ def __init__(self, build_model, optimizer=None, loss_fn=None, **kwargs):
...
         if hasattr(self.model, "train_epoch"):
             self.runner.train_epoch = lambda *args, **kwargs: build_model.train_epoch(
                 self.runner, *args, **kwargs
             )
+
+        # Used to hook the training function when debugging locally
+        if hasattr(self.model, "train_"):
+            self.runner.train_ = lambda *args, **kwargs: build_model.train_(
+                self.runner, *args, **kwargs
+            )
```
teoparvanov, line 90: is it just for local debugging?

teoparvanov (approved):
LGTM!

openfl/federated/task/fl_model.py
```diff
@@ -75,10 +76,17 @@ def __init__(self, build_model, optimizer=None, loss_fn=None, **kwargs):
             self.runner.validate = lambda *args, **kwargs: build_model.validate(
                 self.runner, *args, **kwargs
             )
+        # TODO: not used anywhere - should it be removed?
         if hasattr(self.model, "train_epoch"):
             self.runner.train_epoch = lambda *args, **kwargs: build_model.train_epoch(
                 self.runner, *args, **kwargs
             )
+
+        # Used to hook the training function when debugging locally
```
kminhta, line 86: Is this change related to the FedCurv task runner example?

openfl/native/native.py
```diff
@@ -215,7 +215,7 @@ def init(
     Args:
         workspace_template (str): The template that should be used as the
             basis for the experiment.  Defaults to 'default'.
-            Other options include are any of the template names
+            Other options include any of the template names
```
kminhta: I think we can omit this from this PR. Also because native API will be deprecated and removed eventually anyway

kminhta (changes requested):
Thanks @ynonflumintel ! Looks good overall

**Request:** Can you include a README with commands to run this workspace as well as some information about FedCurv (especially highlighting the diff from the base `torch_cnn_histology`? Good to cite it as well: https://arxiv.org/pdf/1910.07796

It does not need to be a super in depth doc, but should provide users the ability to run it end to end

Also, a couple minor comments/questions below

openfl/federated/task/fl_model.py
ynonflumintel: I needed it to debug fedcurv, I don't mind omitting it if you don't think it should be included in this PR

openfl/federated/task/fl_model.py
ynonflumintel: When using a distributed deployment you can change the train function directly , so yes

openfl/federated/task/fl_model.py
kminhta: Ohh, I think it's okay to keep then. In fact, probably an overdue update since most of new runners (definitely the case for torch) use `train_` instead of `train_epoch`

kminhta (approved):
LGTM! Thanks @ynonflumintel

PR #1207, opened 2024-12-11
Refactor PR - Remove global variable and get_writer() function

payalcha:
This PR is just to refactor the code in different utils.py which is using SummaryWriter.
1. Removal of Global variable usage.
2. removal of get_writer function. As it is not used anywhere else except write_metric.
3. Directly initializing the writer variable.

### DIFF
--- a/openfl-tutorials/deprecated/native_api/Federated_Pytorch_MNIST_Tutorial.ipynb
+++ b/openfl-tutorials/deprecated/native_api/Federated_Pytorch_MNIST_Tutorial.ipynb
@@ -135,10 +135,15 @@
    "source": [
     "from torch.utils.tensorboard import SummaryWriter\n",
     "\n",
-    "writer = SummaryWriter('./logs/cnn_mnist', flush_secs=5)\n",
+    "writer = None\n",
     "\n",
+    "def get_writer():\n",
+    "    global writer\n",
+    "    if not writer:\n",
+    "        writer = SummaryWriter('./logs/cnn_mnist', flush_secs=5)\n",
     "\n",
     "def write_metric(node_name, task_name, metric_name, metric, round_number):\n",
+    "    get_writer()\n",
     "    writer.add_scalar(\"{}/{}/{}\".format(node_name, task_name, metric_name),\n",
     "                      metric, round_number)"
    ]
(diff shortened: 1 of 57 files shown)


openfl-tutorials/deprecated/native_api/Federated_Pytorch_MNIST_Tutorial.ipynb
MasterSkepticista, line 1: This is for a deprecated API, need not update this.

openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/101_torch_cnn_mnist/src/utils.py
```diff
@@ -4,17 +4,7 @@
...
-    global writer
-    if not writer:
-        writer = SummaryWriter('./logs/cnn_mnist', flush_secs=5)
-
-
 def write_metric(node_name, task_name, metric_name, metric, round_number):
     """Write metric callback."""
-    get_writer()
+    writer = SummaryWriter('./logs/cnn_mnist', flush_secs=5)
     writer.add_scalar(f'{node_name}/{task_name}/{metric_name}', metric, round_number)
```
MasterSkepticista, line 10: For all taskrunner example updates in this PR: this function will no longer be required, since a `TensorBoard` callback will internally maintain a writer object and necessary functions to write scalars.

In summary, this (including imports) can be removed too.

openfl/utilities/logs.py
```diff
@@ -13,33 +13,18 @@
...
     Args:
         node_name (str): The name of the node.
         task_name (str): The name of the task.
         metric_name (str): The name of the metric.
         metric (float): The value of the metric.
         round_number (int): The current round number.
     """
-    get_writer()
+    writer = SummaryWriter("./logs/tensorboard", flush_secs=5)
     writer.add_scalar(f"{node_name}/{task_name}/{metric_name}", metric, round_number)
```
MasterSkepticista, line 28: All this can be removed as well.

P.S: Creating a writer inside `write_metric` will create a new file handle on each call, which is slow and wasteful. What I meant was to create a function that _creates_ a writer, passes it to a function and returns it. Example:

```python
def get_metric_writer(log_dir: str = "./logs"):
  writer = SummaryWriter(log_dir)

  def _write_scalar(node_name, task_name, metric_name, metric, round_number):
    writer.add_scalar(f"{node_name}/{task_name}/{metric_name}", metric, round_number)

  return _write_scalar
```

Anyways, this also needs to be removed because it will be taken care of, in TensorBoard callback

MasterSkepticista (changes requested):

MasterSkepticista:
Closing, resolved in #1195.

PR #1213, opened 2024-12-16
Bump Python version to 3.10 - 3.12

tanwarsh:
Support for Python version from 3.10 to 3.12.


openfl-tutorials/experimental/workflow/404_Keras_MNIST_with_FedProx.ipynb
```diff
@@ -51,7 +51,7 @@
    "id": "4dbb89b6",
    "metadata": {},
    "source": [
-    "First we start by installing the necessary dependencies for the workflow interface. Note that this tuorial uses Keras 3, make sure you use python 3.9 or higher."
+    "First we start by installing the necessary dependencies for the workflow interface. Note that this tuorial uses Keras 3, make sure you use python 3.10 or higher."
```
MasterSkepticista, line 54: OpenFL is assumed to be installed at this point, which would not have happened without python requirements being met. Consider removing mention of python altogether.

rahulga1 (approved):

openfl-tutorials/experimental/workflow/404_Keras_MNIST_with_FedProx.ipynb
tanwarsh: The steps mentioned below this comment helps user install Tensorflow, openfl and other required libs. If we are assuming openfl and required lib are installed then we should remove the steps mentioned below as well.

MasterSkepticista: `pip` verifies python version requirement for every package that is installed (which is why we mention in `setup.py`). That is an implicit check. If required anyway, it is better to point the user to our official documentation - [Refer to the installation guide for OpenFL](https://openfl.readthedocs.io/en/latest/get_started/installation.html).

My broader point is to increase maintainability of the framework as part of the PR. 
It is one less line to maintain when we bump minimum python to 3.11 next year.

psfoley (approved):

psfoley:
@tanwarsh The changes look good. Please resolve the merge conflicts. This will need to be brought into `v1.7.x` for consistency with the latest documentation that states support for python 3.10-3.12

Issue #1075, opened 2024-10-02
[Windows] `torchvision=0.18.1` incompatible with `NumPy 2.x` leading to error initializing torch workspaces


kminhta:
A while back NumPy release v2.x which resulted in issues when using packages compiled for NumPy 1.x
One area where `openfl` was specifically affected was in the torch-workspaces [Ref Issue #999]

While updating the workspaces to `torch==2.3.1` and `torchvision==0.18.1` seemed to work on ubuntu, it seems that Windows `torchvision==0.18.1` is still incompatible with `NumPy v2.x` resulting in errors initializing torch workspaces on Windows

We should look into updating `torch` and `torchvision` to later versions to compatible across Ubuntu and Windows

```
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.1.1 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "<stdin>", line 1, in <module>
  File "C:\Documents\openfl\venv\lib\site-packages\torchvision\__init__.py", line 6, in <module>
    from torchvision import _meta_registrations, datasets, io, models, ops, transforms, utils
  File "C:\Documents\openfl\venv\lib\site-packages\torchvision\models\__init__.py", line 2, in <module>
    from .convnext import *
  File "C:\Documents\openfl\venv\lib\site-packages\torchvision\models\convnext.py", line 8, in <module>
    from ..ops.misc import Conv2dNormActivation, Permute
  File "C:\Documents\openfl\venv\lib\site-packages\torchvision\ops\__init__.py", line 23, in <module>
    from .poolers import MultiScaleRoIAlign
  File "C:\Documents\openfl\venv\lib\site-packages\torchvision\ops\poolers.py", line 10, in <module>
    from .roi_align import roi_align
  File "C:\Documents\openfl\venv\lib\site-packages\torchvision\ops\roi_align.py", line 4, in <module>
    import torch._dynamo
  File "C:\Documents\openfl\venv\lib\site-packages\torch\_dynamo\__init__.py", line 64, in <module>
    torch.manual_seed = disable(torch.manual_seed)
[...]

PR #1216, opened 2024-12-16
PyTorch and Torchvision version update

tanwarsh:
Update PyTorch to version 2.4.1 and Torchvision to version 0.19.1.

Fix for https://github.com/securefederatedai/openfl/issues/1075.


tanwarsh:
Pipelines are facing Issue https://github.com/pytorch/pytorch/issues/140914.

kminhta:
> Pipelines are facing Issue [pytorch/pytorch#140914](https://github.com/pytorch/pytorch/issues/140914).

Thanks for raising the issue. Can you look into the possibility of adding the monkey patch to `openfl.interface.cli`?
I didn't do any extensive testing on the e2e functionality, but it got past the `TypeError`

https://github.com/pytorch/pytorch/issues/140914#issuecomment-2558677512

If it works, we can maybe implement it while we wait for a more robust upstream solution, or we can investigate where/how `pip._vendor.typing_extensions` is being forcibly called for a more robust solution in our code base

tanwarsh:
> > Pipelines are facing Issue [pytorch/pytorch#140914](https://github.com/pytorch/pytorch/issues/140914).
> 
> Thanks for raising the issue. Can you look into the possibility of adding the monkey patch to `openfl.interface.cli`? I didn't do any extensive testing on the e2e functionality, but it got past the `TypeError`
> 
> [pytorch/pytorch#140914 (comment)](https://github.com/pytorch/pytorch/issues/140914#issuecomment-2558677512)
> 
> If it works, we can maybe implement it while we wait for a more robust upstream solution, or we can investigate where/how `pip._vendor.typing_extensions` is being forcibly called for a more robust solution in our code base

Thank you @kta-intel. Workaround patch worked.

MasterSkepticista:
Why are we insisting on pinning torch and torchvision, that too to latest versions? Is there anything we lose by using stable, yet recent torch versions?

I am not in favor of instrumenting and applying hotfixes that become a future maintenance burden, especially if `torch~=2.3.0` works without any bugs mentioned in this PR.

WDYT? @psfoley @kta-intel @teoparvanov @theakshaypant

tanwarsh:
> Why are we insisting on pinning torch and torchvision, that too to latest versions? Is there anything we lose by using stable, yet recent torch versions?
> 
> I am not in favor of instrumenting and applying hotfixes that become a future maintenance burden, especially if `torch~=2.3.0` works without any bugs mentioned in this PR.
> 
> WDYT? @psfoley @kta-intel @teoparvanov @theakshaypant


with `torch~=2.3.0` we are facing issue in windows. Please refer this Issue https://github.com/securefederatedai/openfl/issues/1075

noopurintel (approved):

MasterSkepticista:
> > Why are we insisting on pinning torch and torchvision, that too to latest versions? Is there anything we lose by using stable, yet recent torch versions?
> > I am not in favor of instrumenting and applying hotfixes that become a future maintenance burden, especially if `torch~=2.3.0` works without any bugs mentioned in this PR.
> > WDYT? @psfoley @kta-intel @teoparvanov @theakshaypant
> 
> with `torch~=2.3.0` we are facing issue in windows. Please refer this Issue #1075

Did we try 2.4.x?

kminhta:
> > > Why are we insisting on pinning torch and torchvision, that too to latest versions? Is there anything we lose by using stable, yet recent torch versions?
> > > I am not in favor of instrumenting and applying hotfixes that become a future maintenance burden, especially if `torch~=2.3.0` works without any bugs mentioned in this PR.
> > > WDYT? @psfoley @kta-intel @teoparvanov @theakshaypant
> > 
> > 
> > with `torch~=2.3.0` we are facing issue in windows. Please refer this Issue #1075
> 
> Did we try 2.4.x?

I believe 2.4.x was initially triggering the `AssertionError: /python3.11/distutils/core.py`, but it seems that @tanwarsh managed to get past that error. Can you apply your resolution for the `distutil` issue and then check if this type error is also present out of the box? If not, I think we can safely stick to 2.4.x for now.

Though I think it is a practical solution, @MasterSkepticista brings up a good point that using `sys.module` to force consistent modules may be not a good long-term solution. It seems like the issue is directly related `torch` (under `torch==2.3.1`, `pip.__vendor.typing_extensions` does **not** get imported when checking `ParamSpec`, but it does in `torch===2.5.1`), so we may want to revive that thread and help the investigation since there doesn't seem to be a ton of action in it

tanwarsh:
> > > > Why are we insisting on pinning torch and torchvision, that too to latest versions? Is there anything we lose by using stable, yet recent torch versions?
> > > > I am not in favor of instrumenting and applying hotfixes that become a future maintenance burden, especially if `torch~=2.3.0` works without any bugs mentioned in this PR.
> > > > WDYT? @psfoley @kta-intel @teoparvanov @theakshaypant
> > > 
> > > 
> > > with `torch~=2.3.0` we are facing issue in windows. Please refer this Issue #1075
> > 
> > 
> > Did we try 2.4.x?
> 
> I believe 2.4.x was initially triggering the `AssertionError: /python3.11/distutils/core.py`, but it seems that @tanwarsh managed to get past that error. Can you apply your resolution for the `distutil` issue and then check if this type error is also present out of the box? If not, I think we can safely stick to 2.4.x for now.
> 
> Though I think it is a practical solution, @MasterSkepticista brings up a good point that using `sys.module` to force consistent modules may be not a good long-term solution. It seems like the issue is directly related `torch` (under `torch==2.3.1`, `pip.__vendor.typing_extensions` does **not** get imported when checking `ParamSpec`, but it does in `torch===2.5.1`), so we may want to revive that thread and help the investigation since there doesn't seem to be a ton of action in it



Thank you @kta-intel and @MasterSkepticista for your inputs. I will downgrade the version to 2.4.x and check both issues. If neither issue or the type error is not encountered, I will proceed with 2.4.x.

tanwarsh:
@MasterSkepticista @kta-intel, We are using version 2.4.1 now without patch. Please review again.

openfl-workspace/torch_cnn_mnist_straggler_check/requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy>=1.22.2 # not directly required, pinned by Snyk to avoid a vulnerability
 rsa>=4.7 # not directly required, pinned by Snyk to avoid a vulnerability
-setuptools>=65.5.1 # not directly required, pinned by Snyk to avoid a vulnerability
+setuptools==59.6.0
```
kminhta: Seems like pinning to `59.6.0` will trigger Snyk vulnerability concerns. Will this cause us issues down the line?

kminhta:
> @MasterSkepticista @kta-intel, We are using version 2.4.1 now without patch. Please review again.

Thanks for checking into that @tanwarsh. Glad 2.4.1 is not raising the TypeError issue. My open now is if downgrading `setuptools` is the path we want to go. Even if we're able to justify it for a release, if we're stuck at `59.6.0` going forward with torch workloads, that could be unsustainable long term

tanwarsh:
> > @MasterSkepticista @kta-intel, We are using version 2.4.1 now without patch. Please review again.
> 
> Thanks for checking into that @tanwarsh. Glad 2.4.1 is not raising the TypeError issue. My open now is if downgrading `setuptools` is the path we want to go. Even if we're able to justify it for a release, if we're stuck at `59.6.0` going forward with torch workloads, that could be unsustainable long term

@kta-intel , I tried few things, trying different version and patches suggested in few issues logged in setuptools github repo but none of them are working. Looks like we are stuck with this solution.

openfl/interface/workspace.py
```diff
@@ -3,9 +3,11 @@
 
 
 """Workspace module."""
+import os
+
+os.environ['SETUPTOOLS_USE_DISTUTILS'] = 'stdlib'
```
tanwarsh: This patch worked as well with the latest version. Maybe these changes can be made somewhere else. we can add this only for pyTorch and related workspaces.

openfl-workspace/torch_cnn_mnist_straggler_check/requirements.txt
tanwarsh: resolved as we are using latest version now.

tanwarsh:
@psfoley @kta-intel @teoparvanov @rahulga1, 
There are two solutions for the issue with Setuptools:

1. Pin Setuptools to version 59.6.0, which has a vulnerability issue. https://github.com/pytorch/pytorch/issues/143288.
2. Apply a patch to use the latest version of Setuptools (current changes).
 
Please review this PR and share your thoughts so that we can move forward.

kminhta:
> @psfoley @kta-intel @teoparvanov @rahulga1, There are two solutions for the issue with Setuptools:
> 
> 1. Pin Setuptools to version 59.6.0, which has a vulnerability issue. [EXCEPTION : /python3.11/distutils/core.py pytorch/pytorch#143288](https://github.com/pytorch/pytorch/issues/143288).
> 2. Apply a patch to use the latest version of Setuptools (current changes).
> 
> Please review this PR and share your thoughts so that we can move forward.

I am okay with the patch for now, though we should keep it in the backlog to look for a more robust solution
Btw, do you know if this will break older versions (i.e. `torch=2.3`)?

tanwarsh:
> > @psfoley @kta-intel @teoparvanov @rahulga1, There are two solutions for the issue with Setuptools:
> > 
> > 1. Pin Setuptools to version 59.6.0, which has a vulnerability issue. [EXCEPTION : /python3.11/distutils/core.py pytorch/pytorch#143288](https://github.com/pytorch/pytorch/issues/143288).
> > 2. Apply a patch to use the latest version of Setuptools (current changes).
> > 
> > Please review this PR and share your thoughts so that we can move forward.
> 
> I am okay with the patch for now, though we should keep it in the backlog to look for a more robust solution Btw, do you know if this will break older versions (i.e. `torch=2.3`)?

This issue is not seen in torch==2.3.1, but the PR is to update torch from 2.3.1 because we were facing this issue https://github.com/securefederatedai/openfl/issues/1075

kminhta:
> > > @psfoley @kta-intel @teoparvanov @rahulga1, There are two solutions for the issue with Setuptools:
> > > 
> > > 1. Pin Setuptools to version 59.6.0, which has a vulnerability issue. [EXCEPTION : /python3.11/distutils/core.py pytorch/pytorch#143288](https://github.com/pytorch/pytorch/issues/143288).
> > > 2. Apply a patch to use the latest version of Setuptools (current changes).
> > > 
> > > Please review this PR and share your thoughts so that we can move forward.
> > 
> > 
> > I am okay with the patch for now, though we should keep it in the backlog to look for a more robust solution Btw, do you know if this will break older versions (i.e. `torch=2.3`)?
> 
> This issue is not seen in torch==2.3.1, but the PR is to update torch from 2.3.1 because we were facing this issue #1075

I was asking if patching `os.environ["SETUPTOOLS_USE_DISTUTILS"] = "stdlib"` would affect `torch==2.3.1`. If so, we should make note of it in a "known issues" section that a user would have to patch to local in their own application if they cannot upgrade, since we cannot expect a user to modify framework-level code, such as `openfl/federated/__init__.py`. But it doesn't, then it is no problem

tanwarsh:
> > > > @psfoley @kta-intel @teoparvanov @rahulga1, There are two solutions for the issue with Setuptools:
> > > > 
> > > > 1. Pin Setuptools to version 59.6.0, which has a vulnerability issue. [EXCEPTION : /python3.11/distutils/core.py pytorch/pytorch#143288](https://github.com/pytorch/pytorch/issues/143288).
> > > > 2. Apply a patch to use the latest version of Setuptools (current changes).
> > > > 
> > > > Please review this PR and share your thoughts so that we can move forward.
> > > 
> > > 
> > > I am okay with the patch for now, though we should keep it in the backlog to look for a more robust solution Btw, do you know if this will break older versions (i.e. `torch=2.3`)?
> > 
> > 
> > This issue is not seen in torch==2.3.1, but the PR is to update torch from 2.3.1 because we were facing this issue #1075
> 
> I was asking if patching `os.environ["SETUPTOOLS_USE_DISTUTILS"] = "stdlib"` would affect `torch==2.3.1`. If so, we should make note of it in a "known issues" section that a user would have to patch to local in their own application if they cannot upgrade, since we cannot expect a user to modify framework-level code, such as `openfl/federated/__init__.py`. But it doesn't, then it is no problem

@kta-intel , sorry for the misunderstanding. I tested this in local. This patch works with version 2.3.1 as well. User will not have to modify anything.

teoparvanov (approved):
Hey @tanwarsh, thanks for upgrading the PyTorch workspaces to 2.4.1! I see this PR as an intermediate step on our "journey" to 2.5.

PR #1218, opened 2024-12-16
E2E automation: Federation run with dockerized workspace 

noopurintel:
**Main changes:**

1. Added automation for federation run using dockerized workspace.
2. Separate fixtures for dockerized workspace and bare-metal approaches in `common_fixtures.py`
3. Single test file `task_runner_tests.py` to include both approaches.
4. Removed (federation inside) docker based workflow and test file.
5. Added GitHub composite action files under `tr_pre_test_run` and `tr_post_test_run` and moved the relevant steps from the workflows files to these action files, as applicable.

**Triggered jobs with this branch:**

1. Task Runner E2E (bare metal) - https://github.com/noopurintel/openfl/actions/runs/12392299190
2. Task Runner E2E (dockerized) - https://github.com/noopurintel/openfl/actions/runs/12392301266


theakshaypant (commented):
Based on my understanding, in `task_runner_docker`, we are mounting the created certs/workspace to aggregator and collaborator containers. In my opinion, we need (should) not run these tests as this is not a "proper" way of running a containerised federation. Just having `task_runner_dockerized_ws`, where we are running the federation using the `fx workspace dockerize` command, should be sufficient. 
Thoughts?

theakshaypant (approved):
@noopurintel Per our discussion, we can (should) remove `task_runner_docker` related workflow(s) in a future PR. 
Should fix the linting here. 
LGTM otherwise. 😄

ishaileshpant (approved):
LGTM

.github/actions/tr_pre_test_run/action.yml
MasterSkepticista, line 1: If you are calling `fx workspace dockerize` anywhere in your tests, this entire step may be redundant. Base images are built during dockerize.

.github/actions/tr_post_test_run/action.yml
```diff
@@ -0,0 +1,43 @@
+---
+#---------------------------------------------------------------------------
+# Composite Action to run post-test functions for task runner end to end tests
+# Includes:
+# - Print test summary
+# - Create tar file
+# - Upload artifacts
+#
+# Authors - Noopur, Payal Chaurasiya
```
MasterSkepticista: Not necessary. Please remove author names from this file and previously committed files. We use git history and optionally `TODO(noopurintel)` if you intend to add notes.

Also the top-level "Includes" section is self-explanatory, not required. One line will suffice:

```yaml
# Composite Action to run post-test functions for task runner end to end tests.

name: 'Post-Test Functions'
...
```

.github/workflows/task_runner_basic_e2e.yml
MasterSkepticista, line 1: Rename: `basic_e2e.yml -> native_e2e.yml` (I think that was the intention?)

.github/workflows/task_runner_dockerized_ws_e2e.yml
```diff
@@ -0,0 +1,194 @@
+---
+#---------------------------------------------------------------------------
+# Workflow to run Task Runner end to end tests using dockerized workspace
+# Authors - Noopur, Payal Chaurasiya
+#---------------------------------------------------------------------------
+name: Task_Runner_E2E_via_Dockerized_Workspace  # Please do not modify the name as it is used in the composite action
```
MasterSkepticista: Necessary to call it `Dockerized_Workspace`? TaskRunner obviates that it will be a workspace.

tests/end_to_end/models/aggregator.py
```diff
@@ -49,24 +50,27 @@ def generate_sign_request(self):
         except Exception as e:
             raise ex.CSRGenerationException(f"Failed to generate sign request for {self.name}: {e}")
 
-    def start(self, res_file):
+    def start(self, res_file, run_for_dockerized_ws=False):
```
MasterSkepticista: `with_docker` is cleaner?

tests/end_to_end/utils/summary_helper.py
```diff
@@ -41,13 +44,15 @@ def get_aggregated_accuracy(agg_log_file):
...
-                    # If the line does not contain closing bracket "}", then concatenate the next line
-                    reqd_line = line if "}" in line else line + next(f).strip()
-                    agg_accuracy = eval(reqd_line.split("METRIC")[1].strip('"'))["metric_value"]
+        with open(agg_log_file, "r") as file:
+            lines = file.readlines()
+            for line in lines:
+                match = re.search(
+                    r"'metric_origin': 'aggregator', 'task_name': 'aggregated_model_validation', 'metric_name': 'accuracy', 'metric_value':\s*([\d.]+)",
+                    line,
+                )
```
MasterSkepticista: Regex searching for metrics is not a good approach. This will break immediately once logging style changes (FYI - that may happen soon). Better way is to write metrics to a CSV/JSON and read structured data as-is.

@payalcha can provide context.

MasterSkepticista (changes requested):

.github/actions/tr_pre_test_run/action.yml
noopurintel: For fx command to work, we need these steps.

.github/workflows/task_runner_basic_e2e.yml
noopurintel: Not really. Here `basic` simply denotes the bare metal approach.

.github/actions/tr_post_test_run/action.yml
noopurintel: Done. I have removed the author names from all the files which are part of end_to_end module.

.github/workflows/task_runner_dockerized_ws_e2e.yml
noopurintel: Renamed it to `Task_Runner_Dockerized_E2E`

tests/end_to_end/models/aggregator.py
noopurintel: Earlier, we had docker (federation inside container) approach as well, thus suffixed this variable with `dockerized_ws` to avoid confusion. Now that it is removed, a simple name should do.

Renamed to `with_docker` at all relevant places.

tests/end_to_end/utils/summary_helper.py
noopurintel: I discussed it with Payal. If the callback is not implemented yet, we will need to change logic in `aggregator.py` to redirect the metrics to a file. 
If that is fine, I can modify the file, else will wait for your PR to get merged.

noopurintel: We discussed about this change. I will wait for callback implementation for metrics, and then use it (instead of current method of parsing through the file)

noopurintel:
@theakshaypant : linting issues solved + task_runner_docker related workflows removed in this PR.

PR #1221, opened 2024-12-18
Use localhost for the aggregator in the quick-start tutorial

teoparvanov:
To simplify the local OpenFL run, we can deploy the aggregator on `localhost`, instead of a dynamically generated FQDN (which could lead to proxy or certificate issues, among others). Running the federation in a distributed setting requires a different FL plan setup anyway, which seems out of scope for a "quick start" tutorial.

The proposed change aims to make the initial developer experience with OpenFL smoother.


yuliasherman (approved):
LGTM

kminhta (approved):

PR #1226, opened 2024-12-23
Implement task_group filtering for FedEval

ishaileshpant:
- implement a new task_group filtering decorator 'with_selected_task_group' in Assigner class
- update all the sub-classes that use task_groups to use the decorator
- update fedeval sample workspace to use default assigner, tasks and aggregator
- use of federated-evaluation/aggregator.yaml for FedEval specific workspace example to use round_number as 1
- removed assigner and tasks yaml from defaults/federated-evaluation, superseded by default assigner/tasks
- added additional checks for assigner sub-classes that might not have task_groups
- Addressing review comments
- Updated existing test cases for Assigner sub-classes
- Remove hard-coded setting in assigner for torch_cnn_mnist ws, refer to default as in other Workspaces
- Use aggregator supplied --task_group to override the assinger selected_task_group
- update existing test cases of aggregator cli
- add test cases for the decorator
- rebased 25-Jan.1
Signed-off-by: Shailesh Pant <shailesh.pant@intel.com>

- [x] Testcases for task_group_filtering decorator
- [x] Testing for all existing workspaces
- [x] we shouldn't be saving any new models if running in evaluation mode.
- [x] Change all workspaces to use new Assigner [Not needed as defaults ensure all workspace, by default use new assigner]

### DIFF
--- a/openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
+++ b/openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
@@ -32,10 +32,12 @@ network :
   defaults : plan/defaults/network.yaml
 
 assigner :
-  defaults : plan/defaults/federated-evaluation/assigner.yaml
-  
+  defaults : plan/defaults/assigner.yaml
+  settings :
+    selected_task_group : evaluation
+
 tasks :
-  defaults : plan/defaults/federated-evaluation/tasks_torch.yaml
+  defaults : plan/defaults/tasks_torch.yaml
 
 compression_pipeline :
   defaults : plan/defaults/compression_pipeline.yaml
--- a/openfl-workspace/workspace/plan/defaults/assigner.yaml
+++ b/openfl-workspace/workspace/plan/defaults/assigner.yaml
@@ -7,3 +7,7 @@ settings :
         - aggregated_model_validation
         - train
         - locally_tuned_model_validation
+    - name       : evaluation
+      percentage : 0
+      tasks      :
+        - aggregated_model_validation
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -130,15 +130,25 @@ class Aggregator:
         self.straggler_handling_policy = (
             straggler_handling_policy or CutoffTimeBasedStragglerHandling()
         )
-        self._end_of_round_check_done = [False] * rounds_to_train
-        self.stragglers = []
 
         self.rounds_to_train = rounds_to_train
+        if self.task_group == "evaluation":
+            self.rounds_to_train = 1
+            logger.info(
+                f"task_group is {self.task_group}, setting rounds_to_train = {self.rounds_to_train}"
+            )
+
+        self._end_of_round_check_done = [False] * rounds_to_train
+        self.stragglers = []
 
         # if the collaborator requests a delta, this value is set to true
         self.authorized_cols = authorized_cols
         self.uuid = aggregator_uuid
         self.federation_uuid = federation_uuid
+        # # override the assigner selected_task_group
+        # # FIXME check the case of CustomAssigner as base class Assigner is redefined
+        # # and doesn't have selected_task_group as attribute
+        # assigner.selected_task_group = task_group
         self.assigner = assigner
         self.quit_job_sent_to = []
 
--- a/openfl/component/assigner/assigner.py
+++ b/openfl/component/assigner/assigner.py
@@ -4,6 +4,11 @@
 
 """Assigner module."""
 
+import logging
+from functools import wraps
+
+logger = logging.getLogger(__name__)
+
 
 class Assigner:
     r"""
@@ -35,18 +40,27 @@ class Assigner:
         \* - ``tasks`` argument is taken from ``tasks`` section of FL plan YAML file.
     """
 
-    def __init__(self, tasks, authorized_cols, rounds_to_train, **kwargs):
+    def __init__(
+        self,
+        tasks,
+        authorized_cols,
+        rounds_to_train,
+        selected_task_group: str = None,
+        **kwargs,
+    ):
         """Initializes the Assigner.
 
         Args:
             tasks (list of object): List of tasks to assign.
             authorized_cols (list of str): Collaborators.
             rounds_to_train (int): Number of training rounds.
+            selected_task_group (str, optional): Selected task_group.
             **kwargs: Additional keyword arguments.
         """
         self.tasks = tasks
         self.authorized_cols = authorized_cols
         self.rounds = rounds_to_train
+        self.selected_task_group = selected_task_group
         self.all_tasks_in_groups = []
 
         self.task_group_collaborators = {}
@@ -93,3 +107,50 @@ class Assigner:
         if "aggregation_type" not in self.tasks[task_name]:
             return None
         return self.tasks[task_name]["aggregation_type"]
+
+    @classmethod
+    def with_selected_task_group(cls, func):
+        """Decorator to filter task groups based on selected_task_group.
+
+        This decorator should be applied to define_task_assignments() method
+        in Assigner subclasses to handle task_group filtering.
+        """
+
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            # Check if selection of task_group is applicable
+            if hasattr(self, "selected_task_group") and self.selected_task_group is not None:
+                # Verify task_groups exists before attempting filtering
+                if not hasattr(self, "task_groups"):
+                    logger.warning(
+                        "Task group specified for selection but no task_groups found. "
+                        "Skipping filtering. This might be intentional for custom assigners."
+                    )
+                    return func(self, *args, **kwargs)
+
+                assert self.task_groups, "No task_groups defined in assigner."
+
+                # Perform the filtering
+                selected_task_groups = [
+                    group for group in self.task_groups if group["name"] == self.selected_task_group
+                ]
+
+                assert len(selected_task_groups) == 1, (
+                    f"Only one task group with name {self.selected_task_group} should exist"
+                )
+
+                # Since we have filtered to one of the task_groups, we need to ensure that
+                # the selected_task_group percentage compute allocation is defaulted to 1.0
+                current_percentage = selected_task_groups[0]["percentage"]
+                logger.info(
+                    f"`percentage` for task_group {self.selected_task_group} is "
+                    f"{current_percentage}, setting it to 1.0"
+                )
+                selected_task_groups[0]["percentage"] = 1.0
+
+                self.task_groups = selected_task_groups
+
+            # Call the original method
+            return func(self, *args, **kwargs)
+
+        return wrapper
--- a/openfl/component/assigner/random_grouped_assigner.py
+++ b/openfl/component/assigner/random_grouped_assigner.py
@@ -6,7 +6,7 @@
 
 import numpy as np
 
-from openfl.component.assigner.assigner import Assigner
+from openfl.component.assigner import Assigner
 
 
 class RandomGroupedAssigner(Assigner):
@@ -33,16 +33,19 @@ class RandomGroupedAssigner(Assigner):
         \* - Plan setting.
     """
 
+    with_selected_task_group = Assigner.with_selected_task_group
+
     def __init__(self, task_groups, **kwargs):
         """Initializes the RandomGroupedAssigner.
 
         Args:
             task_groups (list of object): Task groups to assign.
-            **kwargs: Additional keyword arguments.
+            **kwargs: Additional keyword arguments, including mode.
         """
         self.task_groups = task_groups
         super().__init__(**kwargs)
 
+    @with_selected_task_group
     def define_task_assignments(self):
         """Define task assignments for each round and collaborator.
 
--- a/openfl/component/assigner/static_grouped_assigner.py
+++ b/openfl/component/assigner/static_grouped_assigner.py
@@ -32,6 +32,8 @@ class StaticGroupedAssigner(Assigner):
         \* - Plan setting.
     """
 
+    with_selected_task_group = Assigner.with_selected_task_group
+
     def __init__(self, task_groups, **kwargs):
         """Initializes the StaticGroupedAssigner.
 
@@ -42,6 +44,7 @@ class StaticGroupedAssigner(Assigner):
         self.task_groups = task_groups
         super().__init__(**kwargs)
 
+    @with_selected_task_group
     def define_task_assignments(self):
         """Define task assignments for each round and collaborator.
 
--- a/tests/openfl/component/assigner/test_assigner.py
+++ b/tests/openfl/component/assigner/test_assigner.py
@@ -19,7 +19,7 @@ def assigner():
 
 def test_get_aggregation_type_for_task_none(assigner):
     """Assert that aggregation type of custom task is None."""
-    task_name = 'test_name'
+    task_name = "test_name"
     tasks = {task_name: {}}
 
     assigner = assigner(tasks, None, None)
@@ -31,11 +31,9 @@ def test_get_aggregation_type_for_task_none(assigner):
 
 def test_get_aggregation_type_for_task(assigner):
     """Assert that aggregation type of task is getting correctly."""
-    task_name = 'test_name'
-    test_aggregation_type = 'test_aggregation_type'
-    tasks = {task_name: {
-        'aggregation_type': test_aggregation_type
-    }}
+    task_name = "test_name"
+    test_aggregation_type = "test_aggregation_type"
+    tasks = {task_name: {"aggregation_type": test_aggregation_type}}
     assigner = assigner(tasks, None, None)
 
     aggregation_type = assigner.get_aggregation_type_for_task(task_name)
@@ -46,13 +44,23 @@ def test_get_aggregation_type_for_task(assigner):
 def test_get_all_tasks_for_round(assigner):
     """Assert that assigner tasks object is list."""
     assigner = Assigner(None, None, None)
-    tasks = assigner.get_all_tasks_for_round('test')
+    tasks = assigner.get_all_tasks_for_round("test")
 
     assert isinstance(tasks, list)
 
+def test_default_task_group(assigner):
+    """Assert that by default learning task_group is assigned."""
+    assigner = Assigner(None,None,None)
+    assert assigner.selected_task_group == None
+
+def test_task_group_filtering_no_task_groups(assigner):
+    """Assert that task_group_filtering does not filter when no task_groups are defined."""
+    assigner = Assigner(None,None,None)
+    assigner.selected_task_group = "test_group"
+    assigner.define_task_assignments()
+    assert not hasattr(assigner, "task_groups")
 
 class TestNotImplError(TestCase):
-
     def test_define_task_assignments(self):
         # TODO: define_task_assignments is defined as a mock in multiple fixtures,
         # which leads the function to behave as a mock here and other tests.
@@ -61,9 +69,9 @@ class TestNotImplError(TestCase):
     def test_get_tasks_for_collaborator(self):
         with self.assertRaises(NotImplementedError):
             assigner = Assigner(None, None, None)
-            assigner.get_tasks_for_collaborator('col1', 0)
+            assigner.get_tasks_for_collaborator("col1", 0)
 
     def test_get_collaborators_for_task(self):
         with self.assertRaises(NotImplementedError):
             assigner = Assigner(None, None, None)
-            assigner.get_collaborators_for_task('task_name', 0)
+            assigner.get_collaborators_for_task("task_name", 0)
--- a/tests/openfl/component/assigner/test_random_grouped_assigner.py
+++ b/tests/openfl/component/assigner/test_random_grouped_assigner.py
@@ -4,7 +4,7 @@
 
 import pytest
 
-from openfl.component.assigner import RandomGroupedAssigner
+from openfl.component.assigner import RandomGroupedAssigner, Assigner
 
 ROUNDS_TO_TRAIN = 10
 
@@ -14,14 +14,21 @@ def task_groups():
     """Initialize task groups."""
     task_groups = [
         {
-            'name': 'train_and_validate',
+            'name': 'learning',
             'percentage': 1.0,
             'tasks': [
                 'aggregated_model_validation',
                 'train',
                 'locally_tuned_model_validation'
             ]
-        }
+        },
+        {
+            'name': 'evaluation',
+            'percentage': 0,
+            'tasks': [
+                'aggregated_model_validation'
+            ]
+        },
     ]
     return task_groups
 
@@ -35,12 +42,12 @@ def authorized_cols():
 @pytest.fixture
 def assigner(task_groups, authorized_cols):
     """Initialize assigner."""
-    assigner = RandomGroupedAssigner
-
-    assigner = assigner(task_groups,
-                        tasks=None,
-                        authorized_cols=authorized_cols,
-                        rounds_to_train=ROUNDS_TO_TRAIN)
+    assigner = RandomGroupedAssigner(
+        task_groups=task_groups,  # Pass task_groups here
+        tasks=None,
+        authorized_cols=authorized_cols,
+        rounds_to_train=ROUNDS_TO_TRAIN
+    )
     return assigner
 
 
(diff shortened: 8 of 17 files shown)


openfl/component/assigner/mode_based_assigner.py
```diff
@@ -0,0 +1,59 @@
...
+    r"""The task assigner maintains a list of tasks.
+
+    Also it decides the policy for which collaborator should run those tasks
+    There may be many types of policies implemented, but a natural place to
+    start is with a:
+
+        - ModeBasedAssigner :
+            Given a set of task groups and mode it filters the task groups based
+            on mode. It futher enforces checks for specific modes.
+            Post filtering it deletgates the task assignment to RandomGroupedAssigner.
```
teoparvanov: Why not keep it simple, and just have this in the docstring:
```
    Given a set of task groups and mode it filters the task groups based
    on mode. It futher enforces checks for specific modes.
    Post filtering it deletgates the task assignment to RandomGroupedAssigner.
```

openfl/component/assigner/mode_based_assigner.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2020-2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""Mode based assigner module."""
+
+from openfl.component.assigner.random_grouped_assigner import RandomGroupedAssigner
+
+
+class ModeBasedAssigner(RandomGroupedAssigner):
```
teoparvanov: Why extending the `RandomGroupedAssigner` specifically? Wouldn't mode switching also be applicable to `StaticGroupedAssigner` (among other possible implementations)?

Alternatively you could consider implementing the `ModeBasedAssigner` as a decorator for `define_task_assignments()` ?

tests/openfl/component/assigner/test_mode_based_assigner.py
```diff
@@ -0,0 +1,74 @@
...
+def sample_task_groups():
+    return [
+        {"name": "train", "tasks": ["task1", "task2"], "percentage": 1.0},
+        {"name": "evaluate", "tasks": ["task3"], "percentage": 1.0},
+        {"name": "random-mode", "tasks": ["task4"], "percentage": 1.0}
+    ]
+
+def test_init_with_valid_mode(sample_tasks, sample_authorized_cols):
+    """Test initialization with valid mode and task groups."""
+    task_groups = [{"name": "train", "tasks": ["task1"], "percentage": 1.0}]
```
teoparvanov: nit: `task_groups` -> `training_task_groups`

tests/openfl/component/assigner/test_mode_based_assigner.py
```diff
@@ -0,0 +1,74 @@
...
+@pytest.fixture
+def sample_authorized_cols():
+    return ["col1", "col2", "col3"]
+
+@pytest.fixture
+def sample_task_groups():
+    return [
+        {"name": "train", "tasks": ["task1", "task2"], "percentage": 1.0},
+        {"name": "evaluate", "tasks": ["task3"], "percentage": 1.0},
+        {"name": "random-mode", "tasks": ["task4"], "percentage": 1.0}
```
teoparvanov: `"train-and-validate"` maybe?

teoparvanov (commented):
Hi @ishaileshpant, it looks good - here are some thoughts and comments from my first read-through

openfl/component/assigner/mode_based_assigner.py
ishaileshpant: This was based on all the current workspaces using RandomGroupAssigner and to ensure backward compat thought of extending it minimizing impacts on existing workspace

kminhta: > This was based on all the current workspaces using RandomGroupAssigner and to ensure backward compat thought of extending it minimizing impacts on existing workspace

That reasoning makes sense regarding minimizing impact; however, I imagine that we want the ability to select the mode for any subclass of `Assigner`. Is it possible to move this logic to the base class `Assigner` instead or even as a decorator as @teoparvanov suggests

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
```diff
@@ -32,10 +32,12 @@ network :
   defaults : plan/defaults/network.yaml
 
 assigner :
-  defaults : plan/defaults/federated-evaluation/assigner.yaml
-  
+  defaults : plan/defaults/assigner.yaml
+  settings :
+    mode : evaluate
```
kminhta: Correct me if I am wrong, but it seems like this would still require a user to redistribute the plan if they wanted to run federated evaluation. There is nothing inherently wrong with this, of course, but I don't think it quite grants the ability for a user to seamlessly switch between modes. In fact, wouldn't they still require two separate plans/workspaces?

kminhta (commented):
Thanks @ishaileshpant! This is good foundation. Overall, I like the idea of distributing the modes as separate task groups and letting the assigner distribute them as appropriate. I have two comments for discussion.

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
ishaileshpant: Agree seems like we need to pull this out of plan itself, possibly aggregator as suggested by @psfoley would be the best place without changing plan across different runs.
the only doubt I have is afaiu its generally different set of collaborators that do training than ones that will do validations - for that purpose anyways it's a different federation run with plan mode change and wouldn't that mandate a plan distribution?

openfl/component/assigner/mode_based_assigner.py
ishaileshpant: Moving to Assigner level makes a lot of sense to me

ishaileshpant: On second thought - Assigner work on tasks rather the task_groups and the filtering is based on group_name not task name - this gives flexibility, even beyond the default tasks, to filter based on group name instead of task name

kminhta: > Assigner work on tasks rather the task_groups and the filtering is based on group_name not task name

Oh, this is true. Good point. 

A follow up question is that `mode` appears directly tied to the `task_group.name` with the primary added feature being that it'll assert that the rounds be set to 1 if in evaluation mode. If this is the case, do you think it would be better to implement this filtering logic directly into `RandomGroupedAssigner` (and the other assigners) rather than creating a separate class? I am still mainly echoing Teo's sentiment that mode switching should be applicable to all assigners

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
kminhta: > Agree seems like we need to pull this out of plan itself, possibly aggregator as suggested by @psfoley would be the best place without changing plan across different runs.

Agreed, this is also the direction I am leaning. My open question here is how to signal to the collaborators that the experiment is starting in evaluation mode since there's a huge difference in commitment between full training and a single evaluation round. 

Maybe I am overthinking this, though, since you bring up a good point in your second point:

> the only doubt I have is afaiu its generally different set of collaborators that do training than ones that will do validations - for that purpose anyways it's a different federation run with plan mode change and wouldn't that mandate a plan distribution?

This is correct intuition, and the current assigners do actually support assigning individual collaborators (statically or randomly) to a task group [[ref](https://github.com/securefederatedai/openfl/blob/develop/openfl/component/assigner/assigner.py#L66)]. So you could have a set of collaborators dedicated entirely to evaluation within the same plan without redistribution,. Afaiu, the biggest caveat is that they perform their tasks alongside the "training" group instead of in a separate run, but I think having a mode switching feature that is triggered outside the plan may potentially resolve this. However, for this to work, I think you would need to able to allow an experiment to start once the subset of collaborators for that task mode are online rather than once all registered participants are online.

openfl/component/assigner/mode_based_assigner.py
ishaileshpant: @kta-intel @teoparvanov i've refactored this to use decorator mode_based_filtering and all those sub-classes which has task_groups to leverage the decorator
removed new class and its test cases

openfl/component/assigner/assigner.py
```diff
@@ -93,3 +102,47 @@ def get_aggregation_type_for_task(self, task_name):
...
+                # Verify task_groups exists before attempting filtering
+                if not hasattr(self, "task_groups"):
+                    logger.warning(
+                        "Mode-based filtering was requested but no task_groups found. "
+                        "Skipping filtering. This might be intentional for custom assigners."
+                    )
+                    return func(self, *args, **kwargs)
+
+                if not self.task_groups:
+                    logger.warning(
```
ishaileshpant: @teoparvanov @kta-intel in this edge case should we err or warn just like line 138 errs post filtering if task groups are empty - these two behaviors are contradictory?

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
ishaileshpant: another point is number_of_rounds - for training it most probably will be >1 but for evaluate we (as of now) only err in case its higher and mandate a manual change to 1 - think while we are at it we should potentially ignore round_number and set it to 1 cause if we don't - it again mandates plan change b/w train and evaluate run - how ever that would mean plan is not implemented verbatim in the run (showing higher round number but in mode evaluate overriden to 1) will that be acceptable? @psfoley ?

openfl/component/assigner/mode_based_assigner.py
teoparvanov: Nice work adding the decorator, @ishaileshpant ! Just wondering though, why removing the test? Isn't it still possible to test the `RandomGroupedAssigner`, even though internally it uses a decorator now?

openfl/component/assigner/assigner.py
teoparvanov: I suggest to "fail fast" in such cases, and raise an error, as on line 138. You could do so via an assertion, instead of a `ValueError` - see my comment below.

openfl/component/assigner/assigner.py
```diff
@@ -93,3 +102,47 @@ def get_aggregation_type_for_task(self, task_name):
...
+                    )
+                    return func(self, *args, **kwargs)
+
+                # Perform the filtering
+                self.task_groups = [
+                    group for group in self.task_groups if group["name"] == self.mode
+                ]
+
+                if not self.task_groups:
+                    raise ValueError(f"No task groups found after filtering for mode: {self.mode}")
```
teoparvanov: this isn't really a `ValueError`, but more of an invariant breach, so I'd suggest refactoring it as an assertion:

```python
assert self.task_groups, f"No task groups found after filtering for mode: {self.mode}"
```

openfl/component/assigner/random_grouped_assigner.py
```diff
@@ -38,11 +38,12 @@ def __init__(self, task_groups, **kwargs):
...
         Args:
             task_groups (list of object): Task groups to assign.
-            **kwargs: Additional keyword arguments.
+            **kwargs: Additional keyword arguments, including mode.
         """
         self.task_groups = task_groups
         super().__init__(**kwargs)
 
+    @Assigner.mode_based_filtering
     def define_task_assignments(self):
```
teoparvanov: If you imported the `mode_based_filtering` symbol separately, you could simply write:
```python
@mode_based_filtering
def define_task_assignments(self):
    ...
```

openfl/component/assigner/static_grouped_assigner.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, task_groups, **kwargs):
         self.task_groups = task_groups
         super().__init__(**kwargs)
 
+    @Assigner.mode_based_filtering
```
teoparvanov: same here?

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
teoparvanov: I'd also be in favor of adding mode switching at the aggregator level (via the `fx aggregator start` command f.e.), in addition to the setting in the FL plan. I don't see a security or trust issue in internally overriding the number of rounds to 1 in evaluation mode (esp. given that by definition, the model isn't updated in this case).

PS: if you think this would bloat the current PR, you could also consider implementing it in a follow-up one.

openfl/component/assigner/mode_based_assigner.py
ishaileshpant: yeah i am adding them to assigner existing test cases file so remove the existing one

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
ishaileshpant: Already on it

ishaileshpant: The only concern i have with overriding without changing the plan is that plan hash, which afaiu, goes into aggregator uuid as well, will not be representative of the actual running plan - but i agree its not a security risk, just a little deflecting

MasterSkepticista (commented):
It is not quite clear to me, the need for a new task group and the associated machinery. I will lay my understanding, please help me understand what is this PR solving that the current design does not support.

This is what is supplied in a plan today:

```yaml
task_groups  :
    - name       : train_and_validate
      percentage : 1.0
      tasks      :
        - aggregated_model_validation
        - train
        - locally_tuned_model_validation
```

Questions:
1. Why is it not sufficient for a user to simply provide `aggregated_model_validation` task?
  ```yaml
    task_groups  :
      - name       : validate
        percentage : 1.0
        tasks      :
          - aggregated_model_validation
  ```
2. Per my understanding, the `name: train_and_validate` is for the user's convenience, and could be anything. OpenFL does not make any internal decisions based on a task group "name".
3. Which brings me to the confusion behind creating `name: evaluate` task group, and all the gymnastics being done around it in this PR.

ishaileshpant:
> It is not quite clear to me, the need for a new task group and the associated machinery. I will lay my understanding, please help me understand what is this PR solving that the current design does not support.

The new task group name is to ensure that we have clear demarcation between train (or train_and_validate) and evaluation /evaluate type of tasks to disambiguate the two - validation during training vs evaluation without training. Thats why renamed the previous name "validate" as you've suggested to "evaluate" to make it explicit

ishaileshpant:
> Questions:
> 
> 1. Why is it not sufficient for a user to simply provide `aggregated_model_validation` task?
> 
> ```yaml
>   task_groups  :
>     - name       : validate
>       percentage : 1.0
>       tasks      :
>         - aggregated_model_validation

Thats exactly what we have done in the defaults tasks yaml - defined both the task groups and then using mode as a filter b/w training and evaluate run using the same plan without mutating the plan or needing a different plan or needing any manual changes to plan to run in train or evaluate mode

This PR though expose the mode field at assigner level but eventually as per the PR #1258 this will be dictated by Aggregator without the need to change the plan yaml b/w training and evaluate run

ishaileshpant:
>Per my understanding, the name: train_and_validate is for the user's convenience, and could be anything. OpenFL does not make any internal decisions based on a task group "name".

well this is true as per the current design/implementation and that's exactly what we are trying to extend by using group name as the driver - i agree we can do better naming or may be define a new attribute to link the mode to - in spirit of keeping the changes to minimal and still give a better UX to use the same plan for train and evaluate, existing task_group.name seems like the best place to map the mode to

If we don't explicitly do valid mode checks (Please see #1258) then we can make it flexible enough for user to define any task_group name of their choice and basically leverage this design further without needing a code change for new task_groups if needed in future

MasterSkepticista:
> If we don't explicitly do valid mode checks (Please see https://github.com/securefederatedai/openfl/pull/1258) then we can make it flexible enough for user to define any task_group name of their choice and basically leverage this design further without needing a code change for new task_groups if needed in future

I prefer this because eval is a problem of task group switching (in essence FedEval vs FedTrain is a "selection" problem, not one of "filtering", correct?).

Training a model is a task, evaluation is a task, perhaps purging and regeneration of cryptography keys is a task (@theakshaypant)? These tasks would be defined beforehand in the plan. OpenFL will verify if the requested task name matches one of the specified ones in the plan. But that is all to it; no "filtering" or modification of assigner code is (or needs to be) necessary IMO.

All that I care as a user, is to be able to select what I want to run from the aggregator side. A user wouldn't be happy if a framework requires them to dig into the source code to add a new task with brand-new `if-else` statements.

Documented an example here: https://github.com/securefederatedai/openfl/pull/1258#discussion_r1911907837 WDYT?

ishaileshpant:
> I prefer this because eval is a problem of task group switching (in essence FedEval vs FedTrain is a "selection" problem, not one of "filtering", correct?).

Selection or filtering they are two ways of saying / doing the same thing IMO :), nevertheless with mode/task/task-group being supplied at aggregator level still needs to be passed to assigner cause that's where we have logic for task group parsing and task assignments and is rightly placed. 

We just need to leverage the same to decide which is the task-group of interest to the assigner as per the aggregator mode

> All that I care as a user, is to be able to select what I want to run from the aggregator side. A user wouldn't be happy if a framework requires them to dig into the source code to add a new task with brand-new `if-else` statements.

Well there is no if-else as of now and IDT will need in future as well except the task-group.name comparison to mode. Now we rename that to --task or better yet task-group nevertheless we still need the check, that way the default assigner can simply have all task-groups and aggregator decides which one is of interest in a specific federation run 

The only thing I feel we need to pull out of this PR and push to aggregator level is the mode/task-group based validation checks - e.g for task-group.name=evaluation we need to enforce round_number to 1 and this, in hindsight, is not rightly placed in assigner but aggregator. I will pull this out of this PR and move to #1258 

@psfoley Does this make sense?

openfl-workspace/torch_cnn_mnist_fed_eval/plan/plan.yaml
ishaileshpant: @kta-intel @teoparvanov please see #1258 where this mode (or selected-task-group) attribute is driven by Aggregator cli and leveraged by assigner to filter out other tasks groups other then selected task group for assignment to collaborators, once that is approved and merged will rebase this and align with name changes

openfl/component/assigner/assigner.py
ishaileshpant: Done

openfl/component/assigner/assigner.py
ishaileshpant: Done

openfl/component/assigner/random_grouped_assigner.py
ishaileshpant: Done

openfl/component/assigner/static_grouped_assigner.py
ishaileshpant: Done

tests/openfl/component/assigner/test_random_grouped_assigner.py
```diff
@@ -35,19 +42,42 @@ def authorized_cols():
...
+
+# @pytest.mark.parametrize('round_number', range(ROUNDS_TO_TRAIN))
+# def test_get_filtered_tasks_for_collaborator(assigner, task_groups,
+#                                     authorized_cols, round_number):
+#     """Test that assigner tasks correspond to task groups defined."""
+#     assigner.selected_task_group=task_groups[1]['name']
+#     assigner.define_task_assignments()
+#     tasks = assigner.get_tasks_for_collaborator(
+#         authorized_cols[0], round_number)
+#     assert tasks == task_groups[1]['tasks']
```
teoparvanov: Is this test disabled on purpose?

tests/openfl/component/assigner/test_assigner.py
```diff
@@ -46,13 +44,23 @@ def test_get_aggregation_type_for_task(assigner):
...
     assigner = Assigner(None, None, None)
-    tasks = assigner.get_all_tasks_for_round('test')
+    tasks = assigner.get_all_tasks_for_round("test")
 
     assert isinstance(tasks, list)
 
+def test_default_task_group(assigner):
+    """Assert that by default learning task_group is assigned."""
+    assigner = Assigner(None,None,None)
+    assert assigner.selected_task_group == 'learning'
```
teoparvanov: For completeness, could you also add a unit test where `assigner.selected_task_group == 'evaluation'`?

openfl/component/assigner/static_grouped_assigner.py
```diff
@@ -42,6 +44,7 @@ def __init__(self, task_groups, **kwargs):
         self.task_groups = task_groups
         super().__init__(**kwargs)
 
+    @task_group_filtering
     def define_task_assignments(self):
```
teoparvanov: I suggest renaming the decorator to `@with_selected_task_groups`. This way the code "reads" almost like in English: "With the selected task groups, define the task assignments...". WDYT?

teoparvanov (approved):
Nice work, @ishaileshpant, I think it's looking good at this stage!
I just have a couple of minor remaining comments:

tanwarsh (approved):

rahulga1 (approved):
LGTM, just modify based on Teo's suggestions.

openfl-workspace/workspace/plan/defaults/assigner.yaml
```diff
@@ -7,3 +7,7 @@ settings :
         - aggregated_model_validation
         - train
         - locally_tuned_model_validation
+    - name       : evaluation
+      percentage : 1.0
+      tasks      :
+        - aggregated_model_validation
```
psfoley: I need to provide some additional context here because this is not how the assigner is supposed to work. Most of the existing examples illustrate the simplest scenario (my fault) where all participants are allocated to a single task group. The RandomGroupedAssigner is really intended for scenarios when collaborators should be placed in _different_ task groups for a given round. The original definition of Federated Averaging called for a small subset of collaborators to be training for a given round, and a much larger set to be performing validation / evaluation. This was done as part of the FeTS Initiative. This allows collaborators to contribute to the final model, but allows computational resources / time to be saved for the totality of the federation. The percentage *_across_* task groups should sum to 1.0 (representing 100% of collaborators), not be set to 1.0 for each task group. An example is "evaluation" percentage set to 0.8 (80% performing validation / evaluation for each round), and "learning" percentage set to 0.2 (20% performing all learning related tasks for a round). If I'm reading the code correctly, the reason you're not running into this problem is you've set the "learning" task group as the default parameter. What should be done instead if you want to filter by task group is to shift the percentage for each task group *and* alert users in some way that this has been done / is a permitted change in the plan. Instead of having a default task group at all, this can be achieved by having multiple task groups in the way you have here, but setting the learning task_group to `1.0`, and evaluation to `0.0`. When the evaluation task group is specified from command line this can shift to learning percentage '0.0', and evaluation percentage `1.0`. This way users will still be able to define multiple task groups to run within a single federation.

openfl/component/assigner/assigner.py
```diff
@@ -35,18 +40,27 @@ class Assigner:
         \* - ``tasks`` argument is taken from ``tasks`` section of FL plan YAML file.
     """
 
-    def __init__(self, tasks, authorized_cols, rounds_to_train, **kwargs):
+    def __init__(
+        self,
+        tasks,
+        authorized_cols,
+        rounds_to_train,
+        selected_task_group: str = "learning",
```
psfoley: See comment as above - we should not have a default task group. This will override user's configuration in the case that multiple task groups are set in the plan. The right way to make this the default is the default plan learning task group set to 1.0 and others (i.e. evaluation) set to 0.0

psfoley (changes requested):
Thanks for the PR, @ishaileshpant. The scenario of multiple user defined task groups needs to be considered before this can be approved. Details are provided in the comments. I can go into more depth on the Assigner design during our standup as well.

openfl-workspace/workspace/plan/defaults/assigner.yaml
payalcha: @psfoley - In case of multiple task-groups for e.g learning and evaluation in same plan.yaml. How the run will behave.

openfl/component/aggregator/aggregator.py
```diff
@@ -130,15 +130,25 @@ def __init__(
         self.straggler_handling_policy = (
             straggler_handling_policy or CutoffTimeBasedStragglerHandling()
         )
-        self._end_of_round_check_done = [False] * rounds_to_train
-        self.stragglers = []
 
         self.rounds_to_train = rounds_to_train
+        if self.task_group == "evaluation":
+            self.rounds_to_train = 1
```
payalcha, line 136: We don't need to modify rounds_to_train in  plan.yaml post this changes. 
Aggregator will take care of the modification.

openfl/component/assigner/assigner.py
payalcha: Does this means that multiple task-groups can be processed in the same experiment?
Currently all examples have only one task-group running.
Do we have any example or use-case in which we can see that multiple task-groups running in same experiment.

openfl/component/assigner/assigner.py
```diff
@@ -93,3 +107,50 @@ def get_aggregation_type_for_task(self, task_name):
...
+                        "Task group specified for selection but no task_groups found. "
+                        "Skipping filtering. This might be intentional for custom assigners."
+                    )
+                    return func(self, *args, **kwargs)
+
+                assert self.task_groups, "No task_groups defined in assigner."
+
+                # Perform the filtering
+                selected_task_groups = [
+                    group for group in self.task_groups if group["name"] == self.selected_task_group
```
psfoley, line 135: What happens if someone accidentally / purposefully lists "learning" or "evaluation" twice in the task groups configuration?

psfoley: Nevermind! I see this is handled

psfoley (approved):

ishaileshpant:
@kta-intel for the last task as discussed in #1258 i've created an issue #1321 and will track that fix via that

PR #1229, opened 2024-12-23
Add persistent db module and recovery logic

cloudnoize:
- Introduce a new persistent DB module.
- Create tables and logic to persist the data needed to recovery the Aggregator after restart.
- Use the database from the Aggregator.
- Introduce recovery logic to use the persisted data to recover the Aggregator state as it was before the crash.
- Fix for collaborator to send correct tensor key.

### DIFF
--- a/openfl-workspace/workspace/plan/defaults/aggregator.yaml
+++ b/openfl-workspace/workspace/plan/defaults/aggregator.yaml
@@ -2,3 +2,4 @@ template : openfl.component.Aggregator
 settings :
     db_store_rounds : 2
     persist_checkpoint: True
+    persistent_db_path: save/tensor.db
--- a/openfl-workspace/workspace/plan/defaults/aggregator.yaml
+++ b/openfl-workspace/workspace/plan/defaults/aggregator.yaml
@@ -1,3 +1,4 @@
 template : openfl.component.Aggregator
 settings :
     db_store_rounds : 2
+    persist_checkpoint: True
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -12,8 +12,7 @@ from typing import List, Optional
 
 import openfl.callbacks as callbacks_module
 from openfl.component.straggler_handling_functions import CutoffTimeBasedStragglerHandling
-from openfl.databases import TensorDB
-from openfl.databases import PersistentTensorDB
+from openfl.databases import PersistentTensorDB, TensorDB
 from openfl.interface.aggregation_functions import WeightedAverage
 from openfl.pipelines import NoCompressionPipeline, TensorCodec
 from openfl.protocols import base_pb2, utils
@@ -85,7 +84,7 @@ class Aggregator:
         write_logs=False,
         callbacks: Optional[List] = None,
         persist_checkpoint=True,
-        persistent_db_path=None
+        persistent_db_path=None,
     ):
         """Initializes the Aggregator.
 
@@ -143,7 +142,11 @@ class Aggregator:
 
         self.tensor_db = TensorDB()
         if persist_checkpoint:
-            logger.info("Persistent checkpoint is enabled")
+            persistent_db_path = persistent_db_path or "tensor.db"
+            logger.info(
+                "Persistent checkpoint is enabled, setting persistent db at path %s",
+                persistent_db_path,
+            )
             self.persistent_db = PersistentTensorDB(persistent_db_path)
         else:
             logger.info("Persistent checkpoint is disabled")
@@ -168,7 +171,6 @@ class Aggregator:
         # these enable getting all tensors for a task
         self.collaborator_tasks_results = {}  # {TaskResultKey: list of TensorKeys}
         self.collaborator_task_weight = {}  # {TaskResultKey: data_size}
-        
 
         # maintain a list of collaborators that have completed task and
         # reported results in a given round
@@ -177,8 +179,13 @@ class Aggregator:
         self.lock = Lock()
         self.use_delta_updates = use_delta_updates
 
+        self.model = None  # Initialize the model attribute to None
         if self.persistent_db and self._recover():
             logger.info("recovered state of aggregator")
+
+        # The model is built by recovery if at least one round has finished
+        if self.model:
+            logger.info("Model was loaded by recovery")
         elif initial_tensor_dict:
             self._load_initial_tensors_from_dict(initial_tensor_dict)
             self.model = utils.construct_model_proto(
@@ -204,19 +211,21 @@ class Aggregator:
         # https://github.com/securefederatedai/openfl/pull/1195#discussion_r1879479537
         self.callbacks.on_experiment_begin()
         self.callbacks.on_round_begin(self.round_number)
-		
+
     def _recover(self):
-        """Populates the aggregator state to the state it was prior a restart
-        """
+        """Populates the aggregator state to the state it was prior a restart"""
         recovered = False
         # load tensors persistent DB
-        logger.info("Recovering previous state from persistent storage")
-        tensor_key_dict = self.persistent_db.load_tensors(self.persistent_db.get_tensors_table_name())
+        tensor_key_dict = self.persistent_db.load_tensors(
+            self.persistent_db.get_tensors_table_name()
+        )
         if len(tensor_key_dict) > 0:
             logger.info(f"Recovering {len(tensor_key_dict)} model tensors")
             recovered = True
             self.tensor_db.cache_tensor(tensor_key_dict)
-            committed_round_number, self.best_model_score = self.persistent_db.get_round_and_best_score()
+            committed_round_number, self.best_model_score = (
+                self.persistent_db.get_round_and_best_score()
+            )
             logger.info("Recovery - Setting model proto")
             to_proto_tensor_dict = {}
             for tk in tensor_key_dict:
@@ -225,24 +234,29 @@ class Aggregator:
             self.model = utils.construct_model_proto(
                 to_proto_tensor_dict, committed_round_number, self.compression_pipeline
             )
-            # round number is the current round which is still in process i.e. committed_round_number + 1
+            # round number is the current round which is still in process
+            #  i.e. committed_round_number + 1
             self.round_number = committed_round_number + 1
-            logger.info("Recovery - loaded round number %s and best score %s", self.round_number,self.best_model_score)
-        
-        next_round_tensor_key_dict = self.persistent_db.load_tensors(self.persistent_db.get_next_round_tensors_table_name())
+            logger.info(
+                "Recovery - loaded round number %s and best score %s",
+                self.round_number,
+                self.best_model_score,
+            )
+
+        next_round_tensor_key_dict = self.persistent_db.load_tensors(
+            self.persistent_db.get_next_round_tensors_table_name()
+        )
         if len(next_round_tensor_key_dict) > 0:
             logger.info(f"Recovering {len(next_round_tensor_key_dict)} next round model tensors")
             recovered = True
             self.tensor_db.cache_tensor(next_round_tensor_key_dict)
-            
-        
-        logger.info("Recovery - Finished populating tensor DB")
+
         logger.debug("Recovery - this is the tensor_db after recovery: %s", self.tensor_db)
 
         if self.persistent_db.is_task_table_empty():
             logger.debug("task table is empty")
             return recovered
-        
+
         logger.info("Recovery - Replaying saved task results")
         task_id = 1
         while True:
@@ -259,8 +273,15 @@ class Aggregator:
                 NamedTensor.FromString(serialized_tensor)
                 for serialized_tensor in serialized_tensors
             ]
-            logger.info("Recovery - Replaying task results %s %s %s",collaborator_name ,round_number, task_name )
-            self.process_task_results(collaborator_name, round_number, task_name, data_size, named_tensors)
+            logger.info(
+                "Recovery - Replaying task results %s %s %s",
+                collaborator_name,
+                round_number,
+                task_name,
+            )
+            self.process_task_results(
+                collaborator_name, round_number, task_name, data_size, named_tensors
+            )
             task_id += 1
         return recovered
 
@@ -342,12 +363,16 @@ class Aggregator:
             # Transaction to persist/delete all data needed to increment the round
             if self.persistent_db:
                 if self.next_model_round_number > 0:
-                    next_round_tensors = self.tensor_db.get_tensors_by_round_and_tags(self.next_model_round_number,("model",))
-                self.persistent_db.finalize_round(tensor_tuple_dict,next_round_tensors,self.round_number,self.best_model_score)
-                logger.info(
-                        "Persist model and clean task result for round %s",
-                        round_number,
+                    next_round_tensors = self.tensor_db.get_tensors_by_round_and_tags(
+                        self.next_model_round_number, ("model",)
                     )
+                self.persistent_db.finalize_round(
+                    tensor_tuple_dict, next_round_tensors, self.round_number, self.best_model_score
+                )
+                logger.info(
+                    "Persist model and clean task result for round %s",
+                    round_number,
+                )
             self.last_tensor_dict = tensor_dict
         self.model = utils.construct_model_proto(
             tensor_dict, round_number, self.compression_pipeline
@@ -690,13 +715,19 @@ class Aggregator:
         # Save task and its metadata for recovery
         serialized_tensors = [tensor.SerializeToString() for tensor in named_tensors]
         if self.persistent_db:
-            self.persistent_db.save_task_results(collaborator_name,round_number,task_name,data_size,serialized_tensors)
-            logger.debug(f"Persisting task results {task_name} from {collaborator_name} round {round_number}")
+            self.persistent_db.save_task_results(
+                collaborator_name, round_number, task_name, data_size, serialized_tensors
+            )
+            logger.debug(
+                f"Persisting task results {task_name} from {collaborator_name} round {round_number}"
+            )
         logger.info(
             f"Collaborator {collaborator_name} is sending task results "
             f"for {task_name}, round {round_number}"
         )
-        self.process_task_results(collaborator_name,round_number,task_name,data_size,named_tensors)
+        self.process_task_results(
+            collaborator_name, round_number, task_name, data_size, named_tensors
+        )
 
     def process_task_results(
         self,
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -84,6 +84,8 @@ class Aggregator:
         log_memory_usage=False,
         write_logs=False,
         callbacks: Optional[List] = None,
+        persist_checkpoint=True,
+        persistent_db_path=None
     ):
         """Initializes the Aggregator.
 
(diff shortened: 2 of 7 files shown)


openfl/component/aggregator/aggregator.py
```diff
@@ -990,6 +1051,7 @@ def _end_of_round_check(self):
 
         self.round_number += 1
         # resetting stragglers for task for a new round
+        #E.L should it be saved?
```
teoparvanov: If my understanding is correct, during restart recovery the round is replayed from the start. So my intuition is that the knowledge of which collaborators are late (stragglers) can be "regained" during the replay of the round. Therefore, no need to persist it IMO.

openfl/component/aggregator/aggregator.py
```diff
@@ -1005,6 +1067,7 @@ def _end_of_round_check(self):
         # Cleaning tensor db
         self.tensor_db.clean_up(self.db_store_rounds)
         # Reset straggler handling policy for the next round.
+        #E.L should it be saved?
         self.straggler_handling_policy.reset_policy_for_round()
```
teoparvanov: Same as above, I'd suggest. Otherwise we could run into weird corner cases in straggler scenarios combined with aggregator recovery.

openfl/component/aggregator/aggregator.py
```diff
@@ -154,8 +158,21 @@ def __init__(
...
-        if initial_tensor_dict:
+        # maintain a list of collaborators that have completed task and
+        # reported results in a given round
+        self.collaborators_done = []
+        # Initialize a lock for thread safety
+        self.lock = Lock()
+        self.use_delta_updates = use_delta_updates
+
+        if self._recover():
+            print("recovered state of aggregator")
```
teoparvanov: nit - use the logger for this

openfl/component/aggregator/aggregator.py
```diff
@@ -255,18 +290,31 @@ def _save_model(self, round_number, file_path):
             for k, v in og_tensor_dict.items()
         ]
         tensor_dict = {}
+        tensor_tuple_dict = {}
         for tk in tensor_keys:
             tk_name, _, _, _, _ = tk
-            tensor_dict[tk_name] = self.tensor_db.get_tensor_from_cache(tk)
+            print(f"transaction - tk is {tk} len {len(tk)} tk_name is {tk_name}")
```
teoparvanov: nit - use logger here too

openfl/component/collaborator/collaborator.py
```diff
@@ -409,17 +409,22 @@ def get_data_for_tensorkey(self, tensor_key):
...
+                    tensor_key, require_lossless=True
+                )
+            else:
+
+                # The original model tensor should be fetched from client
+                tensor_name, origin, round_number, report, tags = tensor_key
+                tags = (self.collaborator_name,) + tags
+                tensor_key = (tensor_name, origin, round_number, report, tags)
+                logger.info(
+                    f"Couldnt not find previous model layer."
```
teoparvanov: typo - Could not find

openfl/component/collaborator/collaborator.py
```diff
@@ -409,17 +409,22 @@ def get_data_for_tensorkey(self, tensor_key):
...
-                    )
             elif "model" in tags:
                 # Pulling the model for the first time
+                logger.info("Getting model from aggregator key %s ", tensor_key)
+                nparray = self.get_aggregated_tensor_from_aggregator(
+                    tensor_key, require_lossless=True
+                )
+            else:
+
+                # The original model tensor should be fetched from client
```
teoparvanov: you mean from the aggregator, right?

openfl/component/collaborator/collaborator.py
```diff
@@ -409,17 +409,22 @@ def get_data_for_tensorkey(self, tensor_key):
...
                     self.tensor_db.cache_tensor({new_model_tk: nparray})
-                else:
-                    logger.info(
-                        "Count not find previous model layer."
-                        "Fetching latest layer from aggregator"
-                    )
-                    # The original model tensor should be fetched from client
-                    nparray = self.get_aggregated_tensor_from_aggregator(
-                        tensor_key, require_lossless=True
-                    )
```
teoparvanov: I don't understand this change - why does the `else` block go outside?

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+        conn: The SQLite connection object.
+        cursor: The SQLite cursor object.
+        mutex: A threading Lock object used to ensure thread-safe operations.
+    """
+
+    def __init__(self, db_file: str = "tensordb.sqlite") -> None:
+        """Initializes a new instance of the PersistentTensorDB class."""
+        self.conn = sqlite3.connect(db_file, check_same_thread=False)
+        self.cursor = self.conn.cursor()
+        self.mutex = Lock()
```
teoparvanov: nit - why not simply `self.lock = Lock()`?

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+        """Create the database schema if it does not exist."""
+        self.cursor.execute("""
+            CREATE TABLE IF NOT EXISTS tensors (
+                id INTEGER PRIMARY KEY AUTOINCREMENT,
+                tensor_name TEXT,
+                origin TEXT,
+                round INTEGER,
+                report INTEGER,
+                tags TEXT,
+                nparray BLOB
```
teoparvanov: shouldn't we add `NOT NULL` constraints here too?

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+        insert_query = """
+        INSERT INTO task_results 
+        (collaborator_name, round_number, task_name, data_size, named_tensors)
+        VALUES (?, ?, ?, ?, ?);
+        """
+        self.cursor.execute(
+            insert_query,
+            (collaborator_name, round_number, task_name, data_size, serialized_blob),
+        )
+        print(f"Saves task result for collaborator {collaborator_name} round number {round_number} taks_name {task_name} data_size {data_size}")
```
teoparvanov: nit - use the logger

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+        Saves task results to the task_results table.
+
+        Args:
+            collaborator_name (str): Collaborator name.
+            round_number (int): Round number.
+            task_name (str): Task name.
+            data_size (int): Data size.
+            named_tensors: Named tensors (e.g., protobuf, stored as serialized binary).
+        """
+        serialized_blob = pickle.dumps(serialized_tensors)
```
teoparvanov: This is already protobuf-serialized, why pickle it additionally? Unless you assume a non-serialized object could be passed here in certain cases?

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+        self.cursor.execute(create_table_query)
+        self.conn.commit()
+
+    def save_task_results(
+        self,
+        collaborator_name: str,
+        round_number: int,
+        task_name: str,
+        data_size: int,
+        serialized_tensors,
```
teoparvanov: nit - `named_tensors`, as in the docstring?

openfl/component/aggregator/aggregator.py
```diff
@@ -194,6 +197,38 @@ def __init__(
         # https://github.com/securefederatedai/openfl/pull/1195#discussion_r1879479537
         self.callbacks.on_experiment_begin()
         self.callbacks.on_round_begin(self.round_number)
+		
+    def _recover(self):
+            if self.persistent_db.is_task_table_empty():
+                return False
+            # load tensor to tensor db
```
teoparvanov: nit - load tensors from persistent DB

openfl/component/aggregator/aggregator.py
```diff
@@ -194,6 +197,38 @@ def __init__(
...
+                if not task_result:
+                    break
+                collaborator_name = task_result["collaborator_name"]
+                round_number = task_result["round_number"]
+                task_name = task_result["task_name"]
+                data_size = task_result["data_size"]
+                serialized_tensors = task_result["named_tensors"]
+                named_tensors = [
+                    NamedTensor.FromString(tensor_string)
+                    for tensor_string in serialized_tensors
```
teoparvanov: `tensor_string` -> `tensor_bytes` ?

teoparvanov (commented):
LGTM, @cloudnoize, based on a first read-through! I have a couple of questions/comments:

openfl/component/aggregator/aggregator.py
psfoley: +1 . Straggler list does not need to be persisted.

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
+import json
+import pickle
+import sqlite3
+import numpy as np
+from threading import Lock
+from typing import Dict, Iterator, Optional
+
+from openfl.interface.aggregation_functions import AggregationFunction
```
psfoley: `AggregationFunction` import seems to be unused.

openfl/component/aggregator/aggregator.py
psfoley: This will pollute the logs. I would suggest changing this to `logger.debug`

openfl/component/aggregator/aggregator.py
```diff
@@ -137,6 +139,8 @@ def __init__(
         self.quit_job_sent_to = []
 
         self.tensor_db = TensorDB()
+        # E.L db path from configuration if exists
+        self.persistent_db = PersistentTensorDB()
```
psfoley: It would be best to make this persistent DB configurable through the plan with a parameter like `experiment_checkpoint`. It can be set to true by default, but there should be a way to disable experiment checkpointing because it increases disk usage (can be in the GB's).

openfl/component/aggregator/aggregator.py
```diff
@@ -154,8 +158,21 @@ def __init__(
...
 
-        if initial_tensor_dict:
+        # maintain a list of collaborators that have completed task and
+        # reported results in a given round
+        self.collaborators_done = []
+        # Initialize a lock for thread safety
+        self.lock = Lock()
+        self.use_delta_updates = use_delta_updates
+
+        if self._recover():
```
psfoley: Again, this should be determined by the addition of a `experiment_checkpoint` configuration in plan.yaml:

i.e. 
```
aggregator:
...
    experiment_checkpoint: True
```

```suggestion
        if experiment_checkpoint:
            self._recover()
```

openfl/component/collaborator/collaborator.py
kminhta: I'm not sure I understand this change either. I believe this internal else block is necessary for situations where the collaborator is using `delta` to update weights, but does not (for whatever reason) have the previous model layer stored in their local db. The failsafe is to just pull it directly from the aggregator. Without this block, I'd expect it to just error if it runs into this scenario (?)

openfl/component/aggregator/aggregator.py
psfoley: In addition, there needs to be some kind of check that the previously serialized state matches the current plan. This can be accomplished by storing the `federation_uuid`, which is determined by taking a hash of the plan.yaml file. Changing a plan in the aggregator's current workspace, then restoring state with the older configuration could result in unexpected behavior.

openfl/component/collaborator/collaborator.py
```diff
@@ -409,17 +409,22 @@ def get_data_for_tensorkey(self, tensor_key):
...
             elif "model" in tags:
                 # Pulling the model for the first time
+                logger.info("Getting model from aggregator key %s ", tensor_key)
+                nparray = self.get_aggregated_tensor_from_aggregator(
+                    tensor_key, require_lossless=True
+                )
+            else:
```
kminhta: Is this `else` block necessary - are we expecting a scenario where the collaborator should not be pulling the layer tagged `"model"` (which is the global model), but instead a layer tagged `(self.collaborator_name,) + tags` (which would be a local model)?

psfoley (commented):
Nice work, @cloudnoize. This is a step in the right direction, and will certainly improve user experience to permit experiment restarts after failure.

My comments for the immediate PR can be grouped into the following feedback:
1. Additional logs should be limited, and set to `debug` or condensed into a single line when state is being saved or restored.
2. There should be a plan configuration to enable / disable the persistent DB as there are space implications of storing all of the aggregator's state
3. There should be stronger checks in place to verify the previously saved state matches the restarted experiment (see individual comments for details).

As an aside - In a separate PR before OpenFL 1.8 is released I think there would be benefit in modifying the existing TensorDB / PersistentDB behind a common interface to store all state information. This should lead to cleaner in-memory / persistent DB options in the future.

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+    def _create_key_value_Store(self) -> None:
+        self.cursor.execute("""
+            CREATE TABLE IF NOT EXISTS key_value_store (
+                key TEXT PRIMARY KEY,
+                value REAL
+            )
+        """)
+        self.conn.commit()
+    
+    def init_task_results_table(self):
```
kminhta: it it necessary to have both an internal and public method for this? It seems the public method is not used

openfl/databases/persistent_db.py
```diff
@@ -0,0 +1,299 @@
...
+        CREATE TABLE IF NOT EXISTS task_results (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            collaborator_name TEXT NOT NULL,
+            round_number INTEGER NOT NULL,
+            task_name TEXT NOT NULL,
+            data_size INTEGER NOT NULL,
+            named_tensors BLOB NOT NULL
+        );
+        """
+        self.cursor.execute(create_table_query)
```
kminhta: should this have `self.conn.commit()` ?

kminhta (commented):
This is looking good overall, thanks @cloudnoize! In additional to the other reviews, I have some comments below

openfl/component/aggregator/aggregator.py
cloudnoize: sure, all the non log prints were for local testing (should have been removed prior publishing the pr)

openfl/component/aggregator/aggregator.py
cloudnoize: Thanks for the clarification

openfl/databases/persistent_db.py
cloudnoize: it's part of a transaction

openfl/databases/persistent_db.py
cloudnoize: the list itself is not serialized

openfl/databases/persistent_db.py
cloudnoize: you're right

openfl/component/collaborator/collaborator.py
cloudnoize: From testing collaborator restarts I found that the collaborator does not enter that ```if``` , I guess that while debugging I found that ```tensor_dependencies == 0``` though I don't completely remember.
I'll try to reproduce that scenario and see.

I might assumed that the condition is not placed correctly, but probably another condition is needed in contrast to rearranging it, anyway I'll reproduce and write the findings.

openfl/databases/persistent_db.py
kminhta: ohh right, I see that now. thanks - no issues, then

openfl/component/aggregator/aggregator.py
```diff
@@ -82,6 +84,8 @@ def __init__(
         log_memory_usage=False,
         write_logs=False,
         callbacks: Optional[List] = None,
+        persist_checkpoint=True,
```
psfoley, line 86: You should also add the `persist_checkpoint` to the default FL plan. 

https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/workspace/plan/defaults/aggregator.yaml

```
settings :
    db_store_rounds : 2
    persist_checkpoint: True
```

This will give users a clue about where to enable / disable this functionality. Would also be useful to add a line to the aggregator plan documentation here: https://github.com/securefederatedai/openfl/blob/6bf229e49678600d516bb0d62d6f2f2f75ff641d/docs/about/features_index/taskrunner.rst#configurable-settings

openfl/component/collaborator/collaborator.py
cloudnoize: I'm not able to reproduce it, might be a transient failure I had. I'll revert this change and handle collaborator restarts in different pr if needed.

openfl/component/aggregator/aggregator.py
cloudnoize: is it valid to change plan on a running experiment?
what shall be done in case of a mismatch? drop data and start from scratch?  does it require notifying the collaborators?

cloudnoize: @psfoley can we discuss it and implement it in a separate PR?

openfl/component/aggregator/aggregator.py
```diff
@@ -82,6 +84,8 @@ def __init__(
...
+        persistent_db_path=None
```
teoparvanov: If `persistent_db_path` is None by default, then I suggest setting `persist_checkpoint=False` by default accordingly.

PS: As part of the bump into OpenFL-Security, you should also set the `persist_checkpoint` param in the defaults there:
https://github.com/intel-innersource/frameworks.ai.openfl.openfl-security/blob/develop/client/openfl_security_workspace/workspace/plan/defaults/aggregator.yaml

openfl-workspace/workspace/plan/defaults/aggregator.yaml
```diff
@@ -1,3 +1,4 @@
 template : openfl.component.Aggregator
 settings :
     db_store_rounds : 2
+    persist_checkpoint: True
```
teoparvanov, line 4: Could you also test E2E with `persist_checkpoint: False`, just to make sure the original flow still works?

teoparvanov (approved):
Looks great, @cloudnoize - this change will make extended OpenFL experiments much more resilient!

openfl-workspace/workspace/plan/defaults/aggregator.yaml
cloudnoize: Already tested it ofc :)

openfl/component/aggregator/aggregator.py
cloudnoize: the `None` db path is to use the current workdir i.e. workspace, not to hint that the feature is disabled,
I added it as preparation for the secure-fl where we want to set it to a dedicated path for the encrypted fs.

openfl/component/aggregator/aggregator.py
kminhta: > is it valid to change plan on a running experiment? what shall be done in case of a mismatch? drop data and start from scratch? does it require notifying the collaborators?

I think the most straightforward and safest response would be to raise an error at the aggregator in the event of a mismatch to block the restart. If the experiment is allowed to restart with a plan mismatch, it would eventually be terminated when the collaborator tries to verify the response ([see](https://github.com/securefederatedai/openfl/blob/develop/openfl/transport/grpc/aggregator_client.py#L347). Adding an additional check before communication even allowed is a good thing

psfoley: After internal conversation it was agreed that looking for changes in the plan before reloading state was needed but can be pushed to another PR. Resolving this comment for now.

psfoley (approved):

kminhta (approved):

psfoley:
@cloudnoize there's a failure in CI related to persistent DB being set to none: https://github.com/securefederatedai/openfl/actions/runs/12747793067/job/35536762434?pr=1229#step:5:520

PR #1231, opened 2024-12-26
Build openfl container image and add PR pipeline for all branches

pasokan-intel:


rajithkrishnegowda (approved):

MasterSkepticista:
@pasokan-intel Add a PR description. Is this image for users who directly want to use `docker pull` for OpenFL?

PR #1238, opened 2025-01-02
1.7.x: Replace `dill` serialization with cloudpickle

theakshaypant:
Pickle and modules that wrap it can be unsafe when used to deserialize untrusted data, possible security issue.
[Ref](https://bandit.readthedocs.io/en/1.7.10/blacklists/blacklist_calls.html#b301-pickle).

Hence replacing [dill](https://pypi.org/project/dill/) with [cloudpickle](https://github.com/cloudpipe/cloudpickle) which is already a requirement and being used in the repository.

[Test run](https://github.com/securefederatedai/openfl/actions/runs/12579488765?pr=1238).


rahulga1:
@scngupta-dsp can you please review this?
we need to merge this as the existing library is being flagged by Bandit tools for security vulnerabilities.

rahulga1 (approved):

psfoley (changes requested):
My recommendation is to close this PR because CloudPickle is not always a drop in replacement for Dill. It can cause problems with frameworks like Tensorflow / Keras (at least historically). This is the reason we had a `DillSerializer`  and `CloudPickleSerializer` for the Interactive API in the past. There are also other places where [Dill](https://github.com/securefederatedai/openfl/blob/e4305f76a462d22130222cbaf31b2526472de140/openfl/experimental/workflow/utilities/metaflow_utils.py#L20) is required across the Workflow API.

Deserializing an entire object can clearly lead to malicious remote code execution. The right way to mitigate this are two-fold:

1. Allow for plan / workspace review prior to experiment launch (TEE's are the main way to ensure runtime integrity after agreement)
2. Add a flag that limits what can be sent across the network (i.e. floating point arrays for model weights / metrics). This feature is targeted for OpenFL 1.8.

theakshaypant:
@psfoley Gotcha, I was sceptical about replacing all `dill` occurrences but the test passing gave me some assurance. 
Thanks for pointing this out, closing this PR and the one opened for the same changes to `develop`.

PR #1242, opened 2025-01-06
Workspaces for different backends of Keras

tanwarsh:
The PR includes the following changes:
  1. Restructured Keras workspaces.
  2. Added NLP workspaces for PyTorch, JAX, and TensorFlow backend with Keras.
  3. Modified template name logic to search for templates in nested directories in openfl-workspaces.


openfl-workspace/keras_jax/nlp/requirements.txt
```diff
@@ -0,0 +1,2 @@
+keras==3.6.0
+jax==0.4.38
```
kminhta, line 2: It doesn't seem like JAX is ever actually explicitly used throughout the experiment

openfl-workspace/keras_torch/nlp/requirements.txt
```diff
@@ -0,0 +1,2 @@
+keras==3.6.0
+torch==2.5.1
```
kminhta, line 2: Similar comment here - it doesn't seem like torch is ever explicitly used in the experiment

openfl/federated/task/runner_keras.py
```diff
@@ -17,6 +19,14 @@
 from openfl.utilities import Metric, TensorKey, change_tags
 from openfl.utilities.split import split_tensor_dict_for_holdouts
 
+# Set the KERAS_BACKEND environment variable based on the available deep learning framework
```
kminhta, line 22: I wonder if it would be better to add a "backend" argument to the keras runner (that could maybe be configured in the `plan.yaml`). Rationale being that a user may have an env that contains multiple frameworks. Idk if that's a super common scenario, but the runner would end up setting the backend to the first one it encounters rather than the intended backend resulting in an explicit error or some unintended consequences

kminhta (commented):
Thanks for getting this started @tanwarsh ! 

My biggest comment here (and I'd like to hear your thoughts on this) is that the JAX/torch workspaces don't seem to showcase JAX or torch. OpenFL offers the ability to define custom training and evaluation loop via `train_` and `validate_` methods of the runner. Keras is able to support custom training loops written in JAX and torch that can potentially preserve a lot of each framework's native API in a similar manner. See [JAX](https://keras.io/guides/writing_a_custom_training_loop_in_jax/) / [Torch](https://keras.io/guides/writing_a_custom_training_loop_in_torch/).

My personal opinion, I think it would be more compelling to showcase OpenFL's framework flexibility in this manner. WDYT?

tanwarsh:
> Thanks for getting this started @tanwarsh !
> 
> My biggest comment here (and I'd like to hear your thoughts on this) is that the JAX/torch workspaces don't seem to showcase JAX or torch. OpenFL offers the ability to define custom training and evaluation loop via `train_` and `validate_` methods of the runner. Keras is able to support custom training loops written in JAX and torch that can potentially preserve a lot of each framework's native API in a similar manner. See [JAX](https://keras.io/guides/writing_a_custom_training_loop_in_jax/) / [Torch](https://keras.io/guides/writing_a_custom_training_loop_in_torch/).
> 
> My personal opinion, I think it would be more compelling to showcase OpenFL's framework flexibility in this manner. WDYT?

I agree with you @kta-intel. The changes in this PR only enable different backends for Keras, but the workspaces do not showcase JAX, Torch, or TensorFlow. It would definitely be compelling to showcase JAX and Torch with Keras. I will work on this.

rahulga1:
> > Thanks for getting this started @tanwarsh !
> > My biggest comment here (and I'd like to hear your thoughts on this) is that the JAX/torch workspaces don't seem to showcase JAX or torch. OpenFL offers the ability to define custom training and evaluation loop via `train_` and `validate_` methods of the runner. Keras is able to support custom training loops written in JAX and torch that can potentially preserve a lot of each framework's native API in a similar manner. See [JAX](https://keras.io/guides/writing_a_custom_training_loop_in_jax/) / [Torch](https://keras.io/guides/writing_a_custom_training_loop_in_torch/).
> > My personal opinion, I think it would be more compelling to showcase OpenFL's framework flexibility in this manner. WDYT?
> 
> I agree with you @kta-intel. The changes in this PR only enable different backends for Keras, but the workspaces do not showcase JAX, Torch, or TensorFlow. It would definitely be compelling to showcase JAX and Torch with Keras. I will work on this.

But will it make sense, to merge this which creates the building block and create a separate PR to bring actual example?

MasterSkepticista (changes requested):
Echoing @kta-intel's comment, three separate tutorials with different backends do not justify the extra lines of code we have to maintain in this repo.

My suggestions, in no particular order:
1. Group tutorials by use-case under `keras`: `keras/{mnist,nlp}`. Backends should be switchable by installing necessary packages (jax, tf or torch). This would go as a multi-backend CI test for OpenFL.
2. Group tutorials by framework; `{jax,torch}`. These will target custom training loops in pure DL framework code, utilizing their respective taskrunner classes. This can be a separate PR.

payalcha:
@tanwarsh These changes seems to impact task_runner_e2e workflow tests. I think changing tests.end_to_end.utils.constants.py enum will work. But please run the E2E test separately to check.

PR #1268, opened 2025-01-14
Recover missing state after collaborator restart

cloudnoize:
After restart the collaborator gets a list of tasks that it needs to finish as part of the current round, that list might be partial as it already executed some tasks prior the restart, previous tasks might have generated state that is needed for the future tasks of this round.
This PR recover or initialized the state of previous tasks enabling the collaborator to resume the execution of the current round.

https://jira.devtools.intel.com/browse/FEDAI-1229

### DIFF
--- a/openfl/federated/task/runner_keras.py
+++ b/openfl/federated/task/runner_keras.py
@@ -182,7 +182,16 @@ class KerasTaskRunner(TaskRunner):
         #  initialization (build_model).
         #  If metrics are added (i.e. not a subset of what was originally
         #  defined) then the model must be recompiled.
-        results = self.model.get_metrics_result()
+        try:
+            results = self.model.get_metrics_result()
+        except ValueError:
+            if "batch_size" in kwargs:
+                batch_size = kwargs["batch_size"]
+            else:
+                batch_size = 1
+            # evaluation needed before metrics can be resolved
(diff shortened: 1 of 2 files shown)


openfl/federated/task/runner_keras.py
```diff
@@ -182,7 +182,16 @@ def train_(self, batch_generator, metrics: list = None, **kwargs):
...
-        results = self.model.get_metrics_result()
+        try:
+            results = self.model.get_metrics_result()
+        except ValueError:
+            if "batch_size" in kwargs:
+                batch_size = kwargs["batch_size"]
+            else:
+                batch_size = 1
+            # evaluation needed before metrics can be resolved
+            self.model.evaluate(self.data_loader.get_valid_loader(batch_size), verbose=1)
```
psfoley, line 193: For other reviewers reference, this is needed for the scenario when the collaborator dies between the `aggregated_model_validation` and `train` task. If the collaborator restarts it will immediately be assigned `train`, and will not be able to resolve the necessary metrics. The names of needed metrics can be resolved by running the `evaluate` function on the model, but these metric values will not be sent to the aggregator.

psfoley (approved):

rahulga1 (approved):

tanwarsh (approved):

PR #1271, opened 2025-01-15
Creating local_state directory, and saving state to it

cloudnoize:


openfl/experimental/workflow/interface/cli/workspace.py
```diff
@@ -54,6 +54,7 @@ def create_dirs(prefix):
     (prefix / "data").mkdir(parents=True, exist_ok=True)  # training data
     (prefix / "logs").mkdir(parents=True, exist_ok=True)  # training logs
     (prefix / "save").mkdir(parents=True, exist_ok=True)  # model weight saves / initialization
+    (prefix / "local_state").mkdir(parents=True, exist_ok=True)  # persistent state
```
teoparvanov: The modules and packages under `openfl/experimental` have to do with Workflow API. In particular, the `FederatedRuntime` uses a different [Aggregator class](https://github.com/securefederatedai/openfl/blob/develop/openfl/experimental/workflow/component/aggregator/aggregator.py#L25), which needs to be additionally updated to persist its local state. 

For now, I suggest not to change anything in this area, and focus on TaskRunner API instead.

teoparvanov (approved):
LGTM, with one comment:

openfl/experimental/workflow/interface/cli/workspace.py
psfoley: Agree - this change should not be made for the workflow API for now

MasterSkepticista (changes requested):
Why not reuse `save`?

cloudnoize:
> Why not reuse `save`?

Save is prepopulated with proto files, which eliminates the option to mount it as an encrypted fs under Gramine.
Furthermore, we need a directory to store the final models that the model owner can retrieve through the file server, and it should not be mounted as an encrypted fs, the ```save``` directory meet those requirements.

MasterSkepticista:
Why not encrypt the state before writing under `save/`?

cloudnoize:
> Why not encrypt the state before writing under `save/`?

Gramine uses proprietary encryption protocol, you can't mount the `save` directory as encrypted fs with files in it and then read them, so it's either you start with those files in a different path and then rewrite them to `save` or present a new directory as done here.
In each case you need a directory that is outside of the encrypted fs to host the final models (which are encrypted but not via Gramine), so the `save` can act both as the initial model and the final model location that is outside of the encrypted fs.

MasterSkepticista:
> > Why not encrypt the state before writing under `save/`?
> 
> Gramine uses proprietary encryption protocol, you can't mount the `save` directory as encrypted fs with files in it and then read them, so it's either you start with those files in a different path and then rewrite them to `save` or present a new directory as done here. In each case you need a directory that is outside of the encrypted fs to host the final models (which are encrypted but not via Gramine), so the `save` can act both as the initial model and the final model location that is outside of the encrypted fs.

I was referring to standard encryption that does not require enclaves

PR #1276, opened 2025-01-16
Remove sysout from 101_MNIST notebook

payalcha:
Install papermill and run the notebook using papermill


teoparvanov (approved):
Awesome, thanks @payalcha ! This way, the notebook doesn't need to be "aware" that it is being tested 😉

noopurintel (approved):

PR #1287, opened 2025-01-17
Fix docker readme wording

psfoley:


psfoley:
Bypassing review due to simple wording change

PR #1294, opened 2025-01-22
PR pipeline restructuring - one workflow to call multiple workflows ...

noopurintel:
.. making it easy to track the jobs under one umbrella, re-run them in single click if required - [an example](https://github.com/securefederatedai/openfl/actions/runs/12902459132)
Quicker release of runner machines as the running jobs get cancelled in case new commit is pushed - [an example](https://github.com/securefederatedai/openfl/actions/runs/12902007132)
Fewer pages on GitHub Actions to scroll through.


.github/workflows/pr_pipeline.yml
```diff
@@ -0,0 +1,147 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.base_ref }}-${{ github.head_ref }}-${{ github.actor }}
+  cancel-in-progress: true
+
+jobs:
+  bandit_code_scan:
```
gbikkiintel, line 13: If none of the jobs depends on other jobs, why do you want to even create separate pr pipeline? instead, you just add status checks to be passed so, no one should be able to merge without passing pipelines

noopurintel: This is what we demoed during our standup - to have all the pipeline jobs under one umbrella for easy tracking and re-runs; also to cancel them once a new commit is pushed.

pasokan-intel (approved):
Look good

payalcha (approved):

teoparvanov (approved):
LGTM, thanks @noopurintel !

PR #1302, opened 2025-01-23
Remove Interactive API

MasterSkepticista:
What is this PR about? < PR title >
What changes have been made? < commit messages >
Tests done < CI status below >


rahulga1:
we should also remove` /openfl/openfl/interface/interactive_api` folder and any references to it.

rahulga1:
> we should also remove` /openfl/openfl/interface/interactive_api` folder and any references to it.

Also I see some of the places in few files where there is a comment it is used only in native api/interactive api. If that is the case, we should clean this dead code also.

MasterSkepticista:
Yes, this will be addressed in future commits. PR is currently in draft (marked ready for CI checks).

payalcha:
@MasterSkepticista Apart from this openfl/plugins/frameworks_adapters/framework_adapter_interface.py  all other plugins are currently used for director and envoy. 
That also need to be deprecated right

Also grpc director_server.py and director_client.py is used only in envoy and director.

MasterSkepticista:
> @MasterSkepticista Apart from this openfl/plugins/frameworks_adapters/framework_adapter_interface.py all other plugins are currently used for director and envoy. That also need to be deprecated right
> 
> Also grpc director_server.py and director_client.py is used only in envoy and director.

Yes, will remove in subsequent commits

rahulga1:
There are still references in following files, Please check if this needs to be taken care
image

MasterSkepticista:
@rahulga1 Resolved until https://github.com/securefederatedai/openfl/pull/1302/commits/d208600561e668fda6163a71ed2241cf1a875369

Could you please check if there are other stale references?

docs/developer_guide/structure/communication.rst
teoparvanov, line 1: Even though we remove the Interactive API layer, we still use the Director/Envoy infra for the FederatedRuntime. IMO this document is still valid, even though the FederatedRuntime uses a separate (experimental) package of core components:
https://github.com/securefederatedai/openfl/tree/develop/openfl/experimental/workflow/component

docs/developer_guide/structure/components.rst
teoparvanov, line 1: same here, let's keep this as is

teoparvanov (approved):
LGTM overall, with a couple of comments/suggestions to consider:

docs/developer_guide/structure/communication.rst
MasterSkepticista: Agree. However, I think it is best to create new documentation (incl pages/sections) solely with workflow API in mind. We ideally do not want to tell the users that Federated Runtime uses "Director Envoy" under the hood, as it is an implementation detail and does not help the user in any way (I mean the fact of its former roots with interactive API).

We can have a dedicated `Experimental Features` section of the documentation.

teoparvanov: I think this makes sense, thanks for the clarification, @MasterSkepticista ! This PR should be good to go then (once you've addressed the remaining comments by Rahul).

rahulga1 (approved):
LGTM

noopurintel (approved):

kminhta (approved):
LGTM, thanks!

PR #1303, opened 2025-01-23
introduce a mechanism for relearning the aggregator certificate

yontyon:
In reality, the aggregator may restart and come up with new TLS certificate. For that, it is not enough to retry on the wrapper/interceptor level, but we also need to fetch the latest cert.
This PR introduces a mechanism for:
1. registering a callback for fetching the aggregator's latest cert
2. instead of intercepting and retrying on the channel itself (which prevents us from refreshing it), we retry on the wrapper level

### DIFF
--- a/openfl/transport/grpc/aggregator_client.py
+++ b/openfl/transport/grpc/aggregator_client.py
@@ -148,15 +148,15 @@ def _resend_data_on_reconnection(func):
         while True:
             try:
                 response = func(self, *args, **kwargs)
+                break
             except grpc.RpcError as e:
-                if e.code() == grpc.StatusCode.UNKNOWN:
-                    self.logger.info(
-                        f"Attempting to resend data request to aggregator at {self.uri}"
-                    )
-                elif e.code() == grpc.StatusCode.UNAUTHENTICATED:
-                    raise
-                continue
-            break
+                self.logger.info(
+                    f"Failed to send data request to aggregator at {self.uri}, error code {e.code()}"
+                )
+                if self.refetch_server_cert_callback is not None:
+                    self.logger.info("Refetching server certificate")
+                    self.root_certificate = self.refetch_server_cert_callback()
+                self.sleeping_policy.sleep()
         return response
 
     return wrapper
@@ -197,6 +197,7 @@ class AggregatorGRPCClient:
         aggregator_uuid=None,
         federation_uuid=None,
         single_col_cert_common_name=None,
+        refetch_server_cert_callback=None,
         **kwargs,
     ):
         """
@@ -226,7 +227,7 @@ class AggregatorGRPCClient:
         self.root_certificate = root_certificate
         self.certificate = certificate
         self.private_key = private_key
-
+        self.sleeping_policy = ConstantBackoff(int(kwargs.get("client_reconnect_interval", 1)), getLogger(__name__), self.uri)
         self.logger = getLogger(__name__)
 
         if not self.use_tls:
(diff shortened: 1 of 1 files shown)


openfl/transport/grpc/aggregator_client.py
```diff
@@ -377,12 +365,10 @@ def reconnect(self):
...
         self.logger.debug("Connecting to gRPC at %s", self.uri)
 
-        self.stub = aggregator_pb2_grpc.AggregatorStub(
-            grpc.intercept_channel(self.channel, *self.interceptors)
-        )
+        self.stub = aggregator_pb2_grpc.AggregatorStub(self.channel)
 
-    @_atomic_connection
     @_resend_data_on_reconnection
+    @_atomic_connection
```
teoparvanov, line 373: Why do you need to rearrange the decorators?

openfl/transport/grpc/aggregator_client.py
```diff
@@ -245,21 +246,8 @@ def __init__(
...
-                    logger=self.logger,
-                    reconnect_interval=int(kwargs.get("client_reconnect_interval", 1)),
-                    uri=self.uri,
-                ),
-                status_for_retry=(grpc.StatusCode.UNAVAILABLE,),
-            ),
-        )
-        self.stub = aggregator_pb2_grpc.AggregatorStub(
-            grpc.intercept_channel(self.channel, *self.interceptors)
-        )
```
teoparvanov, line 262: What's the reasoning behind removing the this interceptor? I think the gRPC connection can generally benefit from retries (f.e. due to temporary network disconnects unrelated to certificates).

PS : on a side note, wouldn't it be more elegant if you extended the `RetryOnRpcErrorClientInterceptor` to also call the certificate callback?

teoparvanov (commented):
Hi @yontyon , I have a couple of comments from my first review pass:

openfl/transport/grpc/aggregator_client.py
yontyon: I want to reconnect on every retry.
Before this change, we reconnect and then loop until we successfully send the message. However, in the case of TLS exchange, we need to loop on the reconnection itself

openfl/transport/grpc/aggregator_client.py
yontyon: As far as I understand, the interceptor works on  a given channel - however, if we need to fetch new certs we also need to create a new channel and this is why intercepting the call won't work here

teoparvanov: OK, this is true for the case when you have to re-establish the channel. But I don't understand why you need to remove the retrying interceptor. 

Doesn't it add resilience in case of temporary network delays/disconnects within an existing channel where the TLS cert hasn't changed?

yontyon: We can have a more fine-tuned error handling and differentiate between errors that are due to a bad TLS handshake (on which we want to refresh the connection) and those that don't require refreshing the certificates.  However, I do think that for the current stage, it is better to take the extreme approach (as we don't have much communication in any case), we can fine-tune it in the future (for example, I saw that on TLS error the error is also UNAVAILABLE, so we can try and refine it but I don't think this should be the focus right now)

teoparvanov: I think I understand now. You are suggesting the retry to happen at the "application" level, in the client's request loop rather than at the lower gRPC level:
```python
        while True:
            try:
                response = func(self, *args, **kwargs)
                break
            except grpc.RpcError as e:
                # try fetching a new certificate in case of a gRPC error
```

PS: AFAIU, on each retry, a new gRPC channel is established (incl. re-establishing the underlying TLS connection)?

openfl/transport/grpc/aggregator_client.py
```diff
@@ -148,15 +148,15 @@ def wrapper(self, *args, **kwargs):
         while True:
             try:
                 response = func(self, *args, **kwargs)
+                break
```
teoparvanov, line 151: I guess this `break` statement replaces the `continue` from the deleted code block below?

rahulga1 (approved):

teoparvanov (approved):

PR #1307, opened 2025-01-24
Build base image on `develop`

MasterSkepticista:
This is to be in-line with new dockerization steps. This method will be replaced with a simpler one in a subsequent PR.


teoparvanov (approved):
Thanks for fixing this, @MasterSkepticista ! I tried it locally, and it worked.

PR #1322, opened 2025-01-30
Nested Workspaces template

tanwarsh:
Changes required to recognise workspace in nested workspaces.
for example: Keras related workspace can be under keras folder in openFL workspace directory.


rahulga1 (approved):

docs/developer_guide/advanced_topics/log_metric_callback.rst
```diff
@@ -95,9 +95,9 @@ Follow the steps below to write your custom callback function instead. As an exa
     defaults : plan/defaults/aggregator.yaml
     template : openfl.component.Aggregator
     settings :
-      init_state_path     : save/torch_cnn_mnist_init.pbuf
-      best_state_path     : save/torch_cnn_mnist_best.pbuf
-      last_state_path     : save/torch_cnn_mnist_last.pbuf
+      init_state_path     : save/torch/mnist_init.pbuf
```
ishaileshpant: should we mandate creation of additional dir in save folder in each workspace now due to this refactor - this potentially can impact on consumption / user side cause model save paths are now changed 

Suggest: just rename to use torch_mnist_ as prefix (and as applicable per WS)

ishaileshpant: Though this is just in doc, it gives an impression that the plan.yaml has impacts as per this refactor which AFAIK is not the case

ishaileshpant (approved):
LGTM apart from small nit to change some docs

kminhta (approved):
LGTM

theakshaypant (approved):

docs/developer_guide/advanced_topics/log_metric_callback.rst
tanwarsh: Thanks for pointing this out. resolved.

PR #1325, opened 2025-01-30
[documentation] Shrink module/class headers

payalcha:
Link post changes - https://payalcha-openfl.readthedocs.io/en/latest/ 

What we changed -
All custom module will be prefixed with  **Module -** 

image 

All Classes will be prefixed with **Class -**

image

All Functions will be prefixed with **Function-**
image


MasterSkepticista (changes requested):
Thanks for volunteering for this change. Beyond our offline discussion, I have two suggestions:
* Can we remove `Class` word in class templates (module is ok though)? Feels cleaner IMO.
* How about disabling the "sidebar" when viewing a class/function definition? That will free up real estate even for expanded definitions written below the title.

P.S.: Verify that `function` pages do not show `Class` or `Module` word.

payalcha:
> Thanks for volunteering for this change. Beyond our offline discussion, I have two suggestions:
> 
> * Can we remove `Class` word in class templates (module is ok though)? Feels cleaner IMO.
> * How about disabling the "sidebar" when viewing a class/function definition? That will free up real estate even for expanded definitions written below the title.
> 
> P.S.: Verify that `function` pages do not show `Class` or `Module` word.

@MasterSkepticista Currently module name and class names are very similar. So I feel its better to have Class/Module/Function in title.
SideBar - We can keep it for now.

MasterSkepticista:
@payalcha I think this is in a good shape to merge given all other constraints discussed offline. Is this ready to merge?

payalcha:
> @payalcha I think this is in a good shape to merge given all other constraints discussed offline. Is this ready to merge?

From our end we had done all the changes.

rahulga1 (approved):

PR #1333, opened 2025-02-03
Pytorch Taskrunner Workspace with Keras 3

tanwarsh:
Pytorch Taskrunner Workspace with Keras 3, showcasing custom model and training and testing loops

Screenshot 2025-02-04 at 11 35 58 AM


kminhta (commented):
Thanks @tanwarsh ! Some open ended comments/questions:
1. We should consider having this one directory down, either under `keras/pytorch` workspace or under `pytorch/keras`. This would simply be a way to further indicate that is either (1) an extended functionality of the native keras taskrunner or (2) a special case for pytorch that uses the keras taskrunner.
2. Is there a reason you decided to go with a customized `fit()` function as opposed to a [custom training loop](https://keras.io/guides/writing_a_custom_training_loop_in_torch/)? This way is not wrong, of course, but the custom training loop seems like a more natural addition to the task runner since a dev would be expected to modify the `train_` and `validate` functions, whereas inheriting `keras.Model` and then modifying `train_step` and `test_step` may be confusing. If we stick to this, I suggest we move the model definition to a separate file then import it in `taskrunner.py`. Also, even if we stick with this, I think we should also add in the a workspace that takes advantage of the custom training loop (it can be added as backlog item to be added in a separate PR)

openfl-workspace/torch/mnist_with_keras/plan/plan.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+assigner :
+  defaults : plan/defaults/assigner.yaml
+
+tasks :
+  defaults : plan/defaults/tasks_keras.yaml
+
+compression_pipeline :
+  defaults : plan/defaults/compression_pipeline.yaml
+  # To use different Compression Pipeline, uncomment the following lines
```
kminhta: We an remove these comments since we have dedicated workspaces / docs for compression pipeline

openfl-workspace/torch/mnist_with_keras/src/dataloader.py
```diff
@@ -0,0 +1,47 @@
...
+        Initialize.
+
+        Args:
+            data_path: File path for the dataset
+            batch_size (int): The batch size for the data loader
+            **kwargs: Additional arguments, passed to super init and load_mnist_shard
+        """
+        super().__init__(batch_size, **kwargs)
+
+        # TODO: We should be downloading the dataset shard into a directory
```
kminhta: We can remove these TODOs and add it as a backlog item. FYI, I don't think it's entirely necessary to download the dataset shard into a directory for the  `--template` workspaces, but automatic sharding based on rank / size of collab list could be interesting

openfl-workspace/torch/mnist_with_keras/src/taskrunner.py
```diff
@@ -0,0 +1,179 @@
+# Copyright (C) 2020-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""You may copy this file as the starting point of your own model."""
+import torch
+import keras
+
+from openfl.federated import KerasTaskRunner
+
+class CNNModel(keras.Model):
```
kminhta: from my previous comment, we should move this to a separate file then import it

openfl-workspace/torch/mnist_with_keras/plan/plan.yaml
tanwarsh: removed

openfl-workspace/torch/mnist_with_keras/src/dataloader.py
tanwarsh: removed

openfl-workspace/torch/mnist_with_keras/src/taskrunner.py
tanwarsh: done

tanwarsh:
> Thanks @tanwarsh ! Some open ended comments/questions:
> 
> 1. We should consider having this one directory down, either under `keras/pytorch` workspace or under `pytorch/keras`. This would simply be a way to further indicate that is either (1) an extended functionality of the native keras taskrunner or (2) a special case for pytorch that uses the keras taskrunner.
> 2. Is there a reason you decided to go with a customized `fit()` function as opposed to a [custom training loop](https://keras.io/guides/writing_a_custom_training_loop_in_torch/)? This way is not wrong, of course, but the custom training loop seems like a more natural addition to the task runner since a dev would be expected to modify the `train_` and `validate` functions, whereas inheriting `keras.Model` and then modifying `train_step` and `test_step` may be confusing. If we stick to this, I suggest we move the model definition to a separate file then import it in `taskrunner.py`. Also, even if we stick with this, I think we should also add in the a workspace that takes advantage of the custom training loop (it can be added as backlog item to be added in a separate PR)

Thank you @kta-intel, for the review.
1. I have changed the directory structure as per your comment.
2. Both methods are ways to write custom training and testing loops. I found this approach to be cleaner and more readable. I have moved it to a separate file. We can discuss this offline and have one workspace that showcase custom training loop same as https://keras.io/guides/writing_a_custom_training_loop_in_torch/.

kminhta:
> 2. Both methods are ways to write custom training and testing loops. I found this approach to be cleaner and more readable.

From my perspective, they achieve different things: the custom `.fit()` function requires modification of `keras.Model` and makes heavy use of Keras API whereas the custom training loop can be achieved by modifying `openfl.federated.KerasTaskRunner` and better resembles how a torch training/validation loop would be implemented by a standard torch user and enables them to use native torch loader, optimizer, and loss function. It works well in a scenario where a user has a Keras model but, has custom optimizers / loss functions written in torch. It's a bit of a niche scenario, but since we are dealing with distributed collaborative training it's not out of the realm of possibility.

More broadly, I am also thinking about general framework support (claiming limited support for JAX/TF via Keras has a stronger case when we maximize usage of their native APIs) and how to limit the number of workspaces to "critical" examples so it does not get too bloated.

tanwarsh:
> > 2. Both methods are ways to write custom training and testing loops. I found this approach to be cleaner and more readable.
> 
> From my perspective, they achieve different things: the custom `.fit()` function requires modification of `keras.Model` and makes heavy use of Keras API whereas the custom training loop can be achieved by modifying `openfl.federated.KerasTaskRunner` and better resembles how a torch training/validation loop would be implemented by a standard torch user and enables them to use native torch loader, optimizer, and loss function. It works well in a scenario where a user has a Keras model but, has custom optimizers / loss functions written in torch. It's a bit of a niche scenario, but since we are dealing with distributed collaborative training it's not out of the realm of possibility.
> 
> More broadly, I am also thinking about general framework support (claiming limited support for JAX/TF via Keras has a stronger case when we maximize usage of their native APIs) and how to limit the number of workspaces to "critical" examples so it does not get too bloated.

Noted. I will address this in a separate PR if that is acceptable.

openfl/federated/data/__init__.py
MasterSkepticista, line 1: Nit: You may remove these changes. Both this and JAX PR do these deletions.

tanwarsh: done

kminhta (approved):

teoparvanov (approved):

PR #1334, opened 2025-02-03
JAX Taskrunner Workspace with Keras 3

tanwarsh:
JAX Taskrunner Workspace with Keras 3, showcasing custom model and training and testing loops
Screenshot 2025-02-04 at 12 03 37 PM


openfl/federated/task/__init__.py
```diff
@@ -4,18 +4,18 @@
...
 
 from openfl.federated.task.runner import TaskRunner  # NOQA
 
 if util.find_spec("keras") is not None:
+    if util.find_spec("torch") is not None:
+        os.environ["KERAS_BACKEND"] = "torch"
+    elif util.find_spec("jax") is not None:
+        os.environ["KERAS_BACKEND"] = "jax"
+    elif util.find_spec("tensorflow") is not None:
+        os.environ["KERAS_BACKEND"] = "tensorflow"
```
MasterSkepticista: Ok, this part I did not understand. The purpose of `KERAS_BACKEND` is for a user to declare it. This code instead checks which backend is available and sets it. I think it is better to only check for `keras`, and let the user deal with the backend missing error when they actually execute code.

openfl-workspace/keras/jax/mnist/requirements.txt
```diff
@@ -0,0 +1,3 @@
+jax==0.5.0
+keras==3.8.0
+tensorflow==2.18.0
```
MasterSkepticista, line 3: tensorflow required for saving?

openfl-workspace/keras/jax/mnist/src/dataloader.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) 2020-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""You may copy this file as the starting point of your own model."""
+
+from openfl.federated import KerasDataLoader
+from .mnist_utils import load_mnist_shard
+
+
+class JAXMNISTInMemory(KerasDataLoader):
```
MasterSkepticista, line 10: We are duplicating this exact class innumerable times. I am concerned about the maintenance effort that comes with this.

What if we simply provide built-in MNIST dataloader under `openfl.datasets` and call it everywhere in the framework/tutorials?

MasterSkepticista (changes requested):

openfl-workspace/keras/jax/mnist/requirements.txt
tanwarsh: yes, It is better to use tensorflow then onnx and we are only using it to save/export model

openfl/federated/task/__init__.py
tanwarsh: Thank you for the review @MasterSkepticista, 

Let me explain why this is needed:

1. If we do not set KERAS_BACKEND and import Keras, it will default to using TensorFlow as the backend even if we set KERAS_BACKEND later.
2. The code in runner_keras.py will import Keras before user code is executed, causing an error if the user is using a custom model and training/testing loop with JAX or PyTorch.
3. If user is not using custom training or testing loop using jax or pyTorch then it will always use tensorflow.

This code runs before any import, sets the KERAS_BACKEND variable, and then imports the Keras library.

This is a strange behavior from Keras, but it is how it works. I also removed the TensorFlow import because we are not using it, and it sets the backend to TensorFlow even if we set the KERAS_BACKEND variable later.

if this is still not clear let me know.

openfl-workspace/keras/jax/mnist/src/dataloader.py
tanwarsh: I agree with you but as the same dataset is already used in other workspaces and it can be extended to other datasets which are used multiple times, If its fine we can take this as a separate task.

openfl/federated/task/__init__.py
MasterSkepticista: I understand your reasoning behind defaults. But we need to remember that `keras` backends cannot change after import. 

The issue I see with this approach is, `torch` will be immediately set as the backend even if the user wants `jax` or `tf` for their experiment - if `torch` is installed in the python environment. Same reasoning applies for `jax`.

It is actually common of users to hold multiple DL frameworks within a python environment.

My recommendation would be to have a section in our documentation for multi-backend `keras`, and mandating users to `export KERAS_BACKEND=<value>` before they run any OpenFL code/commands. This is inline with what `keras` requires all users to do.

Open to discuss alternatives.

tanwarsh: I have removed the code changes and updated documentation. Please review again.

docs/about/features_index/taskrunner.rst
```diff
@@ -162,6 +162,11 @@ STEP 1: Create a Workspace
  - :code:`tf_cnn_histology`: a workspace with a simple `TensorFlow <http://tensorflow.org>`__ CNN model that will download the `Colorectal Histology <https://zenodo.org/record/53169#.XGZemKwzbmG>`_ dataset and train in a federation.
  - :code:`keras/histology`: a workspace with a simple `PyTorch <http://pytorch.org/>`__ CNN model that will download the `Colorectal Histology <https://zenodo.org/record/53169#.XGZemKwzbmG>`_ dataset and train in a federation.
  - :code:`torch/mnist`: a workspace with a simple `PyTorch <http://pytorch.org>`__ CNN model that will download the `MNIST <http://yann.lecun.com/exdb/mnist/>`_ dataset and train in a federation.
+ - :code:`keras/jax/mnist`: a workspace with a simple `Keras <http://keras.io/>`__ CNN model that will download the `MNIST <http://yann.lecun.com/exdb/mnist/>`_ dataset and train in a federation with jax as backend. You can export the environment variable KERAS_BACKEND to configure your backend. Available backend options are: "jax", "tensorflow", "torch". Example:
+
+     .. code-block:: shell
+
+       $ export KERAS_BACKEND="jax"
```
MasterSkepticista, line 169: "Please ensure `KERAS_BACKEND` is set in the environment where you plan on using OpenFL before executing any `fx` command." after the code block.

MasterSkepticista (approved):

docs/about/features_index/taskrunner.rst
theakshaypant, line 169: Second @MasterSkepticista 's comment, it is not very clear, when (before/after which command) the env var should be declared.

tanwarsh: resolved

tanwarsh: resolved

kminhta (approved):

teoparvanov (approved):

PR #1343, opened 2025-02-05
Read tensor.db for bestscore

payalcha:
1. Change torch/mnist plan to get tensor.db in local_state.
This change is done in continuation of Eran's creation of tensor.db file. For keras/mnist it was creating in local_state and torch/mnist was getting created in workspace of aggregator.
3. Change summary_helper to get details from tensor.db instead of metrics.
4. Use defused etree for XMLParser.
5. **Documentation** change - Instead of Exceptions write Exception

### DIFF
--- a/tests/end_to_end/test_suites/tr_with_fedeval_tests.py
+++ b/tests/end_to_end/test_suites/tr_with_fedeval_tests.py
@@ -11,7 +11,7 @@ from tests.end_to_end.utils.tr_common_fixtures import (
 )
 from tests.end_to_end.utils import federation_helper as fed_helper
 from tests.end_to_end.utils.tr_workspace import create_tr_workspace, create_tr_dws_workspace
-from tests.end_to_end.utils.summary_helper import get_aggregated_accuracy
+from tests.end_to_end.utils.summary_helper import get_best_accuracy
 
 log = logging.getLogger(__name__)
 
@@ -37,9 +37,9 @@ def test_eval_federation_via_native(request, fx_federation_tr):
 
     # Set the best model path in request. It is used during plan initialization for evaluation step
     request.config.best_model_path = os.path.join(fx_federation_tr.aggregator.workspace_path, "save", "best.pbuf")
-    metric_file_path = os.path.join(fx_federation_tr.aggregator.workspace_path, "logs", "aggregator_metrics.txt")
-    model_accuracy = get_aggregated_accuracy(metric_file_path)
-    log.info(f"Model accuracy post {request.config.num_rounds} rounds: {model_accuracy}")
+    tensor_db_file_path = os.path.join(fx_federation_tr.aggregator.workspace_path, "local_state", "tensor.db")
+    best_model_accuracy = get_best_accuracy(tensor_db_file_path)
+    log.info(f"Model accuracy post {request.config.num_rounds} rounds: {best_model_accuracy}")
     # Create new workspace with evaluation scope
     new_fed_obj = create_tr_workspace(request, eval_scope=True)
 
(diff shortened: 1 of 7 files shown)


noopurintel (approved):
Suggested a few minor changes offline. Please do those before merging.
Approved.

tests/end_to_end/utils/summary_helper.py
```diff
@@ -32,6 +33,25 @@ def initialize_xml_parser():
...
+        print(f"Database file {database_file} not found. Cannot get best accuracy")
+        return best_accuracy
+
+    db_helper = DBHelper(database_file)
+    round_number, best_score = db_helper.read_key_value_store()
+    print(f"Best accuracy: {best_score} is in round_number {round_number} ")
+    return best_accuracy
+
+
 def get_aggregated_accuracy(agg_log_file):
```
noopurintel: Please remove this function as it is no longer required.

tests/end_to_end/utils/summary_helper.py
```diff
@@ -10,6 +10,7 @@
 
 import tests.end_to_end.utils.constants as constants
 from tests.end_to_end.utils.generate_report import convert_to_json
```
noopurintel: This import statement can be removed along with get_aggregated_accuracy.

noopurintel (changes requested):

tests/end_to_end/utils/summary_helper.py
```diff
@@ -32,6 +33,25 @@ def initialize_xml_parser():
...
+        database_file: the database file
+    Returns:
+        best_accuracy: the best accuracy
+    """
+    best_accuracy = "Not Found"
+    if not os.path.exists(database_file):
+        print(f"Database file {database_file} not found. Cannot get best accuracy")
+        return best_accuracy
+
+    db_helper = DBHelper(database_file)
```
noopurintel: Please rename the variable name from `db_helper` to say `db_obj` or `db_helper_obj`, because the file name is also `db_helper.py` and when imported at some place it would be as `db_helper` mostly - so we better avoid this confusion.

noopurintel (changes requested):

tests/end_to_end/test_suites/tr_with_fedeval_tests.py
```diff
@@ -99,9 +100,9 @@ def test_eval_federation_via_dockerized_workspace(request, fx_federation_tr_dws)
...
-    new_metric_file_path = os.path.join(new_fed_obj.aggregator.workspace_path, "logs", "aggregator_metrics.txt")
-    model_accuracy_eval = get_aggregated_accuracy(new_metric_file_path)
-    log.info(f"Model accuracy during evaluation only on prev trained model is : {model_accuracy_eval})")
+    tensor_db_file_path = os.path.join(new_fed_obj.aggregator.workspace_path, "local_state", "tensor.db")
+    best_model_accuracy_eval = get_best_accuracy(tensor_db_file_path)
+    log.info(f"Model accuracy post {request.config.num_rounds} rounds: {best_model_accuracy}")
 
     # verify that the model accuracy is similar to the previous model accuracy max of 1% difference
-    assert abs(model_accuracy - model_accuracy_eval) <= 0.01, "Model accuracy is not similar to the previous model accuracy"
+    assert abs(best_model_accuracy - best_model_accuracy_eval) <= 0.01, "Model accuracy is not similar to the previous model accuracy"
```
MasterSkepticista: 1% is significant in data science. If logically the expectation is to get same value under floating point error, this tolerance should be around 1e-5.

MasterSkepticista (changes requested):
At all relevant points in the code, replace `accuracy` with `metric` - accuracy is one of thousands of metrics that OpenFL can be used for.

tests/end_to_end/utils/summary_helper.py
payalcha: Done

tests/end_to_end/utils/summary_helper.py
payalcha: Done

tests/end_to_end/utils/summary_helper.py
payalcha: done

tests/end_to_end/test_suites/tr_with_fedeval_tests.py
payalcha: Done. Changed tolerance to 0.00001.

noopurintel (approved):

rajithkrishnegowda (approved):

PR #1346, opened 2025-02-06
Task Runner API: Secure Aggregation

theakshaypant:
## Introduction
This PR introduces [secure aggregation](https://eprint.iacr.org/2017/281.pdf) to Task Runner API.

**Would specially apprecaite comments on how tensors are exchanged between the participants.**

## Changes
- Secure aggregation on client side is implemented as [callback](https://github.com/securefederatedai/openfl/blob/develop/openfl/callbacks/callback.py).
- This callback is run `on_experiment_begin`.
- Collaborator side steps are implemented using `CollaboratorSecAgg` callback.
- Aggregator side steps are implemented using `openfl.utilities.secagg.Setup` whcih performs "aggregation" of the tensors sent by the collaborators..
- `send_local_task_results` is modfied to save secure aggregation setup stage tensors to aggregator tensor db. 
- Masks are added to "metric"s when a task is completed before sending the result to the aggregator.
- `SecureAggregation` is used for "metric"s when enabled instead of `WeightedAverage`.
- Added a `secure_aggregation: bool` flag to the plan under
  - `aggregator.settings`
  - `collaborator.settings`

## NOTE
- Does not cover dropouts, will be handled as part of SecAgg+ integration in future PRs.
- In this implementation, there is a 60 second wait period when collaborator tries to fetch keys from the aggregator, fails if there is no response during this time. 
  - This time can be increased or further retries can be added. 

## Testing
Edit: With 574c102647169de56f4b66452640086e7944ed99
- Aggregator logs
```
           INFO     🧿 Starting the Aggregator Service.                                                                                         aggregator.py:94
           INFO     Building `openfl.component.RandomGroupedAssigner` Module.                                                                        plan.py:235
           INFO     Building `openfl.pipelines.NoCompressionPipeline` Module.                                                                        plan.py:235
           INFO     Importing 🡆 Object CutoffTimePolicy from openfl.component.aggregator.straggler_handling Module.                                  plan.py:259
           INFO     Building `openfl.component.Aggregator` Module.                                                                                   plan.py:235
           WARNING  CutoffTimePolicy is disabled as straggler_cutoff_time is set to np.inf.                                             straggler_handling.py:87
           INFO     Persistent checkpoint is enabled, setting persistent db at path local_state/tensor.db                                      aggregator.py:150
           INFO     Initializing persistent db at local_state/tensor.db                                                                      persistent_db.py:36
           INFO     Starting Aggregator gRPC Server                                                                                     aggregator_server.py:366
[23:54:00] INFO     Sending tasks to collaborator charlie for round 0                                                                          aggregator.py:531
[23:54:01] INFO     Collaborator charlie is sending task results for aggregated_model_validation, round 0                                      aggregator.py:772
[23:54:05] INFO     SecAgg: recreated secrets successfully                                                                                          setup.py:223
           INFO     SecAgg: setup completed, saved required tensors to db.                                                                          setup.py:271
           INFO     Sending tasks to collaborator bob for round 0                                                                              aggregator.py:531
[23:54:06] INFO     Collaborator charlie is sending task results for train, round 0                                                            aggregator.py:772
           INFO     Collaborator bob is sending task results for aggregated_model_validation, round 0                                          aggregator.py:772
[23:54:07] INFO     Collaborator charlie is sending task results for locally_tuned_model_validation, round 0                                   aggregator.py:772
           INFO     Round 0: Collaborators that have completed all tasks: ['charlie']                                                         aggregator.py:1231
[23:54:11] INFO     Collaborator bob is sending task results for train, round 0                                                                aggregator.py:772
[23:54:12] INFO     Collaborator bob is sending task results for locally_tuned_model_validation, round 0                                       aggregator.py:772
           INFO     Round 0: Collaborators that have completed all tasks: ['charlie', 'bob']                                                  aggregator.py:1231
           INFO     Round 0: saved the best model with score 0.075954                                                                         aggregator.py:1138
           INFO     Saving round 0 model...                                                                                                   aggregator.py:1176
           INFO     Committed model for round 0, saved 8 model tensors and 26 next round model tensors  with best_score 0.07595351523210098 persistent_db.py:220
           INFO     Persist model and clean task result for round 0                                                                            aggregator.py:394
           INFO     Starting round 1...                                                                                                       aggregator.py:1189
           INFO     Sending tasks to collaborator bob for round 1                                                                              aggregator.py:531
           INFO     Collaborator bob is sending task results for aggregated_model_validation, round 1                                          aggregator.py:772
[23:54:16] INFO     Collaborator bob is sending task results for train, round 1                                                                aggregator.py:772
[23:54:17] INFO     Sending tasks to collaborator charlie for round 1                                                                          aggregator.py:531
           INFO     Collaborator bob is sending task results for locally_tuned_model_validation, round 1                                       aggregator.py:772
           INFO     Round 1: Collaborators that have completed all tasks: ['bob']                                                             aggregator.py:1231
           INFO     Collaborator charlie is sending task results for aggregated_model_validation, round 1                                      aggregator.py:772
[23:54:27] INFO     Collaborator charlie is sending task results for train, round 1                                                            aggregator.py:772
           INFO     Collaborator charlie is sending task results for locally_tuned_model_validation, round 1                                   aggregator.py:772
           INFO     Round 1: Collaborators that have completed all tasks: ['bob', 'charlie']                                                  aggregator.py:1231
           INFO     Round 1: saved the best model with score 0.935753                                                                         aggregator.py:1138
           INFO     Saving round 1 model...                                                                                                   aggregator.py:1176
           INFO     Committed model for round 1, saved 8 model tensors and 26 next round model tensors  with best_score 0.9357534888987689  persistent_db.py:220
           INFO     Persist model and clean task result for round 1                                                                            aggregator.py:394
           INFO     Experiment Completed. Cleaning up...                                                                                      aggregator.py:1187
           INFO     Sending signal to collaborator charlie to shutdown...                                                                      aggregator.py:478
[23:54:37] INFO     Sending signal to collaborator bob to shutdown...                                                                          aggregator.py:478

 ✔️ OK
```
- Documentation 
Secure Aggregation — OpenFL documentation


setup.py
```diff
@@ -94,6 +94,7 @@ def run(self):
         'tensorboardX',
         'protobuf>=4.22,<6.0.0',
         'grpcio>=1.56.2,<1.66.0',
+        'pycryptodome'
```
payalcha, line 97: If we keep this as setup, even when secure aggregation not needed by customer, they need to install it.

openfl/utilities/secagg/crypto.py
```diff
@@ -0,0 +1,154 @@
+# Copyright 2020-2025 Intel Corporation
```
payalcha: There is cryptography package in openfl. Isn't secagg packages more suitable inside cryptography.

openfl/utilities/secagg/crypto.py
```diff
@@ -0,0 +1,154 @@
...
+        random.seed(seed)
+        seed = random.random()
+
+    # Seed numpy random generator.
+    np.random.seed(seed)
+
+    return np.random.random()
+
+
+def calculate_shared_mask(agreed_keys: list):
```
payalcha: Docstring

theakshaypant: Resolved in 31feb1b16b181aa1466ba31facc4f37cd539e767.

openfl-workspace/keras/mnist_secagg/plan/plan.yaml
```diff
@@ -0,0 +1,48 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
rahulga1: change the year everywhere applicable.

openfl-workspace/keras/mnist_secagg/plan/plan.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+assigner :
+  defaults : plan/defaults/assigner.yaml
+
+tasks :
+  defaults : plan/defaults/tasks_keras.yaml
+
+compression_pipeline :
+  defaults : plan/defaults/compression_pipeline.yaml
+  # To use different Compression Pipeline, uncomment the following lines
```
rahulga1: remove unnecessary comments which were in template from all the places.

openfl-workspace/keras/mnist_secagg/src/mnist_utils.py
```diff
@@ -0,0 +1,118 @@
+# Copyright (C) 2020-2021 Intel Corporation
```
rahulga1: is it not possible to refer the original file, if there is no sec_agg specific changes in here?
If not possible, we can change name to utils.py, mnist is already in the path.

openfl-workspace/keras/mnist_secagg/plan/plan.yaml
tanwarsh: we can remove these comments.

openfl-workspace/keras/mnist_secagg/src/dataloader.py
```diff
@@ -0,0 +1,47 @@
...
+        Initialize.
+
+        Args:
+            data_path: File path for the dataset
+            batch_size (int): The batch size for the data loader
+            **kwargs: Additional arguments, passed to super init and load_mnist_shard
+        """
+        super().__init__(batch_size, **kwargs)
+
+        # TODO: We should be downloading the dataset shard into a directory
```
tanwarsh: remove this.

openfl/component/collaborator/collaborator.py
```diff
@@ -168,6 +175,18 @@ def set_available_devices(self, cuda: Tuple[str] = ()):
     def run(self):
         """Run the collaborator."""
         # Experiment begin
+
+        # FIXME: Not working when added to callbacks on line 157.
```
theakshaypant: @MasterSkepticista Need your input on this.

tanwarsh: please fix this or remove the comment if no longer required.

openfl/component/aggregator/aggregator.py
```diff
@@ -199,13 +202,20 @@ def __init__(
             self._load_initial_tensors()  # keys are TensorKeys
 
         self.collaborator_tensor_results = {}  # {TensorKey: nparray}}
+        self._secure_aggregation_enabled = secure_aggregation
 
         # Callbacks
+        self._secure_aggregation_enabled = True
```
rahulga1: why this is set True by default?, and isnt it overriding the value from previous line?

openfl/component/aggregator/aggregator.py
```diff
@@ -199,13 +202,20 @@ def __init__(
...
+        self.secagg = secagg_setup(self.uuid, self.authorized_cols, self.tensor_db)
```
rahulga1: shouldnt we enable only when sec agg is enabled?

openfl/component/aggregator/aggregator.py
```diff
@@ -1045,13 +1089,19 @@ def _compute_validation_related_task_metrics(self, task_name) -> dict:
...
-            assert collaborators_for_task[0] in tags, (
-                f"Tensor {tensor_key} in task {task_name} has not been processed correctly"
-            )
+            assert (
+                collaborators_for_task[0] in tags
+            ), f"Tensor {tensor_key} in task {task_name} has not been processed correctly"
             # Strip the collaborator label, and lookup aggregated tensor
             new_tags = change_tags(tags, remove_field=collaborators_for_task[0])
             agg_tensor_key = TensorKey(tensor_name, origin, round_number, report, new_tags)
             agg_function = WeightedAverage() if "metric" in tags else task_agg_function
```
rahulga1: why this is not in if-else, as at a time you will be using one aggregation algorithm.

openfl/component/aggregator/aggregator.py
```diff
@@ -1202,3 +1252,45 @@ def stop(self, failed_collaborator: str = None) -> None:
...
+            if "secagg" not in tuple(named_tensor.tags):
+                continue
+            else:
+                secagg_setup = True
+                # Process and save tensor to local tensor db.
+                self._process_named_tensor(named_tensor, collaborator_name)
+                tensor_name = named_tensor.name
+                # Check if all collaborators have sent their data for the
+                # current key.
+                all_collaborators_sent = self.secagg.wait_for_all_collaborators(tensor_name)
```
rahulga1: in case of staggler, will it hang indefinitely?

openfl/utilities/secagg/setup.py
```diff
@@ -0,0 +1,354 @@
...
+                    self._aggregator_uuid,
+                    -1,
+                    False,
+                    (
+                        collaborator,
+                        "secagg",
+                    ),
+                )
+            )
+            # logger.info(
```
rahulga1: remove

openfl/component/collaborator/collaborator.py
theakshaypant: Need input on this. Refer https://github.com/securefederatedai/openfl/pull/1346#discussion_r1950239958.

openfl-workspace/keras/mnist_secagg/plan/plan.yaml
theakshaypant: Resolved in 1579c0559c31ea84dcba316ab2c48f148dd345a3.

openfl-workspace/keras/mnist_secagg/plan/plan.yaml
theakshaypant: Resolved in 1579c0559c31ea84dcba316ab2c48f148dd345a3.

openfl-workspace/keras/mnist_secagg/plan/plan.yaml
theakshaypant: Resolved in 1579c0559c31ea84dcba316ab2c48f148dd345a3

openfl-workspace/keras/mnist_secagg/src/mnist_utils.py
theakshaypant: Have followed the same format as all other workspaces where the funnction definition is in all of them. 
Changed file name in 1579c0559c31ea84dcba316ab2c48f148dd345a3.

openfl/component/aggregator/aggregator.py
theakshaypant: Resolved in a71ba795c29e87d01d880c48b31555d585949c31.

openfl/component/aggregator/aggregator.py
theakshaypant: Resolved in a71ba795c29e87d01d880c48b31555d585949c31.

openfl/component/aggregator/aggregator.py
theakshaypant: Resolved in a71ba795c29e87d01d880c48b31555d585949c31.

openfl/component/aggregator/aggregator.py
theakshaypant: There is no wait in this method. Only checks if all collaborators have shared "tensor_name", will rename it appropriately.

openfl/utilities/secagg/setup.py
theakshaypant: Resolved in ee2392156d9793ab9d1d9c7e5ffb5fa65cc8ab88.

payalcha:
@theakshaypant are we are storing keys and cipher text right in persistent storage? 
Collaborator restart should not impact, secure aggregation.

openfl-workspace/keras/mnist_secagg/src/dataloader.py
tanwarsh: you can remove this comment - refer https://github.com/securefederatedai/openfl/pull/1333#discussion_r1939815667

openfl/utilities/secagg/crypto.py
teoparvanov, line 1: I guess this PR needs rebasing, as this package has already been introduced by the Workflow API PR?
https://github.com/securefederatedai/openfl/tree/develop/openfl/utilities/secagg

MasterSkepticista (changes requested):
Few comments:
* The intended way of using callbacks, is by instantiating them during `__init__` of the component and passing them to `CallbacksList`. These `self.callbacks` internally trigger their own functions at certain points in code. A separate `callback` and `self.callbacks` object is anti-pattern
* I appreciate the boolean enable/disable aspect of this feature in the plan. I don't think we need an entire workspace worth of code just to highlight that the plan has secure aggregation enabled. It will be better to provide a guide (one that resides in the documentation) that talks about [details of] secure aggregation and how to set it in the plan, in my view.

openfl/utilities/secagg/crypto.py
theakshaypant: Rebased the PR.

theakshaypant:
@MasterSkepticista 
> * The intended way of using callbacks, is by instantiating them during `__init__` of the component and passing them to `CallbacksList`. These `self.callbacks` internally trigger their own functions at certain points in code. A separate `callback` and `self.callbacks` object is anti-pattern

Was able to fix the issue I was having with callbacks in 15e43aaf114dab825f47b8992495861a65f38c42.

> * I appreciate the boolean enable/disable aspect of this feature in the plan. I don't think we need an entire workspace worth of code just to highlight that the plan has secure aggregation enabled. It will be better to provide a guide (one that resides in the documentation) that talks about [details of] secure aggregation and how to set it in the plan, in my view.

Removed the example (fb17ee18adb940380cfad9b6a92e65a609d67aad) and updated docs (3e8c0a7449830b662da6f93524dae75c6f2603c3).

openfl/component/collaborator/collaborator.py
theakshaypant: Resolved in 15e43aaf114dab825f47b8992495861a65f38c42.

openfl-workspace/keras/mnist_secagg/src/mnist_utils.py
theakshaypant: Removed the example altogether in fb17ee18adb940380cfad9b6a92e65a609d67aad.

openfl-workspace/keras/mnist_secagg/src/dataloader.py
theakshaypant: Removed the workspace in fb17ee18adb940380cfad9b6a92e65a609d67aad.

openfl/component/collaborator/collaborator.py
theakshaypant: Resolved in 15e43aaf114dab825f47b8992495861a65f38c42.

openfl/utilities/secagg/crypto.py
theakshaypant: Some of these functions are also utilised in Workflow API; all of them are specific to SecAgg hence created a separate directory for all of them.

setup.py
theakshaypant: The SecAgg utility module is referenced by the aggregator and collaborator files which require pycryptodome, openfl installation will fail (ImportError) if this is removed. Open for suggestions from others on this.

openfl/component/aggregator/aggregator.py
theakshaypant: Resolved in 3144428f.

docs/about/features_index/secure_aggregation.rst
```diff
@@ -0,0 +1,120 @@
...
+            function: train_task
+            kwargs:
+            batch_size: 32
+            epochs: 1
+            metrics:
+            - loss
+        
+As can be seen in the above plan, by only enabling ``aggregator.settings.secure_aggregation`` and ``collaborator.settings.secure_aggregation`` in the workspace plan, one can enable SecAgg.
+
+After the flags have been set in plan.yml and the setup for the experiment is compleetd, one can verify that SecAgg was enabled by looking at the aggregator logs
```
rahulga1: typo `compleetd`

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,293 @@
...
+    pseudo_random_generator,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class CollaboratorSecAgg(Callback):
+    """
+    This callback is used by the collaborator to perform the setup steps
+    for scure aggregation on the collaborators.
```
rahulga1: typo `scure'

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,293 @@
...
+
+        self._send_to_aggregator(global_results, "generate_keys")
+        # Update callback params as the results for this step are reused at a
+        # later stage.
+        self.params.update(local_result)
+        logger.debug("SecAgg: Generate key-pair generation successful")
+
+    def _fetch_public_keys(self):
+        """
+        Fetches public keys from participants and identifies the index of the
```
rahulga1: if I am understanding correctly these functions will be called by collaborator only, so will it make sense to rename participant to collaborator for more clarity?
Applicable in other functions as well.

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        Use the private seed and agreed keys to calculate the masks to be
+        added to the gradients.
+        """
+        private_mask = pseudo_random_generator(self.params.get("private_seed"))
+        shared_mask = calculate_shared_mask(self.params.get("agreed_keys"))
+
+        return private_mask, shared_mask
+
+    def _save_tensors(self):
```
rahulga1: rename it to more descriptive one, name doesnt suggest what this tensor is for.

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,293 @@
...
+        """
+        Generates private and shared masks, stores them in a local tensor
+        dictionary, and caches the dictionary in the tensor database.
+
+        These tensors are then added to the gradient before sharing them
+        with the aggregator during trainign task.
```
rahulga1: typo `trainign `

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class CollaboratorSecAgg(Callback):
+    """
+    This callback is used by the collaborator to perform the setup steps
+    for scure aggregation on the collaborators.
+
+    Required params include:
+    - origin: Name of the collaborator using the callback.
+    - client: Client for aggregator secure aggregation setup.
```
rahulga1: what is this client? grpcclient object?

openfl/component/aggregator/aggregator.py
```diff
@@ -3,6 +3,7 @@
 
 """Aggregator module."""
 
+import json
```
rahulga1, line 6: I know json package might already be there, but have you added all the packages which are required with your changes in requirements.txt ?

openfl/utilities/secagg/setup.py
```diff
@@ -0,0 +1,346 @@
+# Copyright 2020-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+"""
+Thsi file contains the Setup class used on the server side for secure aggregation setup.
```
rahulga1: typo `Thsi`

openfl/utilities/secagg/setup.py
```diff
@@ -0,0 +1,346 @@
...
+            self._aggregate_key_shares()
+
+        if "seed_shares" in self._results and "key_shares" in self._results:
+            self._reconstruct_secrets()
+            # Generate agreed keys between all pairs of collaborators.
+            self._generate_agreed_keys()
+            # Save the local tensors to the tensor database.
+            self._save_tensors()
+
+    def _aggregate_public_keys(self):
```
rahulga1, line 101: I feel lots of code duplication, only thing which is changed the actual tensor name, so why not create a common function to do this:
`aggregated_tensor = []
        self._results["public_keys"] = {}
        self._results["index"] = {}
        index = 1
        for collaborator in self._collaborator_list:
            # Fetching public key for each collaborator from tensor db.
            nparray = self._tensor_db.get_tensor_from_cache(
                TensorKey(
                    "public_key",
                    self._aggregator_uuid,
                    -1,
                    False,
                    (
                        collaborator,
                        "secagg",
                    ),
                )
            )
            aggregated_tensor.append([index, nparray[0], nparray[1]])`

setup.py
rahulga1: I think this should be ok.

openfl/interface/aggregation_functions/secure_aggregation.py
```diff
@@ -0,0 +1,61 @@
...
+        for item in db_iterator:
+            if (
+                "tags" in item
+                and item["tags"] == ("secagg",)
+                and item["tensor_name"] == "masks_sum"
+            ):
+                masks = item["nparray"]
+                break
+
+        tensor_sum_with_masks = np.sum([tensor.tensor for tensor in local_tensors], axis=0)
```
rahulga1: how these  lines are calculating average, to me its calculating sum only?

openfl/component/aggregator/aggregator.py
theakshaypant: `pycryptodome` is the only additional module that needs to be installed which has been added to the requirements.

openfl/utilities/secagg/setup.py
theakshaypant: In 86f09a4b, created a `_aggregate_secret_shares` method which does the common aggregation between "seed_share" and "key_share". Public keys and cipher texts have a different way in which they are "aggregated".

docs/about/features_index/secure_aggregation.rst
theakshaypant: Fixed in e2fccffa

openfl/callbacks/secure_aggregation.py
theakshaypant: Fixed in 86f09a4b

openfl/callbacks/secure_aggregation.py
theakshaypant: Fixed in 86f09a4b

openfl/callbacks/secure_aggregation.py
theakshaypant: Fixed in 86f09a4b

openfl/callbacks/secure_aggregation.py
theakshaypant: Fixed in 86f09a4b

openfl/callbacks/secure_aggregation.py
theakshaypant: Fixed in 86f09a4b

openfl/utilities/secagg/setup.py
theakshaypant: Fixed in 86f09a4b

docs/about/features_index/secure_aggregation.rst
```diff
@@ -0,0 +1,120 @@
...
+            percentage: 0
+            tasks:
+            - aggregated_model_validation
+        template: openfl.component.RandomGroupedAssigner
+    collaborator:
+        settings:
+            db_store_rounds: 1
+            delta_updates: false
+            opt_treatment: RESET
+            secure_aggregation: true
```
teoparvanov: Shouldn't secure_aggregation be a property of the aggregator _only_? In any case, setting it in two places in the FL plan is error-prone. What if someone sets `secure_aggregation: true` for the aggregator and `secure_aggregation: false` for the collaborator? See my related suggestion below.

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,280 @@
...
+    pseudo_random_generator,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class CollaboratorSecAgg(Callback):
+    """
+    This callback is used by the collaborator to perform the setup steps
+    for secure aggregation on the collaborators.
```
teoparvanov: The naming and overview are a bit misleading. In fact, this callback prepares the collaborator to participate in the MPC required for secure aggregation. Why not rename this class to something along the lines of `SecAggBootstrapping` and adjust the description accordingly?

openfl/callbacks/secure_aggregation.py
```diff
@@ -0,0 +1,280 @@
...
+        # Fetch public keys for all collaborators from the aggregator.
+        collaborator_keys = self._fetch_public_keys()
+        # Generate ciphertexts for each collaborator and share them with the
+        # aggregator.
+        self._generate_ciphertexts(collaborator_keys)
+        # Decrypt the addressed ciphertexts and share them with the
+        # aggregator.
+        self._decrypt_ciphertexts(collaborator_keys)
+        # Save the tensors which are required for masking of gradients.
+        self._save_mask_tensors()
```
teoparvanov, line 60: If I understand correctly, this is essentially a "task 0" of the FL plan, in which the collaborators collectively compute their local masks. If so, I'm wondering whether, instead of a callback, it couldn't be implemented as a "first-class citizen" task like `locally_tuned_model_validation` or `aggregated_model_validation`. It could for example be called `compute_secagg_masks`. WDYT?

openfl/component/aggregator/aggregator.py
```diff
@@ -1049,7 +1101,19 @@ def _compute_validation_related_task_metrics(self, task_name) -> dict:
...
+                else:
+                    agg_function = WeightedAverage()
+            else:
+                agg_function = task_agg_function
+            # Check if secure aggregation is enabled, set aggregation function.
+            agg_function = (
+                SecureAggregation()
+                if "metric" in tags and self._secure_aggregation_enabled
+                else agg_function
+            )
```
teoparvanov: The logic here seems convoluted and repetitive. I'm sure it's possible to simplify and make it a little more readable 😅

openfl/component/collaborator/collaborator.py
```diff
@@ -85,7 +85,8 @@ def __init__(
         db_store_rounds=1,
         log_memory_usage=False,
         write_logs=False,
-        callbacks: Optional[List] = None,
+        callbacks: Optional[List] = [],
+        secure_aggregation=False,
```
teoparvanov, line 89: AFAIK the collaborator has access to the FL plan. So, can't it simply read the property from the aggregator section?

openfl/component/collaborator/collaborator.py
```diff
@@ -327,6 +340,9 @@ def do_task(self, task, round_number) -> dict:
             input_tensor_dict=input_tensor_dict,
             **kwargs,
         )
+        # If secure aggregation is enabled, add masks to the dict to be shared
+        # with the aggregator.
+        self._secure_aggregation_masking(global_output_tensor_dict)
```
teoparvanov: Instead of the comment, I suggest to pull the `if` statement outside of the function, effectively becoming:
```python
if self._secure_aggregation_enabled:
    self._secure_aggregation_masking(global_output_tensor_dict)
```

openfl/utilities/secagg/setup.py
```diff
@@ -0,0 +1,289 @@
...
+        local_tensor_dict = {
+            # TensorKey(
+            #     "private_seeds", "agg", -1, False, ("secagg", )
+            # ): [
+            #     [index, seed]
+            #     for index, seed in self._results["private_seeds"].items()
+            # ],
+            # TensorKey(
+            #     "agreed_keys", "agg", -1, False, ("secagg", )
+            # ): self._results["agreed_keys"],
```
teoparvanov: nit - is the commented-out code still needed?

teoparvanov (commented):
Hi @theakshaypant , I've done a first high-level review of this PR, mainly focused on interfaces and abstractions. I plan to look more closely into the secagg-specific logic in a follow-up review.

To answer your main question, I like the way you use tensors to transmit/aggregate metadata related to Secure Aggregation (via the `secagg` tag). Effectively, FL itself is a sort of multi-party computation scheme, so I don't see why not to leverage a similar mechanism for the bootstrap phase. I'd like to hear @kta-intel's opinion on this matter too!

I have a couple of suggestions, mostly focused on improving interfaces and modularity:

openfl/callbacks/secure_aggregation.py
theakshaypant: Your assessment of being "task 0" is correct, that was my initial thought as well - to implement the secagg bootstrapping as task(s).
However, I did see some challenges with the implementation. 
- Essentially each task is a method of the `TaskRunner`. Since there are couple of back-and-forth interaction with the aggregator required, do we give `AggregatorGRPCClient` access to the `TaskRunner`?
  - The user extends a base TaskRunner (`KerasTaskRunner`, `PyTorchTaskRunner`, etc), they would have access to the aggregator client if we go with this approach. How do we feel about that?
  - Do we add this `compute_secagg_masks` in the base `TaskRunner` to allow compatibility with keras, torch, etc.?
  - If we move over the above two points, since there are a few to and fro calls between aggregator and collaborator, would that not go against the current openfl principle of "aggregation" happening at the end of a round or am I overthinking this?
- We cannot implement these steps as multiple tasks  since the aggregation of task results happen at the end of the round (for all tasks at once) and here we would require result of one task as the input of another. 
- Would the user need to add a task to enable secagg, in this case? Adding a new secagg `task_group` perhaps?

With these thoughts in my mind, I felt `on_experiment_begin` callback would be a better place to have this functionality. 

Edit: Tagging @kta-intel here since this is in response to the review comment. 
> Effectively, FL itself is a sort of multi-party computation scheme, so I don't see why not to leverage a similar mechanism for the bootstrap phase.

openfl/utilities/secagg/setup.py
theakshaypant: With the current implementation, we can only get an average (not weighted) of the tensors since we only store the total sum of the masks. If we want to keep `WeightedAverage` functionality, we would need to store these tensors and calculate the masks at a later stage (when aggregating) allowing for the same weight to be used for the collaborator's tensor and mask to get a weighted average. 
Wanted input on what would be a more suitable way forward.

openfl/component/aggregator/aggregator.py
theakshaypant: What do we feel about the implementation in 54f218e or maybe an even more concise representation 
```
agg_function = (
    task_agg_function
    if "metric" not in tags
    else SecureAggregation() if self._secure_aggregation_enabled
    else WeightedAverage()
)
```

theakshaypant: Right, can remove lines 1112-1116 or do we want to keep it in that format
```
agg_function = (
    task_agg_function
    if "metric" not in tags
    else SecureAggregation() if self._secure_aggregation_enabled
    else WeightedAverage()
)
```

teoparvanov: I like this compact form better, as long as the logic is equivalent!

teoparvanov: I like this version better, as it is more compact. Does it capture the full logic of lines 1104-1116 in an equivalent way?

openfl/component/collaborator/collaborator.py
theakshaypant: Moved conditional outside the method in d47d209f33bfdf36a93d5002e205c5d65080a8cc.

openfl/component/aggregator/aggregator.py
theakshaypant: Yes, this would be a shorthand representation of the same logic. Added it to the code in 4e0ac0bbea4f7e6ce510401dfabeaf6281be0f16.

openfl/callbacks/secure_aggregation.py
theakshaypant: Renamed the callback and updated docstring in 359912b4a0d17b6ac3b9246a22ef40dfc8c6b5bd.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Updated it (in cc9225ff) such that `secure_aggregation: true` needs to be only set under the aggregator settings. Collaborator settings are updated by reading the value from aggregator. 
The named argument is still present in the `Collaborator` class but read from the same key as then for `Aggregator`.

openfl/component/collaborator/collaborator.py
theakshaypant: Refer https://github.com/securefederatedai/openfl/pull/1346#discussion_r1963247290

openfl/callbacks/secure_aggregation.py
teoparvanov: I am generally ok with the callback-based approach (esp. for the bootstrapping task). Just wanted to get @kta-intel 's point of view as he's more deeply knowledgeable about the original intentions and the history of the TaskRunner framework.

kminhta: I might have to give it deeper thought, but I don't see anything wrong with doing this as a callback-based approach. While it would be nice to have a more "built-in" method for this step, I'm inclined to agree with @theakshaypant 's concerns with having it implemented as a method of the Task Runner.

openfl/utilities/secagg/setup.py
kminhta: I think we need to preserve `WeightedAverage` - in practice, datasets will be pretty disparate and not all sources can be weighted equally

theakshaypant: Update the aggregation logic in 574c102647169de56f4b66452640086e7944ed99.

openfl/interface/aggregation_functions/secure_aggregation.py
theakshaypant: Resolved in 574c102647169de56f4b66452640086e7944ed99.

theakshaypant: Added testing details in PR description.

openfl/component/aggregator/aggregator.py
theakshaypant: Also updated testing details in the PR description with these changes.

openfl/component/collaborator/collaborator.py
theakshaypant: Updated testing details in description with this change.

openfl/utilities/secagg/setup.py
theakshaypant: Testing details also added in the description.

docs/about/features_index/secure_aggregation.rst
```diff
@@ -0,0 +1,119 @@
...
+        settings: {}
+        train:
+            function: train_task
+            kwargs:
+            batch_size: 32
+            epochs: 1
+            metrics:
+            - loss
+        
+As can be seen in the above plan, by only enabling ``aggregator.settings.secure_aggregation`` and ``collaborator.settings.secure_aggregation`` in the workspace plan, one can enable SecAgg.
```
payalcha: What will happen if aggregator.settings.secure_aggregation is set to true but collaborator.settings.secure_aggregation is set to false.

docs/about/features_index/secure_aggregation.rst
```diff
@@ -0,0 +1,119 @@
...
+
+After the flags have been set in plan.yml and the setup for the experiment is completed, one can verify that SecAgg was enabled by looking at the aggregator logs
+
+.. code-block:: bash
```
payalcha, line 102: If secure_aggregation and metrics both enabled, what details will be published in metrics tensorboard.

docs/about/features_index/secure_aggregation.rst
theakshaypant: User needs to only set `aggregator.settings.secure_aggregation`. Related discussion in https://github.com/securefederatedai/openfl/pull/1346#discussion_r1961809884. 
Thanks for pointing this out! Fixed doc in  4730aa4. 

Edit: If the user still specifies secure_aggregation under collaborator settings, the value will be overridden by whatever is provided under aggregator settings. Refer [these](https://github.com/securefederatedai/openfl/blob/4730aa4387d5ed00eec9f70d84b4f4b2b4c11170/openfl/federated/plan/plan.py#L517-L519) lines in the changes.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Good catch, with metrics enabled only the aggregated round metrics will be accurate. All the other values will be logged with masks.

theakshaypant:
@teoparvanov Also ran openfl-security CI tests in https://github.com/intel-innersource/frameworks.ai.openfl.openfl-security/pull/931 using one of the newer commits 52a3a04d91e9ac11edc851bdb36bcb35d201b4aa.

openfl/interface/aggregation_functions/secure_aggregation.py
```diff
@@ -0,0 +1,186 @@
...
+from openfl.interface.aggregation_functions.core import AggregationFunction
+from openfl.interface.aggregation_functions.weighted_average import WeightedAverage
+from openfl.utilities import LocalTensor
+from openfl.utilities.secagg import (
+    calculate_shared_mask,
+    pseudo_random_generator,
+)
+
+
+class SecureAggregation(AggregationFunction):
```
teoparvanov: It would be even more elegant to call this class `SecureWeightedAverage`, and extend `WeightedAverage` (effectively "adding security" to the base WeightedAvg implementation).

openfl/interface/aggregation_functions/secure_aggregation.py
```diff
@@ -0,0 +1,186 @@
...
+            tags: tuple of tags for this tensor
+        Returns:
+            np.ndarray: aggregated tensor
+        """
+        # Generate masks for the collaborators if not laready done.
+        self._generate_masks(db_iterator)
+        # Calaculate the weighted avreage of collaborator masks.
+        weighted_mask = self._calculcate_weighted_mask_average(self._private_masks, local_tensors)
+        # Get weighted average for shared tensors.
+        tensor_avg = WeightedAverage().call(local_tensors)
```
teoparvanov: ... then this line would become:
```python
tensor_avg = super().call(local_tensors)
```

WDYT?

docs/about/features_index/secure_aggregation.rst
```diff
@@ -0,0 +1,119 @@
...
+=======================================
+
+In Federated Learning (FL), Secure Aggregation (SecAgg) restricts the aggregator to only learn the summation or average of the updates from collaborators.
+
+OpenFL integrates `SecAgg <https://eprint.iacr.org/2017/281.pdf>`_ into TaskRunner API as well as the Workflow API.
+
+TaskRunner API
+-------------------------------------
+
+OpenFL treats SecAgg as a first class citizen and can be enabled for any experiment by simply modifying the plan.
```
teoparvanov: I'd suggest replacing the somewhat socially loaded term "first-class citizen", with the more neutral and informative "core security feature".

docs/about/features_index/secure_aggregation.rst
```diff
@@ -0,0 +1,119 @@
...
+        settings:
+            best_state_path: save/best.pbuf
+            db_store_rounds: 2
+            init_state_path: save/init.pbuf
+            last_state_path: save/last.pbuf
+            persist_checkpoint: true
+            persistent_db_path: local_state/tensor.db
+            rounds_to_train: 1
+            secure_aggregation: true
+        template: openfl.component.Aggregator
```
teoparvanov, line 31: Could you highlight just the changes/additions relative to the "non-secure" FL plan?

teoparvanov (approved):
Really nice work, thank you @theakshaypant ! I've left a few final comments, but this PR is good to go from my perspective.

PS: Making SecAgg a core framework capability is a great step forward in enabling software-based privacy in OpenFL (without requiring a TEE to protect the local updates).

docs/about/features_index/secure_aggregation.rst
theakshaypant: How do we feel about adding it in a git diff kind of format?
```
    @@ -8,7 +8,8 @@ aggregator :
        init_state_path : save/init.pbuf
        rounds_to_train : 10
    +    secure_aggregation: true
    
    collaborator :
        defaults : plan/defaults/collaborator.yaml
```
Also looking at [`emphasize-lines`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-emphasize-lines) with the entire plan and highlighted ` secure_aggregation: true`.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Addressed in 09da2d8d.

openfl/interface/aggregation_functions/secure_aggregation.py
theakshaypant: Thanks! I like this suggestion, made the appropriate changes in 525086e3.

openfl/interface/aggregation_functions/secure_aggregation.py
theakshaypant: Same as https://github.com/securefederatedai/openfl/pull/1346#discussion_r1969721185, addressed in https://github.com/securefederatedai/openfl/commit/525086e3ac54c5d2920c3f0ee2231d968553e2e4

docs/about/features_index/secure_aggregation.rst
teoparvanov: I think the emphasize-lines approach would better stand out!

openfl/interface/aggregation_functions/secure_weighted_average.py
```diff
@@ -0,0 +1,185 @@
...
+from openfl.interface.aggregation_functions.weighted_average import WeightedAverage
+from openfl.utilities import LocalTensor
+from openfl.utilities.secagg import (
+    calculate_shared_mask,
+    pseudo_random_generator,
+)
+
+
+class SecureWeightedAverage(WeightedAverage):
+    """FedAvg with secure aggregation."""
```
teoparvanov: nit - WeightedAvg with secure aggregation

theakshaypant: Updated in 53a823d9.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Updated in a2f4631e. This is what it looks like now. 

Screenshot 2025-02-25 at 8 45 16 PM

teoparvanov: Awesome, thank you!

PR #1347, opened 2025-02-06
Task Runner E2E - Added Resiliency Tests

noopurintel:
This PR aims to add resiliency test cases for Task Runner.

1. Native/basic environment - by killing the "fx start" processes and starting the participants again.
2. Dockerized environment - by stopping and starting the docker containers with a few sec gap in between.

**Scenarios covered (in the same order for single federation run):**
1. Restart aggregator
2. Restart all collaborators
3. Restart all participants (aggregator + collaborators)

After every restart, check if the current round > previous round, in addition to the existing overall federation completion check.

**Major changes:**
1. Added workflow `task_runner_basic_e2e_resiliency.yml` to run native env scenarios.
2. Added `tr_resiliency_tests.py` to test native and dockerized approaches.
3. Modified `conftest.py` to have a different (increased) for num_rounds default value, so that restarts can be properly tested.
4. Modified functions `run_federation` and `run_federation_for_dws` to return bool instead of result files. Assert statements added in accordance.
5. Modified `aggregator.py` to include params `res_file`, `tensor_db_path` and `collaborator.py` to include the latter. This is done to avoid using os.path.join() multiple times whenever these values are needed.
6. Modified `conftest_helper.py` to include Default values in the help section, so that if `--help` is used for a test, it is more helpful.
7. **IMP** - Previously all docker containers were started with no additional command (as part of fixture) and later the fx commands were run using container ids. But when a container restarts, it won't know which process to start causing federation to get stuck. Now passing the relevant fx start commands during the container start, such that any restart or stop+ start has no issues.
8. **IMP** - Previously the workspace path didn't include testcase name in it, for e.g. `/home/runner/results/torch_cnn_mnist/aggregator/workspace`, thus if two tests were running in sequence, there was overwriting. Now added it, for e.g. `/home/runner/results/test_federation_via_native_with_restarts/torch_cnn_mnist/aggregator/workspace`

**Successful runs (by adding resiliency jobs temporarily to the existing workflows - removed later):**

1. Task Runner E2E native - https://github.com/noopurintel/openfl/actions/runs/13181523345
2. Task Runner E2E dockerized - https://github.com/noopurintel/openfl/actions/runs/13181144915

**After review comment incorporation:**

1. Task Runner E2E - https://github.com/noopurintel/openfl/actions/runs/13202045096
2. Task Runner E2E dockerized - https://github.com/noopurintel/openfl/actions/runs/13201428211
3. Task Runner FedEval - https://github.com/noopurintel/openfl/actions/runs/13201595587
4. Task Runner FedEval dockerized - https://github.com/noopurintel/openfl/actions/runs/13204364598

(Failures in python 3.12 runs is due to this issue - https://github.com/securefederatedai/openfl/issues/1352)


noopurintel:
> Bandit found more than 20 potential problems in the proposed changes. Check the [Files changed](/securefederatedai/openfl/pull/1347/files) tab for more details.

All the assert statements reported are part of pytest functions inside e2e test framework only, thus dismissed those alerts with "used in tests".

payalcha (approved):

PR #1354, opened 2025-02-07
Fix `fx --version`

shgandhi:
Re-enables list commands when issuing fx without any args or commands by removing the use of `invoke_without_commands`, introduced for invoking `fx --version`.

### DIFF
--- a/openfl/interface/cli.py
+++ b/openfl/interface/cli.py
@@ -14,7 +14,6 @@ from logging import basicConfig
 from pathlib import Path
 from sys import argv, path
 
-import pkg_resources
 from click import (
     Group,
     argument,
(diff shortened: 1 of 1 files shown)


MasterSkepticista (changes requested):
Can we just use this?

```python
import openfl

# ... code ...
if version:
  echo(f"OpenFL version: {openfl.__version__}")
  return
# ... code ...
```

No need for try-except, importlib, context invocation functions etc,

noopurintel:
@shgandhi - to test the changes, you can run this workflow (from your own forked repo's Actions) - https://github.com/noopurintel/openfl/actions/workflows/task_runner_basic_e2e.yml by selecting Python version as 3.12 from the dropdown list.

teoparvanov (approved):

MasterSkepticista (approved):

PR #1356, opened 2025-02-09
Migrate defaults to a separate file

ishaileshpant:
This PR refactors the aggregator and collaborator modules by replacing hard-coded “magic” numbers and strings with centralized constant definitions. A new file, constants.py, has been introduced in the openfl/component directory to store these default values. This change improves maintainability and consistency across the codebase.

### Motivation

- Avoiding Magic Numbers/Strings:

 Hard-coded values (e.g., 256, "tensor.db", "RESET", etc.) were scattered in the code. Extracting them into a constants file helps prevent errors and makes future updates easier.

- Enhanced Readability & Maintainability:

 Using descriptive constant names (like ROUNDS_TO_TRAIN and CERT_COMMON_NAME) clarifies the purpose of each value and allows for a single point of modification if defaults need to change.

### DIFF
--- a/openfl/component/constants.py
+++ b/openfl/component/constants.py
@@ -1,0 +1,10 @@
+# Centralized constants for openfl/component modules
+
+ROUNDS_TO_TRAIN = 256
+EVALUATION_ROUNDS = 1
+DB_STORE_ROUNDS = 1
+PERSISTENT_DB_PATH = "tensor.db"
+CERT_COMMON_NAME = ""
+OPT_TREATMENT = "RESET"
+DEVICE_ASSIGNMENT_POLICY = "CPU_ONLY"
+DELTA_UPDATES = False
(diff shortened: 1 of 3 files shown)


MasterSkepticista (commented):
Thanks for picking this up @ishaileshpant.

Two suggestions:
* It is recommended to keep constants at a [module-level](https://peps.python.org/pep-0008/#constants).
* Can remove `DEFAULT_` string for all constants. Upper case obviates the need to be explicit.

ishaileshpant:
> Thanks for picking this up @ishaileshpant.
> 
> Two suggestions:
> 
> * It is recommended to keep constants at a [module-level](https://peps.python.org/pep-0008/#constants).
> * Can remove `DEFAULT_` string for all constants. Upper case obviates the need to be explicit.

The only case that it won't address is some constants span across modules and hence i move to global package level else duplication of constants will be there

agree on the "default" prefix, in fact we can even name the file as defaults, that would further simplify the names and purpose becomes explicit

openfl/component/constants.py
theakshaypant, line 1: Would suggest renaming it to `defaults.py`. 
Using `defaults.ROUNDS_TO_TRAIN` increases readability compared to `constants.ROUNDS_TO_TRAIN` in my opinion.

PR #1357, opened 2025-02-10
WIP: Pytorch version updated to 2.5.1

tanwarsh:
Pytorch version updated to 2.5.1.

As we want to update PyTorch to version 2.5.1, we are proceeding with this PR along with a patch to fix the issue:
Type parameter +RV without a default follows type parameter with a default


rahulga1 (approved):

rahulga1 (approved):

kminhta (changes requested):
Have we done any further investigation into why this redirection is necessary (i.e. why/when is `pip._vendor.typing_extensions` being called to enforce parameter ordering as opposed to `typing_extensions`)?

I know this was our originally proposed workaround, and I get that it might still be the only solution, but since we already needed to patch `distutil` in order run `torch==2.4.x`, I am very hesitant to introduce additional patches to get it working further.

My reason being that it is better for long term maintainability that we find a robust solution and avoid patching if we can.

tanwarsh:
Thank you @kta-intel for the review. closing this PR for now.

theakshaypant (commented):
Please take a look at the pipelines, they have the parameter error.

tanwarsh:
> Please take a look at the pipelines, they have the parameter error.

Thank you for the review but this is still WIP.

tanwarsh:
@teoparvanov, @kta-intel, @rahulga1. 

After investigating this issue in detail, the only solution other than a patch is to fix this in the `typing_extensions` library by adding a more robust way to detect NoDefault from different modules. I will raise an issue in `typing_extensions` regarding this.

tanwarsh:
Logged the issue in https://github.com/securefederatedai/openfl/issues/1441.

PR #1358, opened 2025-02-10
TaskRunner E2E - Adding keras/jax/mnist and keras/torch/mnist models

noopurintel:
Also, added Task Runner E2E and resiliency workflows as part of PR pipeline, only to be triggered if the PR label contains `task_runner_e2e`

Refer - https://github.com/securefederatedai/openfl/actions/runs/13236383645

Successful runs:

1. Task Runner E2E native - https://github.com/noopurintel/openfl/actions/runs/13235166107
2. Task Runner E2E dockerized - https://github.com/noopurintel/openfl/actions/runs/13235167787 (working for jax, but failing for torch due to space issue)


.github/workflows/task_runner_dockerized_ws_e2e.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        model_name: ["keras/mnist"]
+        model_name: ["keras/mnist", "keras/jax/mnist", "keras/torch/mnist"]
```
payalcha, line 61: Noopur .. In every docker tests write a comment that it will not work in current github runner env

payalcha (approved):

.github/workflows/task_runner_dockerized_ws_e2e.yml
noopurintel: Done.

Issue #1367, opened 2025-02-11
Aggregator unable to recover if it is restarted during model save


noopurintel:
**Describe the bug**

If the aggregator is restarted while the model is being saved, it is not recoverable and keeps throwing exceptions.

**To Reproduce**

1. Start the federation using say keras/mnist model and 2 collaborators.
2. Wait for 1-2 min - so that the tensor.db is created and a few iterations of models are saved.
3. As soon as tensor.db -> task_results table contains 6 entries (= 2 collaborators *3 tasks), restart the aggregator.
4. Aggregator starts throwing error and is unable to recover.

**Expected behavior**

Irrespective of the timing of restart, aggregator should always come up in a healthy state.

**Screenshots**

Tensor database **task_results** table status just before model saving.

<img width="1386" alt="Image" src="https://github.com/user-attachments/assets/a110a161-9c8a-434a-b80b-bbb7fb7f4254" />

Aggregator errors:

<img width="1514" alt="Image" src="https://github.com/user-attachments/assets/17bf5196-e639-49b7-ae66-e93ca2a6b7a1" />

Wait time added in `_save_model` (to increase the chances of getting into this issue): 

<img width="853" alt="Image" src="https://github.com/user-attachments/assets/b85a729d-3235-4033-9228-9906cc79c324" />

**Additional context**

1. In most of the scenarios, aggregator is able to recover properly.
2. To narrow down the issue, we tried multiple things. And this is what helped us achieve the error scenario - we added a sleep time of 30 seconds inside `_save_model` function (file: `aggregator.py`) before cleaning of task_results table of tensor db.

Failed aggregator logs (same issue but different errors probably due to failure happening at diff stages):

[failed_aggregator_post_restart.log](https://github.com/user-attachments/files/18751592/failed_aggregator_post_restart.log)

[failed_agg_callbacks_error.log](https://github.com/user-attachments/files/18752222/failed_agg_callbacks_error.log)

cloudnoize:
@noopurintel nice catch! the bug happens only on round 0 when the aggregator crashes after collecting all data needed to finish the round and save the model. 
in that scenario, the model proto is not constructed by the recovery process, leading to this crash.

cloudnoize:
https://github.com/securefederatedai/openfl/pull/1373

noopurintel:
[...]

PR #1373, opened 2025-02-12
Initialize proto model with initial tensors before recovery

cloudnoize:
Address https://github.com/securefederatedai/openfl/issues/1367

if the Aggregator crashes after collecting all data needed to finish round 0 but before it saved the model, on restart the model proto is not constructed by the recovery process, leading to a crash.

The PR initialize the proto model with the initial tensors that can be overridden by the recovery process if needed.
For safety, I moved other initializations before the recovery process as well.


noopurintel:
Thanks for fixing this @cloudnoize . I have successfully tested these changes and provided the run details on the Issue page itself.

openfl/component/aggregator/aggregator.py
```diff
@@ -198,15 +203,10 @@ def __init__(
             self.model: base_pb2.ModelProto = utils.load_proto(self.init_state_path)
             self._load_initial_tensors()  # keys are TensorKeys
 
-        self.collaborator_tensor_results = {}  # {TensorKey: nparray}}
+        if self.persistent_db and self._recover():
+            logger.info("recovered state of aggregator")
```
noopurintel: nit: **R**ecovered (as all other log statements start with the title case)

noopurintel (approved):
Minor review comment provided. Changes look good otherwise. Tested successfully.

teoparvanov (approved):

rahulga1 (approved):

PR #1327, opened 2025-01-31
[Workflow Interface]: Fix for FederatedRuntime with TLS (Issue #1265)

ishant162:
**Background**
Issue: #1265
 
**Change Description & Modifications**
 
- Updated the `agg_addr` field to use the Director's hostname and Synchronized the TLS configuration with the values provided by `FederatedRuntime` in the 
network section of `plan.yaml` for alignment with the system configuration.
- Ensured all changes made to `plan.yaml` are correctly propagated to the prepared workspace.
- Updated `network.yaml` by removing unused keys from its settings.
 
**Verfication**
 
- Verified that the `agg_addr` is correctly updated with the Director's hostname in `plan.yaml`, preventing recurrence of the issue.
- Verified that the experiment runs successfully with mTLS certificates generated using both `localhost` and the machine's FQDN (e.g `soc-PF2F2E02.clients.intel.com`) 
as the Common Name for the Director's certificates.
 
 
**Files modified**
 
- `openfl/experimental/workflow/federated/plan/plan.py`
- `openfl/experimental/workflow/runtime/federated_runtime.py`
- `openfl/experimental/workflow/workspace_export/export.py`
- `openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/workspace/plan/defaults/network.yaml`
 
**NOTE**
The Director Admin must ensure mTLS certificates have the correct Common Name (CN), matching the hostname in the Director's configuration. Envoy Admin and Experiment Manager must use this hostname to facilitate proper communication and prevent connectivity issues.
 
This PR fixes #1265

PR #1376, opened 2025-02-13
[Workflow Interface] Fix for FederatedRuntime bringup in distributed infrastructure (backport PR #1327 in OpenFL v1.7.x)

ishant162:
**Background:**
FederatedRuntime bringup is not successful in a distributed infrastructure setup when Director, Envoys and User are located on different machines. This PR is backporting the fix from openfl/develop branch to OpenFL v1.7.x release
Reference: Issue #1265.
Fix: #1327.
 
**Change Description & Modifications:**
 
- The `agg_addr` field in `plan.yaml` is now updated with the user-provided value in `FederatedRuntime` (`director_node_fqdn`).
- A detailed description of this change is provided in PR #1327
 
**Verification**
- [X] Confirmed that agg_addr is correctly updated with the Director's hostname in plan.yaml, preventing the issue from recurring.
- [X] Local infrastructure: Tested with & without TLS .
- [X]  Distributed infrastructure: Without TLS.
 
**Files modified**
- `openfl/experimental/workflow/federated/plan/plan.py`
- `openfl/experimental/workflow/runtime/federated_runtime.py`
- `openfl/experimental/workflow/workspace_export/export.py`
- `openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/workspace/plan/defaults/network.yaml`
 
**Note:**
1. The Envoy admin and Experiment Manager must ensure they provide the correct Director IP to enable proper communication and prevent connectivity issues.
2. Admins and users must ensure that the same Python version is maintained across the infrastructure.

### DIFF
--- a/openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/workspace/plan/defaults/network.yaml
+++ b/openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/workspace/plan/defaults/network.yaml
@@ -2,8 +2,6 @@ template: openfl.federation.Network
 settings:
     agg_addr                   : auto
     agg_port                   : auto
-    hash_salt                  : auto
     tls                        : True
     client_reconnect_interval  : 5
     disable_client_auth        : False
-    cert_folder                : cert
--- a/openfl/experimental/workflow/federated/plan/plan.py
+++ b/openfl/experimental/workflow/federated/plan/plan.py
@@ -163,6 +163,7 @@ class Plan:
                     f"[blue]{plan_config_path}[/].",
                     extra={"markup": True},
                 )
+                Plan.dump(plan_config_path, plan.config)
                 Plan.logger.info(dump(plan.config))
 
             return plan
@@ -338,7 +339,7 @@ class Plan:
         private_key=None,
         certificate=None,
         client=None,
-        tls=False,
+        tls=True,
         envoy_config=None,
     ) -> "Collaborator":
         """Get collaborator.
@@ -406,7 +407,7 @@ class Plan:
         root_certificate=None,
         private_key=None,
         certificate=None,
-        tls=False,
+        tls=True,
     ) -> AggregatorGRPCClient:
         """Get gRPC client for the specified collaborator.
 
@@ -425,8 +426,8 @@ class Plan:
         Returns:
             AggregatorGRPCClient: gRPC client for the specified collaborator.
         """
-        common_name = collaborator_name
-        if not root_certificate or not private_key or not certificate:
+        if tls and not (root_certificate and private_key and certificate):
+            common_name = collaborator_name
             root_certificate = "cert/cert_chain.crt"
             certificate = f"cert/client/col_{common_name}.crt"
             private_key = f"cert/client/col_{common_name}.key"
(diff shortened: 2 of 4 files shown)


openfl/experimental/workflow/federated/plan/plan.py
```diff
@@ -406,7 +407,7 @@ def get_client(
         root_certificate=None,
         private_key=None,
         certificate=None,
-        tls=False,
+        tls=True,
```
payalcha, line 410: TLS is passed as true by default. But as per the discussions with Patrick, Teo and Sachin, is that TLS is not production ready feature for workflow, so we should not mark it as true (default) in the plan.

ishant162: The default `tls=True` is set because `get_client()` is used by both `FederatedRuntime` and Experimental `Aggregator-based workflow`. In the `Aggregator-based workflow`, TLS is always enabled, so having `tls=True` by default ensures consistency.

Previously, when `tls=False`, For `FederatedRuntime`, the function was unnecessarily generating certificate paths. My implementation avoids that by ensuring certificate paths are only set when `tls=True` and no certificates are provided. This keeps the logic clean and prevents unintended certificate path generation.

payalcha: Aggregator-based workflow is currently not used and tested anywhere. Can you please confirm once with Teo and Sachin, as I really think we shouldn't take decision based on Aggregator-based workflow for Workflow API.

Also default must be what is production ready.

payalcha: As discussed with Sachin, it seems internal APIs. 
But as per my understanding by default value should be disable-tls, will raise separate issue for the same.

payalcha (approved):

openfl/experimental/workflow/federated/plan/plan.py
ishant162: As aligned over the call, This change involves an internal API update and does not impact any user-facing APIs.
For FederatedRuntime, TLS can be adjusted based on user input.

**Note:**
When starting the Director and Envoy via the CLI, users must specify the `--disable-tls` option to disable TLS. By default, TLS remains enabled.

openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/workspace/plan/defaults/network.yaml
```diff
@@ -2,8 +2,6 @@ template: openfl.federation.Network
 settings:
     agg_addr                   : auto
     agg_port                   : auto
-    hash_salt                  : auto
     tls                        : True
```
teoparvanov, line 5: Have you tested both with/without TLS in the distributed setup?
If so, please add it explicitly to the PR description.

ishant162: Updated the PR description.

teoparvanov (approved):

openfl-workspace/experimental/workflow/AggregatorBasedWorkflow/workspace/plan/defaults/network.yaml
```diff
@@ -2,8 +2,6 @@ template: openfl.federation.Network
 settings:
     agg_addr                   : auto
     agg_port                   : auto
-    hash_salt                  : auto
```
lior1-malka, line 5: Why is this being removed?

ishant162: This attribute is not used anywhere in the OpenFL codebase, so it has been removed.

PR #1382, opened 2025-02-18
Do not run windows workflow in draft stage

noopurintel:
Condition to run the workflow in case of events like a scheduled run or a ready pull request was missed out earlier. Added the same.

Previously - https://github.com/securefederatedai/openfl/actions/runs/13367634548

Successful skip after code change (with this PR in draft stage) - https://github.com/securefederatedai/openfl/actions/runs/13382494757/job/37373380833?pr=1382

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -14,6 +14,10 @@ env:
 
 jobs:
   pytest-coverage: # from pytest_coverage.yml
+    if: |
+      (github.event_name == 'schedule' && github.repository_owner == 'securefederatedai') ||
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -2,6 +2,7 @@ name: Windows (latest)
 
 on:
   workflow_call:
+  workflow_dispatch:
   schedule:
```
payalcha, line 6: Since it is running in workflow, can you please remove schedule as well.

payalcha (approved):

PR #1388, opened 2025-02-18
Make GRPC transport decorators configurable & disable atomic decorator by default

bsenthilr:
### Problem statement
The atomic connection decorators always reestablish connections on transport IO. This is not a desired IO characteristic especially due to penalty of reconnecting.
This PR proposes configuration to make these decorators option. By default atomic connections are disabled and can be provisioned when required.

### Background information on atomic connections (from @psfoley)
The background of reestablishing a connection for every request was due to early issues. There were problems resulting from long established connections when there was any kind of network instability. This started happening more with certain releases of gRPC around that time (1.32?). 
The atomic connection decorator, and establishing the connection only when communication was needed, was the solution for that. 
As long as the retry logic is left intact, I don't expect there would be problems making the change back to long standing connections (as long as this is thoroughly tested) 

### Summary of changes
- Add knobs to provision these decorators optionally when required.
- Log connection points since this is critical for debugging and tuning other attributes of service such as rate limiting etc.,

### Testing done
- Verified federation run with atomic connectivity disabled hence eliminating service level reconnects.
- Verified federation run with atomic connectivity enabled and correlating reconnects with logs.

### DIFF
--- a/openfl-workspace/workspace/plan/defaults/network.yaml
+++ b/openfl-workspace/workspace/plan/defaults/network.yaml
@@ -7,3 +7,4 @@ settings:
     client_reconnect_interval  : 5
     require_client_auth        : True
     cert_folder                : cert
+    enable_atomic_connections  : False
--- a/openfl/transport/grpc/aggregator_client.py
+++ b/openfl/transport/grpc/aggregator_client.py
@@ -96,7 +96,10 @@ class RetryOnRpcErrorClientInterceptor(
             if isinstance(response, grpc.RpcError):
                 # If status code is not in retryable status codes
                 self.sleeping_policy.logger.info("Response code: %s", response.code())
-                if self.status_for_retry and response.code() not in self.status_for_retry:
+                if (
+                    self.status_for_retry
+                    and response.code() not in self.status_for_retry
+                ):
                     return response
 
                 self.sleeping_policy.sleep()
@@ -117,7 +120,9 @@ class RetryOnRpcErrorClientInterceptor(
         """
         return self._intercept_call(continuation, client_call_details, request)
 
-    def intercept_stream_unary(self, continuation, client_call_details, request_iterator):
+    def intercept_stream_unary(
+        self, continuation, client_call_details, request_iterator
+    ):
         """
         Wrap intercept call for stream->unary RPC.
 
@@ -135,6 +140,8 @@ class RetryOnRpcErrorClientInterceptor(
 
 def _atomic_connection(func):
     def wrapper(self, *args, **kwargs):
+        if not self.enable_atomic_connections:
+            return func(self, *args, **kwargs)
         self.reconnect()
         response = func(self, *args, **kwargs)
         self.disconnect()
@@ -145,6 +152,8 @@ def _atomic_connection(func):
 
 def _resend_data_on_reconnection(func):
     def wrapper(self, *args, **kwargs):
+        if not self.resend_data_on_reconnection:
+            return func(self, *args, **kwargs)
         while True:
             try:
                 response = func(self, *args, **kwargs)
@@ -198,6 +207,8 @@ class AggregatorGRPCClient:
         federation_uuid=None,
         single_col_cert_common_name=None,
         refetch_server_cert_callback=None,
+        enable_atomic_connections=False,
+        resend_data_on_reconnection=True,
         **kwargs,
     ):
         """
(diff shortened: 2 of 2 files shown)


bsenthilr:
Log snippets with atomic connects:
```
[collaborator1]	           INFO     Waiting for tasks...                                                                                                     collaborator.py:233
[collaborator1]	           INFO     Disconnecting from gRPC server at **redacted**                                                aggregator_client.py:365
[collaborator1]	           INFO     Connecting to gRPC at **redacted**                                                            aggregator_client.py:385

```
Verified absence of reconnects which disabling atomic connects.

openfl/transport/grpc/aggregator_client.py
```diff
@@ -345,7 +355,7 @@ def validate_response(self, reply, collaborator_name):
 
     def disconnect(self):
         """Close the gRPC channel."""
-        self.logger.debug("Disconnecting from gRPC server at %s", self.uri)
+        self.logger.info("Disconnecting from gRPC server at %s", self.uri)
```
kminhta, line 361: Should we keep this as `debug` ? It may get too verbose if a user sets `enable_atomic_connections=True`

openfl/transport/grpc/aggregator_client.py
```diff
@@ -365,7 +375,7 @@ def reconnect(self):
                 self.private_key,
             )
 
-        self.logger.debug("Connecting to gRPC at %s", self.uri)
+        self.logger.info("Connecting to gRPC at %s", self.uri)
```
kminhta, line 381: Same question here

kminhta (commented):
Thanks Senthil, LGTM

Just wanted to confirm (and for historical context, in case anyone comes across this thread), that a user should still be able to use atomic connections via the plan.yaml?

```
network:
  settings:
    enable_atomic_connections: True
```

openfl/transport/grpc/aggregator_client.py
bsenthilr: its generally good practice to log connection logs since its key aspect of debugging and operation. Even with atomic would anticipate only few of these per round correct?

kminhta: > its generally good practice to log connection logs since its key aspect of debugging and operation.

Ah yeah, makes sense

> Even with atomic would anticipate only few of these per round correct?

That's true. On second thought, it should not be overwhelmingly verbose even with atomic connections

bsenthilr: In general we would need to add rate limited log support for such concerns.

bsenthilr:
> Just wanted to confirm (and for historical context, in case anyone comes across this thread), that a user should still be able to use atomic connections via the plan.yaml?
> 
> ```
> network:
>   settings:
>     enable_atomic_connections: True
> ```

good call, fixed @kta-intel . There is certainly gap on lacking capability to set on desired collabs. But this is a good start. Defaulted to false under network settings for now.

kminhta (approved):
> good call, fixed @kta-intel . There is certainly gap on lacking capability to set on desired collabs. But this is a good start. Defaulted to false under network settings for now.

Thanks for checking. Considering that the original atomic connection didn't have that functionality either, this is already a good addition

openfl/transport/grpc/aggregator_client.py
MasterSkepticista: @bsenthilr We can issue a one-time [`RuntimeWarning: Atomic connections may trigger rate-limiting. For more details, see <PR-1388 URL>.`](https://docs.python.org/3/library/exceptions.html#RuntimeWarning) 

Edit: With this warning, we could revert to `debug` messages (?)

MasterSkepticista (changes requested):
Minor change, good to go.

bsenthilr:
@kagrawa2 is going to add some updates for collab reconnection resiliency under reconnect decorator

bsenthilr:
> Minor change, good to go.

better to do add rate limiting log support on next PR.  That would be handy for several situations.
one reference (lets look for the most adopted library and plumb it in):
`from log_rate_limit import StreamRateLimitFilter
logger.addFilter(StreamRateLimitFilter(period_sec=30))
`

openfl-workspace/workspace/plan/defaults/network.yaml
```diff
@@ -7,3 +7,4 @@ settings:
     client_reconnect_interval  : 5
     require_client_auth        : True
     cert_folder                : cert
+    enable_atomic_connections  : False
```
ishaileshpant, line 10: Would be great to add a section in document detailing this new flag in network defaults, including default behavior change of atomic decorator

ishaileshpant (approved):
apart from updates to documents LGTM

openfl-workspace/workspace/plan/defaults/network.yaml
bsenthilr: It is generally hard to explain such custom options and we should look for options to eliminate this with grpc hardening. If we enable keep alive we might not require this option. Lets tentatively plan to eliminate this and stick to grpc norms.

openfl/transport/grpc/aggregator_client.py
bsenthilr: lets take this action item on next PR ideally via rate limited logs.

PR #1391, opened 2025-02-19
Changes to reduce memory footprint in Aggregator

snehal-das:
Changes made:

- Explicit deletion of local variables reduces the run-time footprint as the garbage collector
- Moved the measurement of memory to after tensor_db::clean_up() to get a clearer picture of memory usage.
- Added garbage collection at the end of round post db clean_up to free up unused memory.

### DIFF
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -3,10 +3,12 @@
 
 """Aggregator module."""
 
+import ctypes
+import gc
 import logging
+import os
 import queue
 import time
-import gc
 from threading import Lock
 from typing import List, Optional
 
@@ -20,6 +22,7 @@ from openfl.protocols.base_pb2 import NamedTensor
 from openfl.utilities import TaskResultKey, TensorKey, change_tags
 
 logger = logging.getLogger(__name__)
+
 
 class Aggregator:
     """An Aggregator is the central node in federated learning.
@@ -62,6 +65,7 @@ class Aggregator:
     .. note::
         - plan setting
     """
+
     def __init__(
         self,
         aggregator_uuid,
@@ -802,7 +806,6 @@ class Aggregator:
 
             self._end_of_round_with_stragglers_check()
 
-
     def _end_of_round_with_stragglers_check(self):
         """
         Checks if the minimum required collaborators have reported their results,
@@ -1120,6 +1123,13 @@ class Aggregator:
         self.stragglers = []
         # resetting collaborators_done for next round
         self.collaborators_done = []
+        self.tensor_db.clean_up(self.db_store_rounds)
+        self.release_memory()
+        # End of round callbacks.
+        self.callbacks.on_round_end(self.round_number, logs)
+
+        # Reset straggler handling policy for the next round.
+        self.straggler_handling_policy.reset_policy_for_round()
 
         # TODO This needs to be fixed!
         if self._time_to_quit():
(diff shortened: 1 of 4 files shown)


openfl/component/aggregator/aggregator.py
```diff
@@ -1131,8 +1129,12 @@ def _end_of_round_check(self):
             # https://github.com/securefederatedai/openfl/pull/1195#discussion_r1879479537
             self.callbacks.on_round_begin(self.round_number)
 
-        # Cleaning tensor db
         self.tensor_db.clean_up(self.db_store_rounds)
+        gc.collect()
+
+        # End of round callbacks.
+        self.callbacks.on_round_end(self.round_number, logs)
```
theakshaypant: Shouldn't `on_round_end` callback be called before the next `on_round_begin` (line 1130)

theakshaypant (commented):
Would also suggest running formatting script as the pipeline is failing. 
```
sh scripts/format.sh
```

rahulga1:
@snehal-das can you run it with memory_callback with large number of rounds to see the impact of this change?

MasterSkepticista (commented):
@snehal-das Does this PR "fix" memory leak once and for all, or it only optimizes memory usage?

openfl/component/aggregator/aggregator.py
snehal-das: Yes, I've corrected the order in the latest commit. Thanks for pointing it out.

snehal-das:
> can you run it with memory_callback with large number of rounds to see the impact of this change?

@rahulga1 I've run these changes for 100-200 rounds with torch/mnist and torch/histology tests. I've also tested it with 32 collaborators to evaluate the impact of the change.

> Does this PR "fix" memory leak once and for all, or it only optimizes memory usage?

@MasterSkepticista The current state of code does not fix memory issues. It only reduces the footprint of the aggregator, but there is still a growth in overall memory usage with time, albeit at a slower rate. There are a couple more approaches being considered for this:
1. Reducing memory fragmentation: Tracking fordblks show that memory fragmentation is increasing proportional to the number of rounds.
2. Delegating memory intensive tasks to child processes - Return the memory allocated at the time of execution back to the allocator.

MasterSkepticista:
> > can you run it with memory_callback with large number of rounds to see the impact of this change?
> 
> @rahulga1 I've run these changes for 100-200 rounds with torch/mnist and torch/histology tests. I've also tested it with 32 collaborators to evaluate the impact of the change.
> 
> > Does this PR "fix" memory leak once and for all, or it only optimizes memory usage?
> 
> @MasterSkepticista The current state of code does not fix memory issues. It only reduces the footprint of the aggregator, but there is still a growth in overall memory usage with time, albeit at a slower rate. There are a couple more approaches being considered for this:
> 
> 1. Reducing memory fragmentation: Tracking fordblks show that memory fragmentation is increasing proportional to the number of rounds.
> 2. Delegating memory intensive tasks to child processes - Return the memory allocated at the time of execution back to the allocator.

Thanks for the clarification. In addition to our discussion offline, I suggest doing the next experiments with a "reduced" set of patches, i.e., disabling changes that are unlikely to be part of the actual fix (like calling `gc.collect()` or `del obj` calls). Mainly because these interfere with other changes we plan on testing and while useful for debugging, not a python best practice.

MasterSkepticista:
Since code has diverged (merge conflicts) and we want to avoid using `del` or `gc.collect()` calls, I am closing this PR. There will be a follow up that addresses memory issues

PR #1398, opened 2025-02-24
Trivy Nightly Scan Enhancements

rajithkrishnegowda:
## Summary
Enhancing the Trivy workflow to run nightly scans, update packages if vulnerabilities are found, and rescan to verify the resolution of vulnerabilities.

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This pull request enhances the existing Trivy workflow by:

- Running nightly scans to detect vulnerabilities.
- Updating packages if any vulnerabilities are detected during the initial scan.
- Rescanning with Trivy to verify if the vulnerabilities have been resolved.

## Testing
Testing done in fork repo. Sharing link for reference: GitHub Actions Run

https://github.com/rajithkrishnegowda/openfl/actions/runs/13514941177


theakshaypant (commented):
Why are we adding an unresolved vulnerability to `.trivyignore`?
As per my understanding, wouldn't this lead to confusion when a fix is available at a later stage given that trivy would not be flagging it?

rajithkrishnegowda:
> Why are we adding an unresolved vulnerability to `.trivyignore`?
> As per my understanding, wouldn't this lead to confusion when a fix is available at a later stage given that trivy would not be flagging it?

 In some cases, the reported vulnerabilities might be false positives, and some of the packages are related to env specific not project specific hence adding it to .trivyignore

rajithkrishnegowda:
> Why are we adding an unresolved vulnerability to `.trivyignore`? As per my understanding, wouldn't this lead to confusion when a fix is available at a later stage given that trivy would not be flagging it?

Pushed trivy workflow to nightly job and removed adding to vulnerability to `.trivyignore`

rahulga1 (approved):

tanwarsh (approved):

PR #1403, opened 2025-02-25
Quick Fix - Pipeline not running

payalcha:
This PR is to fix the pipeline. It is not running as small issue got pushed in the below PR.
Trivy is now supposed to be run as nightly job instead in PR pipeline. 

https://github.com/securefederatedai/openfl/pull/1398


theakshaypant (approved):

theakshaypant (changes requested):
Trivy gets called in pr_pipeline again. We want that to only run nightly.

theakshaypant (approved):
Thanks!

noopurintel (approved):

PR #1405, opened 2025-02-27
Fix FedEVal changes as per documentation

payalcha:
## Summary
Changes needed in FedEVAL test to match it as per the documentation.

## Type of Change (Mandatory)
- Feature enhancement  

1. Earlier evaluation task-group needs to be added in plan and other task-group were supposed to be removed. But now as per documentation no such changes needed in the plan.
2. Earlier round must be passes as 1, now as per the documentation no need to change the plan for num_rounds.

## Description (Mandatory)
Make FedEval test more effective by changing the steps as per documentation.

## Testing
https://github.com/payalcha/openfl/actions/runs/13560108651 
https://github.com/payalcha/openfl/actions/runs/13560124882

### DIFF
--- a/tests/end_to_end/models/model_owner.py
+++ b/tests/end_to_end/models/model_owner.py
@@ -122,7 +122,7 @@ class ModelOwner():
             raise e
         return True
 
-    def modify_plan(self, param_config, plan_path, eval_scope=False):
+    def modify_plan(self, param_config, plan_path):
         """
         Modify the plan to train the model
         Args:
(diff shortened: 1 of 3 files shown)


noopurintel (approved):

kminhta (approved):

ishaileshpant (approved):
LGTM - thanks for aligning these changes

tanwarsh (approved):

MasterSkepticista (commented):
My suggestion would be to add evaluation task group in the plan defaults. Later, simply switch `--task_group` as `evaluation`. 

That is how we intend a user use it. This will also be the case when analytics task makes its way into OpenFL.

payalcha:
> My suggestion would be to add evaluation task group in the plan defaults. Later, simply switch `--task_group` as `evaluation`.
> 
> That is how we intend a user use it. This will also be the case when analytics task makes its way into OpenFL.

This is already done.

PR #1409, opened 2025-02-27
Do Not Review: Remove Workflow Interface Tests workflow and related test folder

noopurintel:
## Type of Change (Mandatory)
- Code improvement and cleanup

## Description (Mandatory)
**Workflow Functional E2E** which is already a part of PR pipeline covers all the tests present in **Workflow Interface Tests** in Pytest framework, thus removing the workflow as well as folder `LocalRuntime` specific to it.

Earlier, both of these were kept just to ensure smooth transition and stability of pytest based workflow.

## Testing
PR pipeline already covers Functional E2E workflow.


teoparvanov:
Hi @noopurintel , could you update the PR description as per the template?

theakshaypant (commented):
This means task runner and workflow API tests will be run for all PRs, right? 
Using labels is one way to reduce the number of tests run per PR. Is it possible to run tests based on the directories where changes have been made, thus reducing the CI run time and not running unnecessary tests?

scngupta-dsp (commented):
While this PR does clean-up duplication of Workflow API test cases, there are two folders which now contain end-to-end test cases
- `tests/github`, and
- `tests/end_to_end`

As per my understanding the old convention was 
- `tests/github`: End to End test cases 
- `tests/openfl`: Module level test cases

Request to please harmonize the test directory also to avoid any confusion

PR #1413, opened 2025-02-28
Remove implicit package installs

MasterSkepticista:
This PR moves package management to the user, instead of forcing a pip install during `fx workspace create` and `fx plan initialize`.

**Reasons:**
* `pip` is not necessarily a package manager of choice.
* Every run of this command forces an install of specific package versions, overriding user choices if any.
* separation of concerns.

**Tests:** CI.

**Misc:** This may impact OpenFL-Security launch script. It will be modified accordingly post finalization of the changes here.

**Additional context:** https://github.com/securefederatedai/openfl/issues/73

**Documentation:** [Link](https://openfl.readthedocs.io/en/karansh1-manual_pkg_install/index.html)

This is part 1 (of n). In an upcoming PR, `data.yaml` will be removed in favor of directly providing `--data-path` during collaborator start.

### DIFF
--- a/.github/workflows/tr_docker_gramine_direct.yml
+++ b/.github/workflows/tr_docker_gramine_direct.yml
@@ -33,6 +33,7 @@ jobs:
       run: |
         fx workspace create --prefix example_workspace --template keras/mnist
         cd example_workspace
+        pip install -r requirements.txt
         fx plan initialize -a localhost
         
         # Disable tensorboard logging as multiprocessing is not supported in Gramine
--- a/.github/workflows/tr_docker_native.yml
+++ b/.github/workflows/tr_docker_native.yml
@@ -33,6 +33,7 @@ jobs:
       run: |
         fx workspace create --prefix example_workspace --template keras/mnist
         cd example_workspace
+        pip install -r requirements.txt
         fx plan initialize -a localhost
         fx workspace dockerize --base-image openfl --save
 
--- a/docs/tutorials/taskrunner.ipynb
+++ b/docs/tutorials/taskrunner.ipynb
@@ -31,15 +31,12 @@
     {
       "cell_type": "code",
       "execution_count": null,
-      "metadata": {
-        "tags": [
-          "hide-output"
-        ]
-      },
+      "metadata": {},
       "outputs": [],
       "source": [
         "!fx workspace create --prefix ./mnist_example --template keras/mnist\n",
-        "%cd ./mnist_example"
+        "%cd ./mnist_example\n",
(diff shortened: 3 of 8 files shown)


noopurintel (approved):

kminhta (approved):
I think this is a reasonable change. It adds an expected amount of flexibility back to the user and may even reduce some dependency issues if users can freely manage their env packages rather than rely entirely on `pip install -r requirements.txt` (which should be the case, IMO)

rahulga1 (approved):

tanwarsh (approved):

payalcha:
One issue I foresee is, if user mesh-up with requirements, user will know it at until aggregator/collaborator starts. 
One other issue is we are not exiting the process in case of exception. So the process will keep running even if import error comes.

tanwarsh:
Although I agree that we should not install packages during `fx workspace create `and `fx plan initialize`, this could lead to many errors when users run experiments if requirements are missing. For example, libraries like `nibabel` that are used in specific use cases. This will result in a poor user experience. We should check for requirements, perhaps after `fx plan initialize`, to ensure that at least feature-related requirements are satisfied and errors are minimised when experiments start.

MasterSkepticista:
@payalcha @tanwarsh 

While I agree that a missing package is an inconvenience, the fix is generally very straightforward. In real-world deployments, dependency installations are a common step and explicitly stated. In addition `pip` not being the most popular package manager means we make assumptions that can break a user's package management. Our goal with a framework is also to educate users what common practices are and not automate everything unless it is a standard practice.

Overall, it is preferable for functions to be idempotent and for them to be consistent with what a user would expect it to do by its _name_. It is strange for `fx workspace create` or `fx plan initialize` to override package installs (which is what @rstoki tried to convey in #73)

Related: [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

psfoley (changes requested):
I think this will break [GaNDLF's](https://github.com/mlcommons/GaNDLF) CI in it's current state: 

https://github.com/mlcommons/GaNDLF/blob/master/.github/workflows/openfl-test.yml#L55
See here for the missing call to `pip install `: https://github.com/securefederatedai/openfl/blob/develop/tests/github/test_gandlf.py#L50-L63

Once you can confirm that this is either tested, or you add a commit to add the pip install, I think this is ready to merge.

MasterSkepticista:
Good catch, resolved in https://github.com/securefederatedai/openfl/pull/1413/commits/f76bd7b34c81ea044c60941be27428fc86825a6b

PR #1420, opened 2025-03-03
[taskrunner] Secure Aggregation cleanup

theakshaypant:
## Summary
Perform cleanup of secure aggregation code in openfl based on the checklist below. 

## Type of Change (Mandatory)
- Documentation update  
- Security improvement
- Restructuring

## Description (Mandatory)
- [x] Remove `pycryptodome` from openfl dependencies.
- [x] Rename `openfl.utilities.secagg.setup`
- [x] Remove "bloat" from aggregator and collaborator.
- [x] Do not mask metrics that are shared by the collaborator

## Testing
Normal run
- Result: Committed model for round 9, saved 8 model tensors and 26 next round model tensors  with best_score 0.9884999990463257
- Logs: [normal.log](https://github.com/user-attachments/files/19223860/normal.log)
- Installed packages (no `pycryptodome`): [normal-pip-list.txt](https://github.com/user-attachments/files/19223862/normal-pip-list.txt)

Secagg run
- Result: Committed model for round 9, saved 8 model tensors and 26 next round model tensors  with best_score 0.990200012922287
- Logs: [secagg.log](https://github.com/user-attachments/files/19223868/secagg.log)
- Installed packages (`pycrytodome` installed before running the experiment): [secagg-pip-list.txt](https://github.com/user-attachments/files/19223873/secagg-pip-list.txt)


openfl/utilities/secagg/bootstrap.py
```diff
@@ -5,20 +5,28 @@
...
+    from openfl.utilities.secagg import (
+        calculate_shared_mask,
+        generate_agreed_key,
+        reconstruct_secret,
+    )
+except Exception as error:
+    raise Exception(
+        "'pycryptodome' not installed."
+        f"This package is necessary when secure aggregation is enabled: {error}"
+    )
```
teoparvanov: How do we know this general `Exception` is caused by a missing `pycryptodome` dependency?

openfl/utilities/secagg/crypto.py
```diff
@@ -195,3 +197,59 @@ def calculate_mask(collaborator_index, agreed_keys, private_seed) -> float:
...
+    # Fetch private mask from tensor db if not already fetched.
+    if not private_mask:
+        private_mask = tensor_db.get_tensor_from_cache(
+            TensorKey("private_mask", collaborator_name, -1, False, ("secagg",))
+        )[0]
+    # Fetch shared mask from tensor db if not alreday fetched.
+    if not shared_mask:
+        shared_mask = tensor_db.get_tensor_from_cache(
+            TensorKey("shared_mask", collaborator_name, -1, False, ("secagg",))
+        )[0]
```
teoparvanov: IMHO it's not great to have a dependency on the TensorDB in a module dedicated to cryptography. Could you consider moving the code for retrieving the private and shared masks back to the collaborator module?

teoparvanov (commented):
Thanks @theakshaypant, I believe these changes are a move in the right direction! A couple of comments from my initial read-through:

docs/about/features_index/secure_aggregation.rst
```diff
@@ -12,6 +12,7 @@ TaskRunner API
 -------------------------------------
 
 OpenFL treats SecAgg as a core security feature and can be enabled for any experiment by simply modifying the plan.
+**NOTE**: `pycryptodome <https://pypi.org/project/pycryptodome/>`_ is a required dependency that must be installed on the participant nodes before starting the experiment.
```
payalcha: Please add limitations as well i.e Straggler policies will not work with secure aggregation. Also resiliency or restart of aggregator or collaborator during experiment may cause issues.

psfoley:
@theakshaypant I haven't completed a full review yet (as this is still WIP), but it's clearly moving in the right direction. I have a question about `pycryptodome`. I see it's use in https://github.com/securefederatedai/openfl/blob/a208651689b603776172617b422709451ba96d0f/openfl/utilities/secagg/crypto.py#L14-L15
for ciphertext and mask generation. Instead could you use the `cryptography` package primitives that we already import? https://cryptography.io/en/latest/hazmat/primitives/

That would eliminate the need for the package detection code and make user's lives a bit easier

theakshaypant:
> @theakshaypant I haven't completed a full review yet (as this is still WIP), but it's clearly moving in the right direction. I have a question about `pycryptodome`. I see it's use in
> 
> https://github.com/securefederatedai/openfl/blob/a208651689b603776172617b422709451ba96d0f/openfl/utilities/secagg/crypto.py#L14-L15
> 
> 
> for ciphertext and mask generation. Instead could you use the `cryptography` package primitives that we already import? https://cryptography.io/en/latest/hazmat/primitives/
> That would eliminate the need for the package detection code and make user's lives a bit easier

Thanks @psfoley! I can replace the `pycroptodome` usage in all utility files except `openfl.utilities.secagg.shamir`. `cryptography` does not offer an equivalent package for secret sharing so we will still need `pycrptodome`.

openfl/utilities/secagg/crypto.py
theakshaypant: Considering there a few utility functions for calculating masks and such, we can move it to a separate secagg submodule, perhaps `openfl.utilities.secagg.masking`, what do you think?

openfl/utilities/secagg/bootstrap.py
theakshaypant: Fixed in d3a04a98dd226928d3f54e16ef902b98c3653c59.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Thanks for pointing it out, added in 750e161cabce0502e632a80e0e3f0d3c9a17307b.

openfl/utilities/secagg/crypto.py
teoparvanov: I'd prefer to keep this snippet with the collaborator code, as it simply reads the masks from the TensorDB (rather than doing actual cryptography).

theakshaypant: Moved it in 75c93fa96fbd45f4fc95466748a5001f84dc6750.

rahulga1 (approved):

openfl/callbacks/secure_aggregation.py
```diff
@@ -39,6 +31,19 @@ class SecAggBootstrapping(Callback):
...
+        import pkg_resources
+
+        try:
+            pkg_resources.get_distribution(_required_package)
+        except pkg_resources.DistributionNotFound:
+            raise Exception(
+                f"'{_required_package}' not installed."
+                "This package is necessary when secure aggregation is enabled."
+            )
+
```
MasterSkepticista: Does this work?

```python
try:
  import pycryptodome
except ImportError:
  raise ImportError("`pycryptodome` must be installed")
```

openfl/utilities/secagg/bootstrap.py
```diff
@@ -269,3 +323,23 @@ def _save_tensors(self):
...
+            named_tensor.name,
+            self._aggregator_uuid,
+            named_tensor.round_number,
+            named_tensor.report,
+            tuple(named_tensor.tags),
+        )
+        _, _, _, _, tags = tensor_key
+        # Secure aggregation setup stage key
+        if "secagg" in tags:
+            nparray = json.loads(named_tensor.data_bytes)
```
MasterSkepticista, line 337: I see a custom `NumpyEncoder` being used to serialize arrays. Have you tried using `np.ndarray.tobytes(...)` directly?

openfl/component/aggregator/aggregator.py
```diff
@@ -209,7 +208,9 @@ def __init__(
         self.collaborator_tensor_results = {}  # {TensorKey: nparray}}
         self._secure_aggregation_enabled = secure_aggregation
         if self._secure_aggregation_enabled:
-            self.secagg = secagg_setup(self.uuid, self.authorized_cols, self.tensor_db)
+            from openfl.utilities.secagg.bootstrap import SecAggSetup
+
+            self.secagg = SecAggSetup(self.uuid, self.authorized_cols, self.tensor_db)
 
         # Callbacks
         self.callbacks = callbacks_module.CallbackList(
```
MasterSkepticista: I see two declarations of `self.callbacks`:

First:

https://github.com/securefederatedai/openfl/blob/063fe03df4de622e3296d5e319560bffd7cf078d/openfl/component/aggregator/aggregator.py#L188-L194

Second:
https://github.com/securefederatedai/openfl/blob/063fe03df4de622e3296d5e319560bffd7cf078d/openfl/component/aggregator/aggregator.py#L214-L223

openfl/utilities/secagg/bootstrap.py
```diff
@@ -29,7 +43,47 @@ def __init__(self, aggregator_uuid, collaborator_list, tensor_db):
...
+        secagg_setup = False
+        for named_tensor in named_tensors:
+            # Check if the tensor belongs to one from secure aggregation
+            # setup stages.
+            if "secagg" not in tuple(named_tensor.tags):
+                continue
+            else:
+                secagg_setup = True
+                # Process and save tensor to local tensor db.
+                self._named_tensor_to_nparray(named_tensor)
```
teoparvanov: I think the name of this method is misleading; the comment helps to some extent, although I believe there's a better way:

Instead, I'd suggest removing the comment and simply "telling the truth": `self._save_secagg_tensor()` 😉

openfl/utilities/secagg/bootstrap.py
```diff
@@ -29,7 +43,47 @@ def __init__(self, aggregator_uuid, collaborator_list, tensor_db):
...
+        for the current key. If all collaborators have sent their data, it
+        proceeds with aggregation for the key.
+
+        Args:
+            named_tensors (list): A list of named tensors to be processed.
+
+        Returns:
+            bool: True if the setup is complete or if the tensor does not
+                belong to secure aggregation setup, otherwise waits for all
+                collaborators.
```
teoparvanov: Could you be a bit more explicit when the result is `True`, and when is it `False` ?

openfl/component/collaborator/collaborator.py
```diff
@@ -549,28 +545,21 @@ def named_tensor_to_nparray(self, named_tensor):
 
         return decompressed_nparray
 
-    def _secure_aggregation_masking(self, global_output_tensor_dict, task_name):
+    def _calulcate_masked_input_vectors(
+        self,
+        tensor_dict,
+    ):
```
teoparvanov: I suggest simply calling this method `_apply_masks()`, as it effectively "mutates" the `tensor_dict` passed as a parameter.

teoparvanov (approved):
LGTM, with a couple of minor comments. Thanks @theakshaypant !

openfl/component/collaborator/collaborator.py
theakshaypant: Changed method name in c316d0c0

openfl/utilities/secagg/bootstrap.py
theakshaypant: Made the return value distinction clear in c316d0c0

openfl/utilities/secagg/bootstrap.py
theakshaypant: Changed method name in c316d0c0

docs/about/features_index/secure_aggregation.rst
```diff
@@ -12,6 +12,10 @@ TaskRunner API
 -------------------------------------
 
 OpenFL treats SecAgg as a core security feature and can be enabled for any experiment by simply modifying the plan.
+**NOTE**: 
+- `pycryptodome <https://pypi.org/project/pycryptodome/>`_ is a required dependency that must be installed on the participant nodes before starting the experiment.
+- The `straggler_handling_policy` is not yet supported with secure aggregation and will be addressed in future enhancements.
```
MasterSkepticista: What is the behavior when both are enabled? Does the code check and flag this incompatibility?

docs/about/features_index/secure_aggregation.rst
```diff
@@ -12,6 +12,10 @@ TaskRunner API
...
+- The impact of secure aggregation on the aggregator and/or collaborator will be assessed, and additional resiliency features will be introduced in subsequent updates.
```
MasterSkepticista: Could you please explain what this means?

openfl/utilities/secagg/bootstrap.py
```diff
@@ -29,7 +43,46 @@ def __init__(self, aggregator_uuid, collaborator_list, tensor_db):
...
+                False otherwise.
+        """
+        secagg_setup = False
+        for named_tensor in named_tensors:
+            # Check if the tensor belongs to one from secure aggregation
+            # setup stages.
+            if "secagg" not in tuple(named_tensor.tags):
+                continue
+            else:
+                secagg_setup = True
```
MasterSkepticista: Can be rewritten without an `else` block by inverting the `if` statement.

openfl/component/aggregator/aggregator.py
```diff
@@ -1103,11 +1097,11 @@ def _compute_validation_related_task_metrics(self, task_name) -> dict:
             agg_tensor_key = TensorKey(tensor_name, origin, round_number, report, new_tags)
             # Check if secure aggregation is enabled, set aggregation function.
             agg_function = (
-                task_agg_function
-                if "metric" not in tags
+                WeightedAverage()
+                if "metric" in tags
                 else SecureWeightedAverage()
                 if self._secure_aggregation_enabled
```
MasterSkepticista: Isn't the goal of secure aggregation to compute a weighted average to unmask the values first? This implies that all collaborator updates are the same. Beyond which, the standard flow should work. A weighted average of tensors where all tensors are same, is the tensor itself. 

Motivation is to not do a lot of conditional branching (here and everywhere in the code)

MasterSkepticista (changes requested):
Do we have unit tests to verify this works? (Non FL settings)

openfl/callbacks/secure_aggregation.py
theakshaypant: Raises an issue with Flake [F401](https://www.flake8rules.com/rules/F401.html).
```
openfl/callbacks/secure_aggregation.py:37:20: F401 `Crypto` imported but unused; consider using `importlib.util.find_spec` to test for availability
   |
35 |         # Check if pycryptodome is installed.
36 |         try:
37 |             import Crypto
   |                    ^^^^^^ F401
38 |         except ImportError:
39 |             raise Exception(
   |
   = help: Remove unused import: `Crypto`
```

We can either stick with `pkg_resources.get_distribution` or change it to the flake suggested `importlib.util.find_spec` which is already used in the repo.

openfl/utilities/secagg/bootstrap.py
theakshaypant: I had given that a shot but was having issues keeping the deserialisation code clean. 
I was trying to use [`numpy.frombuffer(...)`](https://numpy.org/doc/2.1/reference/generated/numpy.frombuffer.html) but given the different reshaping involved with each step, I went ahead with using a custom `NumpyEncoder` which can be deserialised directly using `json.loads(...)`.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Missed a word here
The impact of secure aggregation on the aggregator and/or collaborator **restart** will be assessed, and additional resiliency features will be introduced in subsequent updates.

Edit: Update in e39822493bf7b87fb923eada3eaa71b8903f65f5.

openfl/utilities/secagg/bootstrap.py
MasterSkepticista: We use `np.frombuffer` and `np.ndarray.tobytes()` within compression/decompression pipelines. I suggest maintaining the same for consistency.

openfl/callbacks/secure_aggregation.py
MasterSkepticista: Then I suggest `find_spec`. We have had bad experiences with `pkg_resources` for unrelated reasons

openfl/utilities/secagg/bootstrap.py
theakshaypant: Modified in 3e3ad80bc100ad0ac6fddbbea6df655eb147d4cd.

openfl/component/aggregator/aggregator.py
theakshaypant: Removed duplicate declaration in b0d1624ad790ce7646859a8cd37090522fa5b84f.

openfl/callbacks/secure_aggregation.py
theakshaypant: Updated in 6141b5f3

openfl/component/aggregator/aggregator.py
theakshaypant: Restructured this conditional in 754070f7dbdcdb879458fb5c968dd9a7e3603561 to increase readability.

openfl/utilities/secagg/bootstrap.py
theakshaypant: Tried making the change to use `np.frombuffer` and `np.ndarray.tobytes` passing the size and type using the `int_list` in metadata, something like this.
```
            type = int(re.findall(r"\d+", str(nparray.dtype))[0])
            named_tensor = utils.construct_named_tensor(
                tensor_key,
                nparray.tobytes(),
                [{"int_list": list(nparray.shape)}, {"int_list": [type]}],
                lossless=True,
            )
```
and
```
        meta = [metadata.int_list for metadata in tensor.transformer_metadata]
        nparray_type = f"|S{meta[1][0]}" if "public_key" not in key_name else "<U112"

        np.frombuffer(tensor.data_bytes, dtype=nparray_type).reshape(meta[0])
```

After this is done, there would be further processing of the tensors required as all the element types are not homogenous. 

In my opinion, this hampers the code readability and we can stick to `json` encoding for the first iteration and take this up perhaps during codec restructuring in the near future as discussed.

docs/about/features_index/secure_aggregation.rst
theakshaypant: There are no code checks present at the moment. 
If both are enabled, we will face issues during the bootstrap itself.

tanwarsh (approved):
Thanks @theakshaypant for the cleanup. 👍

docs/about/features_index/secure_aggregation.rst
MasterSkepticista: I suggest putting a check in place. I believe we can do this in the aggregator `__init__` (?)

It is better for code to fail than tolerate undefined behavior (in addition to the fact that users do not often review documentation for all edge cases)

openfl/component/aggregator/aggregator.py
```diff
@@ -209,18 +208,10 @@ def __init__(
         self.collaborator_tensor_results = {}  # {TensorKey: nparray}}
```
tanwarsh: this line is repeated as well. please remove it.

tanwarsh (changes requested):

openfl/component/aggregator/aggregator.py
theakshaypant: Removed in ba9e0215.

docs/about/features_index/secure_aggregation.rst
psfoley: A check is warranted here. Not everyone will read the documentation thoroughly. For the immediate PR, the check can either be added in `plan.py` or `aggregator.py`. As an aside, I know there's other cases where we have conflicting parameters that do not explicitly fail. In a future PR (1.9), we can check for known configuration mismatches at time of plan initialization, and provide relevant recommendations for model author's to fix them.

Also in this documentation page it would be worth showing in the keras mnist plan that `straggler_handling_policy: None` to connect the plan configuration to the recommendation you are making in this statement

psfoley (changes requested):
Thanks @theakshaypant for the updates to Secure Aggregation! I see this as an improvement on the original, and sets us up for building upon these capabilities in 1.9.

I agree with @MasterSkepticista's point that we should have an explicit check to make sure that secure aggregation and straggler handling aren't enabled at the same time. Once that's added, this is ready to merge IMO.

docs/about/features_index/secure_aggregation.rst
theakshaypant: Having discussed this with @MasterSkepticista, created a `NoPolicy` straggler handling in #1461 which would serve as the default policy (wait for all collaborators to send results) and secure aggregation check would also be done with the said policy. 

cc @teoparvanov

teoparvanov: I agree, and I've approved https://github.com/securefederatedai/openfl/pull/1461.

So, the only remaining thing in this PR would be to check which straggler policy has been selected if secure aggregation is enabled, right @theakshaypant ?

theakshaypant: Updated docs in 69e2244b and added plan verification in 35afe2ad. 
`fx plan initialize` will raise an Exception if straggler handling other than `WaitForAllPolicy` is used with secure aggregation.

theakshaypant:
> I agree with @MasterSkepticista's point that we should have an explicit check to make sure that secure aggregation and straggler handling aren't enabled at the same time. Once that's added, this is ready to merge IMO.

Resharing [comment](https://github.com/securefederatedai/openfl/pull/1420#discussion_r2005386475) from above thread:
>  Updated docs in https://github.com/securefederatedai/openfl/commit/69e2244bfef725e0f4909f6d9093a37da8fb7fa3 and added plan verification in https://github.com/securefederatedai/openfl/commit/35afe2ad838e0620de6e2f3841e839b06ebeac0e.
fx plan initialize will raise an Exception if straggler handling other than WaitForAllPolicy is used with secure aggregation.

Edit: CI test result can be found [here](https://github.com/securefederatedai/openfl/actions/runs/13968131386/job/39103182256?pr=1420).

openfl/federated/plan/plan.py
```diff
@@ -137,6 +137,7 @@ def parse(
             plan.authorized_cols = Plan.load(cols_config_path).get("collaborators", [])
 
             Plan._load_collaborator_data_paths(plan, data_config_path)
+            Plan.verify(plan)
```
teoparvanov: I'd prefer the non-static `plan.verify()` approach, by analogy with `plan.resolve()`. Is there a strong reason to use a @staticmethod instead?

openfl/federated/plan/plan.py
```diff
@@ -677,3 +678,33 @@ def save_model_to_state_file(self, tensor_dict, round_number, output_path):
...
+                {
+                    TEMPLATE: "openfl.component.aggregator.straggler_handling.WaitForAllPolicy",
+                    SETTINGS: {},
+                },
+            )
+            if straggler_handling_policy.get(TEMPLATE) not in [
+                "openfl.component.WaitForAllPolicy",
+                "openfl.component.aggregator.WaitForAllPolicy",
+                "openfl.component.aggregator.straggler_handling.WaitForAllPolicy",
+            ]:
```
teoparvanov: Why do we need to check for all these alternative packages, instead of just:
```python
if straggler_handling_policy.get(TEMPLATE) is not "openfl.component.aggregator.straggler_handling.WaitForAllPolicy":
    raise Exception(
        "Only WaitForAllPolicy straggler handling is supported with secure aggregation."
    )
```
?

theakshaypant: Essentially, in the plan manifest, we specify the class that should be used for the object in `template`. In this case, what straggler class should be used/initialised. 
The reason for having this list is that the `WaitForAllPolicy` class can be used in all of these ways and are equivalent. By checking for all the three ways of "importing", we ensure that the check happens properly and we do not raise an exception if the user has specified `openfl.component.WaitForAllPolicy` or `openfl.component.aggregator.WaitForAllPolicy` since those are legitimate ways of importing the class. 

Edit: Using `"openfl.component.aggregator.straggler_handling.WaitForAllPolicy"` as the default value since it looks cleaner and more descriptive IMO.

openfl/federated/plan/plan.py
theakshaypant: Changed to non-static in 0172973a. 

There was no strong reason to use but trying to match the pre-existing static methods such as  `_ensure_settings_in_sections`, `_load_defaults`, `_load_collaborator_data_paths`, apart from `load` and `dump` in the class.

teoparvanov: Thanks, I believe encapsulation helps with readability and reasoning about the code's correctness.

openfl/federated/plan/plan.py
teoparvanov: Thanks for clarifying, @theakshaypant !

teoparvanov (approved):

theakshaypant:
Test for initialize failing when a different (Edit: `PercentagePolicy`) policy is used can be found [here](https://github.com/theakshaypant/openfl/actions/runs/13970380380/job/39110384919). 
```
ERROR    tests.end_to_end.utils.ssh_helper:ssh_helper.py:106 Error output: Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.16/x64/bin/fx", line 8, in <module>
    sys.exit(entry())
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/interface/cli.py", line 310, in entry
    error_handler(e)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/interface/cli.py", line 229, in error_handler
    raise error
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/interface/cli.py", line 308, in entry
    cli(max_content_width=120)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/interface/cli.py", line 131, in invoke
    super().invoke(ctx)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/interface/plan.py", line 126, in initialize
    plan = Plan.parse(
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/federated/plan/plan.py", line 140, in parse
    plan.verify()
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/openfl/federated/plan/plan.py", line 704, in verify
    raise Exception(
Exception: Only WaitForAllPolicy straggler handling is supported with secure aggregation.
```

Issue #1406, opened 2025-02-27
1001_Workspace_Creation_from_JupyterNotebook.ipynb is not functioning properly


ishant162:
**Describe the bug**
Encountering a `TypeError `while running the `WorkspaceExport.export(...)` cell in `1001_Workspace_Creation_from_JupyterNotebook.ipynb.`
 
**To Reproduce**
1. Activate experimental features (`fx experimental activate`)
2. Go to `openfl-tutorials/experimental/workflow/1001_Workspace_Creation_from_JupyterNotebook.ipynb`
3. Scroll down to last cell containing 
```
WorkspaceExport.export(
    notebook_path='./1001_Workspace_Creation_from_JupyterNotebook.ipynb',
    output_workspace=f"/home/{os.environ['USER']}/generated-workspace"
)
```
3. Run the cell
4. See Error
 
**Expected behavior**
The cell should execute without any issues
 
**Error Logs**
```
TypeError                                 Traceback (most recent call last)
Cell In[1], line 4
      1 import os
      2 from openfl.experimental.workflow.workspace_export import WorkspaceExport
----> 4 WorkspaceExport.export(
      5     notebook_path='./1001_Workspace_Creation_from_JupyterNotebook.ipynb',
      6     output_workspace=f"/home/{os.environ['USER']}/generated-workspace"
      7 )
 
File ~/dev_openfl/openfl/openfl/experimental/workflow/workspace_export/export.py:316, in WorkspaceExport.export(cls, notebook_path, output_workspace)
    314 instance.generate_requirements()
    315 instance.generate_plan_yaml()
--> 316 instance.generate_data_yaml()
    317 print_tree(output_workspace, level=2)
 
File ~/dev_openfl/openfl/openfl/experimental/workflow/workspace_export/export.py:449, in WorkspaceExport.generate_data_yaml(self)
    446 # Import flow class
    447 federated_flow_class = getattr(self.exported_script_module, self.flow_class_name)
--> 449 flow_name, runtime = self._find_runtime_instance(federated_flow_class)
    450 data_yaml = self.created_workspace_path.joinpath("plan", "data.yaml").resolve()
    451 data = self._read_or_initialize_yaml(data_yaml)
 
TypeError: cannot unpack non-iterable LocalRuntime object
```
 
**Desktop:**
- OS: Ubuntu (WSL)
- OS Version: 22.04.3 LTS
- Python: 3.10

**Additional context**
[...]

PR #1363, opened 2025-02-11
[Workflow Interface] Refactor FLSpec and Runtime to enhance modularity

ishant162:
**Background**
Proposal: #1317 

**Change Description**
- Updates to `FLSpec`, `LocalRuntime`, `FederatedRuntime` and `Aggregator` classes.
- Shift the `run()` method implementation from `FLSpec` to `Runtime`.
- Runtime now passes collaborator information to `FLSpec`, enabling foreach operations, necessitating changes to the execution flow.
- Updated `WorkspaceExport` functionality to align with the refactor changes.
- Documentation has been updated to align with the refactored implementation.

**Verification**
- Verified all testcases from `tests/github/experimental/workflow/LocalRuntime`.
- Verified all testcases from `tests/github/experimental/workflow/FederatedRuntime`.
- Verified Tutorials `101_MNIST` of `LocalRuntime` and `101_MNIST`, `301_MNIST_Watermarking` of `FederatedRuntime`.

**File Modifications**
- `docs/about/features_index/workflowinterface.rst`
- `openfl-tutorials/experimental/workflow/101_MNIST.ipynb`
- `openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb`
- `openfl-tutorials/experimental/workflow/FederatedRuntime/301_MNIST_Watermarking/workspace/MNIST_Watermarking.ipynb`
- `openfl/experimental/workflow/component/aggregator/aggregator.py`
- `openfl/experimental/workflow/interface/fl_spec.py`
- `openfl/experimental/workflow/runtime/federated_runtime.py`
- `openfl/experimental/workflow/runtime/local_runtime.py`
- `openfl/experimental/workflow/utilities/runtime_utils.py`
- `openfl/experimental/workflow/workspace_export/export.py`
- All testcases from `tests/github/experimental/workflow/LocalRuntime`.
- All testcases from `tests/github/experimental/workflow/FederatedRuntime`.

**Changelog:**

- [X] FLSpec Refactor.
- [X] LocalRuntime Refactor.
- [X] Update LocalRuntime Tutorial (101_MNIST) & Testcases.
- [X] FederatedRuntime Refactor.
- [X] Update FederatedRuntime Tutorial (101_MNIST, 301_MNIST_Watermarking) & Testcases.

**Known Issues**
- With ray backend, prints are appearing in previous cell (observed in the openfl/develop branch)

**Note:**
Following changes will be done in a follow-up PR:
[...]

PR #1421, opened 2025-03-04
[Workflow Interface] Fix for WorkspaceExport (Issue #1406)

refai06:
## Summary
This PR fixes an issue related with `WorkspaceExport`(#1406)

## Type of Change (Mandatory)

- Bug fix  

## Description (Mandatory)
Added flow instance name and modified the return statement to include both flow instance name and runtime instance 

## Testing

Validated below tutorial
- `openfl-tutorials/experimental/workflow/1001_Workspace_Creation_from_JupyterNotebook.ipynb`


theakshaypant (commented):
Since this would also affect `FLSpec` refactor, would the respective PR (#1363 ) cover that?

scngupta-dsp:
> Since this would also affect `FLSpec` refactor, would the respective PR (#1363 ) cover that?

@theakshaypant : This PR does not affect refactoring in #1363. Could you please elaborate your concern in more detail?

teoparvanov:
> > Since this would also affect `FLSpec` refactor, would the respective PR (#1363 ) cover that?
> 
> @theakshaypant : This PR does not affect refactoring in #1363. Could you please elaborate your concern in more detail?

AFAIU, @theakshaypant refers to the fact that the `FLSpec` will no longer have a `_runtime` member variable as part of the https://github.com/securefederatedai/openfl/pull/1363/ refactoring.

Which additionally raises the question why @ishant162 doesn't modify the `_find_runtime_instance()` method accordingly?

openfl/experimental/workflow/workspace_export/export.py
```diff
@@ -462,15 +462,16 @@ def _find_flow_class_name_if_needed(self):
...
-            if isinstance(t, federated_flow_class):
-                if not hasattr(t, "_runtime"):
+        for module_name in self.available_modules_in_exported_script:
+            module_instance = getattr(self.exported_script_module, module_name)
+            if isinstance(module_instance, federated_flow_class):
+                flow_name = module_name
+                if not hasattr(module_instance, "_runtime"):
                     raise AttributeError("Unable to locate LocalRuntime instantiation")
-                runtime = t._runtime
+                runtime = module_instance._runtime
```
theakshaypant, line 471: @teoparvanov Precisely! Per my understanding, this line would not work after the refactor since the Flow does not have a runtime associated with it.

ishant162:
> > > Since this would also affect `FLSpec` refactor, would the respective PR (#1363 ) cover that?
> > 
> > 
> > @theakshaypant : This PR does not affect refactoring in #1363. Could you please elaborate your concern in more detail?
> 
> AFAIU, @theakshaypant refers to the fact that the `FLSpec` will no longer have a `_runtime` member variable as part of the #1363 refactoring.
> 
> Which additionally raises the question why @ishant162 doesn't modify the `_find_runtime_instance()` method accordingly?

@teoparvanov, @theakshaypant,
`The _find_runtime_instance()` method supports the `Aggregator-based workflow`, where `_runtime` is still in use. Refactoring in #1363 focused on `FederatedRuntime` and `LocalRuntime`, with Aggregator workflow changes planned for a follow-up PR
 
Since the changes in this PR will be overridden by the follow-up PR maybe the issue resolution could wait for the follow-up PR.

teoparvanov:
@ishant162, IMHO for better consistency, we should wait for your follow-up PR instead of merging a custom fix.

teoparvanov (changes requested):
Requesting changes as per my previous comment, to put this PR on hold for now.

refai06:
While changes in this PR fix the referenced issue, due to refactoring of `FLSpec` and `Runtime` these changes shall need to be adjusted again. As per discussion this PR is not considered necessary and therefore closing the same

PR #1422, opened 2025-03-05
Cleanup

MasterSkepticista:
Summary of changes:

* Dead code elimination (remnant of interactive API like `TASK_REGISTRY`)
* `delta_updates` -> `use_delta_updates`, to make it consistent with aggregator that went through a similar change.
* All loggers module level (!). No more passing around `self.logger`.
* Headers are created in a stateless manner for grpc client and server. See https://github.com/securefederatedai/openfl/pull/1422/commits/2372d1e286a51a9ae24ea854cfe5adfa3343a8cc
* All gRPC checks are now inline with actionable error messages https://github.com/securefederatedai/openfl/pull/1422/commits/6f135b71691844c206d198407ce8589f97aa0f45
* Certain one-liner functions have been unfolded into main source code for readability https://github.com/securefederatedai/openfl/pull/1422/commits/573201e1d7c2f1fceb6fc0b76a42f0d70a38a160

Prerequisite for a larger refactor. Did not want to burden refactor PRs with cleanups like these.


rahulga1 (approved):

rahulga1:
@MasterSkepticista can you please keep the PR in WIP/Draft state when it is being updated, that will avoid confusion for ppl to review it pre-maturely.

MasterSkepticista:
@rahulga1 Sure. Marked as WIP. I have kept it in this state for CI to trigger, hope this is ok.

MasterSkepticista:
Ready for review.

P.S.: Thanks @noopurintel for digging this https://github.com/securefederatedai/openfl/pull/1422/commits/50ebdefd40ec6efb3ddd00997642351e48baf924

openfl/transport/grpc/aggregator_client.py
```diff
@@ -154,21 +153,64 @@ def wrapper(self, *args, **kwargs):
...
-                    self.logger.info("Reconnecting to aggregator")
+                    logger.info("Reconnecting to aggregator")
                     self.reconnect()
                 self.sleeping_policy.sleep()
         return response
 
     return wrapper
 
 
+def create_insecure_channel(uri):
```
rahulga1: should we move this to utils.py file for modularity purpose?
Also please put function doc string.

openfl/transport/grpc/aggregator_client.py
```diff
@@ -237,104 +279,30 @@ def __init__(
         self.certificate = certificate
         self.private_key = private_key
         self.sleeping_policy = ConstantBackoff(
-            int(kwargs.get("client_reconnect_interval", 1)),
-            getLogger(__name__),
-            self.uri,
+            reconnect_interval=kwargs.get("client_reconnect_interval", 1),
```
rahulga1, line 226: not sure why earlier the int type conversion was used? if its really not needed, never mind.

openfl/transport/grpc/common.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2020-2024 Intel Corporation
```
rahulga1, line 1: I think this file could be right place to put those create_secure/insecure channel method and may be renamed as utils.py

openfl/transport/grpc/aggregator_client.py
MasterSkepticista: It is an inline function, I'm not sure if a docstring will add value in this case. What do you recommend?

openfl/transport/grpc/aggregator_client.py
MasterSkepticista: It will fail if a non-integer is passed anyway, will let this be unless we have to revisit...

openfl/transport/grpc/common.py
MasterSkepticista: We could, but channel functions are only used by client

rahulga1: My point is from long term perspective, where this file can be used as a placeholder for any such functions which has possibility to be reused in other classes. Just a thought.

openfl/transport/grpc/aggregator_client.py
rahulga1: I would recommend doing it for consistency purpose.

psfoley: +1 to moving these functions (`create_insecure_channel`, `create_tls_channel`) to utils.py

psfoley (approved):
Overall changes look good here, @MasterSkepticista! `use_delta_updates` parameter name will also need to be updated in OpenFL-Security once this is merged.

rahulga1 (approved):

PR #1429, opened 2025-03-06
Support for the exclusion/inclusion of specific data types to be passed through the network.

yuliasherman:
## Summary
The Workflow API allows object code to be sent today - in part because we want the Workflow API to be more flexible and give model owners / workflow definers the ability to do things outside of horizontal federated learning. But there should be a way to configure a Workflow (at the Runtime level) that ensures that unwanted data types could not be sent. The benefit of this is that when we eventually implement the `SecureFederatedRuntime`, we can attest OpenFL code to ensure that malicious code can not be sent during execution.

## Type of Change (Mandatory)
Security improvement

## Description (Mandatory)
Allow a system to be configured in a way that prohibits/allows passing arguments of chosen types through the FLSpec 'next' function. The new parameters are defined as a Runtime class attributes to be able injecting them into the aggregator object from which the runtime object is not available.

## Testing
Manual

## Additional Information
The mechanism to terminate a failed experiment in a clean way while using FederatedRuntime yet to be developed.

### DIFF
--- a/openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
+++ b/openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
@@ -609,7 +609,9 @@
     "    collaborators=collaborator_names,\n",
     "    director=director_info, \n",
     "    notebook_path='./101_MNIST_FederatedRuntime.ipynb'\n",
-    ")"
+    ")\n",
+    "# Set allowed data types\n",
+    "FederatedRuntime.allowed_data_types = [\"str\", \"int\", \"float\", \"bool\"]"
    ]
   },
   {
--- a/openfl/experimental/workflow/component/aggregator/aggregator.py
+++ b/openfl/experimental/workflow/component/aggregator/aggregator.py
@@ -198,6 +198,10 @@ class Aggregator:
         Returns:
             flow (FLSpec): Updated instance.
         """
+        # As an aggregator is created before a runtime, set prohibited/allowed data types for the
+        # runtime before running the flow.
+        self.flow.runtime.prohibited_data_types = FederatedRuntime.prohibited_data_types
+        self.flow.runtime.allowed_data_types = FederatedRuntime.allowed_data_types
         # Start function will be the first step if any flow
         f_name = "start"
         # Creating a clones from the flow object
@@ -394,7 +398,7 @@ class Aggregator:
                     # Create list of selected collaborator clones
                     selected_clones = ([],)
                     for name, clone in self.clones_dict.items():
-                        # Check if collaboraotr is in the list of selected
+                        # Check if collaborator is in the list of selected
                         # collaborators
                         if name in self.selected_collaborators:
                             selected_clones[0].append(clone)
--- a/openfl/experimental/workflow/interface/fl_spec.py
+++ b/openfl/experimental/workflow/interface/fl_spec.py
@@ -22,6 +22,7 @@ from openfl.experimental.workflow.utilities import (
     filter_attributes,
     generate_artifacts,
     should_transfer,
+    validate_data_types,
 )
 
 
@@ -127,16 +128,16 @@ class FLSpec:
     def run(self) -> None:
         """Starts the execution of the flow."""
         # Submit flow to Runtime
-        if str(self._runtime) == "LocalRuntime":
+        if str(self.runtime) == "LocalRuntime":
             self._run_local()
-        elif str(self._runtime) == "FederatedRuntime":
+        elif str(self.runtime) == "FederatedRuntime":
             self._run_federated()
         else:
             raise Exception("Runtime not implemented")
 
     def _run_local(self) -> None:
         """Executes the flow using LocalRuntime."""
-        self._setup_initial_state()
+        self._setup_initial_state_local()
         try:
             # Execute all Participant (Aggregator & Collaborator) tasks and
             # retrieve the final attributes
@@ -164,7 +165,7 @@ class FLSpec:
         for name, attr in final_attributes:
             setattr(self, name, attr)
 
-    def _setup_initial_state(self) -> None:
+    def _setup_initial_state_local(self) -> None:
         """
         Sets up the flow's initial state, initializing private attributes for
         collaborators and aggregators.
@@ -176,6 +177,7 @@ class FLSpec:
         self._foreach_methods = []
         FLSpec._reset_clones()
         FLSpec._create_clones(self, self.runtime.collaborators)
+
         # Initialize collaborator private attributes
         self.runtime.initialize_collaborators()
         if self._checkpoint:
@@ -334,7 +336,7 @@ class FLSpec:
         parent = inspect.stack()[1][3]
         parent_func = getattr(self, parent)
 
-        if str(self._runtime) == "LocalRuntime":
+        if str(self.runtime) == "LocalRuntime":
             # Checkpoint current attributes (if checkpoint==True)
             checkpoint(self, parent_func)
 
@@ -343,10 +345,15 @@ class FLSpec:
         if aggregator_to_collaborator(f, parent_func):
             agg_to_collab_ss = self._capture_instance_snapshot(kwargs=kwargs)
 
-        # Remove included / excluded attributes from next task
-        filter_attributes(self, f, **kwargs)
+        if kwargs:
+            # Remove unwanted attributes from next task
+            filter_attributes(self, f, **kwargs)
+            if self.runtime.prohibited_data_types or self.runtime.allowed_data_types:
+                validate_data_types(
+                    self.runtime.prohibited_data_types, self.runtime.allowed_data_types, **kwargs
+                )
 
-        if str(self._runtime) == "FederatedRuntime":
+        if str(self.runtime) == "FederatedRuntime":
             if f.collaborator_step and not f.aggregator_step:
                 self._foreach_methods.append(f.__name__)
 
@@ -359,6 +366,6 @@ class FLSpec:
                 kwargs,
             )
 
-        elif str(self._runtime) == "LocalRuntime":
+        elif str(self.runtime) == "LocalRuntime":
             # update parameters required to execute execute_task function
             self.execute_task_args = [f, parent_func, agg_to_collab_ss, kwargs]
--- a/openfl/experimental/workflow/runtime/runtime.py
+++ b/openfl/experimental/workflow/runtime/runtime.py
@@ -10,7 +10,77 @@ from openfl.experimental.workflow.interface.fl_spec import FLSpec
 from openfl.experimental.workflow.interface.participants import Aggregator, Collaborator
 
 
-class Runtime:
+class AttributeValidationMeta(type):
+    """
+    Metaclass that enforces validation rules on class attributes.
+
+    This metaclass ensures that `prohibited_data_types` and `allowed_data_types`
+    are lists of strings and that both cannot be set at the same time.
+
+    Example:
+        class MyClass(metaclass=AttributeValidationMeta):
+            pass
+
+        MyClass.prohibited_data_types = ["int", "float"]  # Valid
+        MyClass.allowed_data_types = ["str", "bool"]  # Valid
+        MyClass.prohibited_data_types = "int"  # Raises TypeError
+        MyClass.allowed_data_types = 42  # Raises TypeError
+    """
+
+    def __setattr__(cls, name, value):
+        """
+        Validates and sets class attributes.
+
+        Ensures that `prohibited_data_types` and `allowed_data_types`, when assigned,
+        are lists of strings and that they are not used together.
+
+        Args:
+            name (str): The attribute name being set.
+            value (any): The value to be assigned to the attribute.
+
+        Raises:
+            TypeError: If `prohibited_data_types` or `allowed_data_types` is not a list
+                or contains non-string elements.
+            ValueError: If both `prohibited_data_types` and `allowed_data_types` are set.
+        """
+        if name in {"prohibited_data_types", "allowed_data_types"}:
+            if not isinstance(value, list):
+                raise TypeError(f"'{name}' must be a list, got {type(value).__name__}")
+            if not all(isinstance(item, str) for item in value):
+                raise TypeError(f"All elements of '{name}' must be strings")
+
+            # Ensure both attributes are not set at the same time
+            other_name = (
+                "allowed_data_types" if name == "prohibited_data_types" else "prohibited_data_types"
+            )
+            if getattr(cls, other_name, []):  # Check if the other attribute is already set
+                raise ValueError(
+                    "Cannot set both 'prohibited_data_types' and 'allowed_data_types'."
+                )
+
+        super().__setattr__(name, value)
+
+
+class Runtime(metaclass=AttributeValidationMeta):
+    """
+    Base class for federated learning runtimes.
+    This class serves as an interface for runtimes that execute FLSpec flows.
+
+    Attributes:
+        prohibited_data_types (list): A list of data types that are prohibited from being
+            transmitted over the network.
+        allowed_data_types (list): A list of data types that are explicitly allowed to be
+            transmitted over the network.
+
+    Notes:
+        - Either `prohibited_data_types` or `allowed_data_types` may be specified.
+        - If neither is specified, all data types are allowed to be transmitted.
+        - If both are specified, a `ValueError` will be raised.
+    """
+
+    prohibited_data_types = []
+    allowed_data_types = []
+
     def __init__(self):
         """Initializes the Runtime object.
 
--- a/openfl/experimental/workflow/utilities/runtime_utils.py
+++ b/openfl/experimental/workflow/utilities/runtime_utils.py
@@ -7,6 +7,7 @@
 import inspect
 import itertools
 from types import MethodType
+from typing import List
 
 import numpy as np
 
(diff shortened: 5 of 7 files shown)


openfl/experimental/workflow/utilities/runtime_utils.py
```diff
@@ -96,6 +97,31 @@ def filter_attributes(ctx, f, **kwargs):
...
+def validate_data_types(
+    prohibited_data_types: List[str], reserved_words=["collaborators"], **kwargs
+):
+    """Validates that the types of attributes in kwargs are not among the prohibited data types.
+    Raises a TypeError if any prohibited data type is found.
+
+    Args:
+        prohibited_data_types (List[str]): A list of prohibited data type names
+            (e.g., ['int', 'float']).
+        kwargs (dict): Arbitrary keyword arguments representing attribute names and their values.
```
teoparvanov: Could you add some documentation regarding the `reserved_words` parameter?

openfl/experimental/workflow/component/aggregator/aggregator.py
```diff
@@ -198,6 +198,9 @@ async def run_flow(self) -> FLSpec:
         Returns:
             flow (FLSpec): Updated instance.
         """
+        # As an aggregator is created before a runtime, set prohibited data types for the runtime
+        # before running the flow.
+        self.flow.runtime.prohibited_data_types = FederatedRuntime.prohibited_data_types
```
teoparvanov, line 203: Where is `FederatedRuntime.prohibited_data_types` defined?

teoparvanov (commented):
A couple of questions and comments from my first read-through:

openfl/experimental/workflow/utilities/runtime_utils.py
yuliasherman: Done

openfl/experimental/workflow/component/aggregator/aggregator.py
yuliasherman: In the Runtime (base class)

rahulga1:
Please rebase you PR.

openfl/experimental/workflow/utilities/runtime_utils.py
```diff
@@ -96,6 +97,33 @@ def filter_attributes(ctx, f, **kwargs):
         _process_exclusion(ctx, cls_attrs, kwargs["exclude"], f)
 
 
+# This function validates the data types of the attributes in kwargs. 'reserved_words' parameter
+# can be passed as a list of strings that should be accepted as 'str' attributes.
```
theakshaypant: Nit: As per python docstring conventions, we should keep these lines (100-101) in the multi-line comment below. Doing so will also add the above info to the [openfl API reference](https://openfl.readthedocs.io/en/latest/openfl.html).

References: [1](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings), [2](https://peps.python.org/pep-0257/).

yuliasherman: Done

openfl/experimental/workflow/interface/fl_spec.py
```diff
@@ -343,8 +345,11 @@ def next(self, f, **kwargs) -> None:
         if aggregator_to_collaborator(f, parent_func):
             agg_to_collab_ss = self._capture_instance_snapshot(kwargs=kwargs)
 
-        # Remove included / excluded attributes from next task
-        filter_attributes(self, f, **kwargs)
+        # Remove unwanted attributes from next task
+        if kwargs:
+            filter_attributes(self, f, **kwargs)
+            if self._runtime.prohibited_data_types:
```
ishant162: I suggest using the runtime getter instead of directly accessing `_runtime`.

openfl/experimental/workflow/runtime/runtime.py
```diff
@@ -10,7 +10,19 @@
 from openfl.experimental.workflow.interface.participants import Aggregator, Collaborator
 
 
-class Runtime:
+class AttributeValidationMeta(type):
```
ishant162, line 13: Adding docstrings to `AttributeValidationMeta` and `Runtime` classes would be helpful to understand the code at a glance.

openfl/experimental/workflow/interface/fl_spec.py
```diff
@@ -343,8 +345,11 @@ def next(self, f, **kwargs) -> None:
...
+                validate_data_types(self._runtime.prohibited_data_types, **kwargs)
```
ishant162: It might be cleaner to move this functionality into `filter_attributes` (i.e., call it from within `filter_attributes` itself). This would also remove the need to import it in `flspec`. WDYT?

openfl/experimental/workflow/runtime/runtime.py
yuliasherman: Done

openfl/experimental/workflow/interface/fl_spec.py
yuliasherman: OK, fixed also other functions that use it directly.

openfl/experimental/workflow/interface/fl_spec.py
yuliasherman: I initially thought about it, but those two functions have different functionality, so I prefer not to bind them together.

openfl/experimental/workflow/component/aggregator/aggregator.py
psfoley, line 203: It would be helpful to follow the Metaflow pattern in the next function for `include`/`exclude`; we need to have `allowed_data_types` as well. I expect it will be far easier to specify what's allowed vs. trying to capture all the possible types that are not, i.e. `allowed_data_types=[np.ndarray]`

eyalrund (approved):

openfl/experimental/workflow/component/aggregator/aggregator.py
yuliasherman: AFAIS currently we pass 'str' type as the 'next' function argument in notebooks:
  self.next( self.aggregated_model_validation, foreach="collaborators")
Following your suggestion: "allowed_data_types=[np.ndarray]", we are also prohibiting numeric types, even though there is no reason for doing so.
I will add 'allowed_data_types' alongside 'prohibited_data_types', allowing the user to choose the option that best suits their needs.

ishant162 (commented):
For the FederatedRuntime, how does the Aggregator become aware of the allowed or prohibited data types configured in the Jupyter Notebook, especially in a distributed infrastructure setup?
Also, have these scenarios been tested on a distributed infrastructure?

openfl/experimental/workflow/interface/fl_spec.py
```diff
@@ -343,10 +345,15 @@ def next(self, f, **kwargs) -> None:
         if aggregator_to_collaborator(f, parent_func):
             agg_to_collab_ss = self._capture_instance_snapshot(kwargs=kwargs)
 
-        # Remove included / excluded attributes from next task
-        filter_attributes(self, f, **kwargs)
+        if kwargs:
```
scngupta-dsp, line 348: Is it possible for the flow to inject / define a prohibited data type in any step? In this case I think that `prohibited_data_types` or `allowed_data_types` should be handled even if no kwargs (`include` or `exclude`) are specified

openfl/experimental/workflow/runtime/runtime.py
```diff
@@ -10,7 +10,79 @@
...
+    Example:
+        class MyClass(metaclass=AttributeValidationMeta):
+            pass
+
+        MyClass.prohibited_data_types = ["int", "float"]  # Valid
+        MyClass.allowed_data_types = ["str", "bool"]  # Valid
+        MyClass.prohibited_data_types = "int"  # Raises TypeError
+        MyClass.allowed_data_types = 42  # Raises TypeError
+        MyClass.prohibited_data_types = ["int"]
+        MyClass.allowed_data_types = ["str"]  # Raises ValueError
```
scngupta-dsp: Can you clarify why this usage `LocalRuntime.allowed_data_types = ["str"]` shall raise a `ValueError` ?

openfl/experimental/workflow/runtime/runtime.py
```diff
@@ -10,7 +10,79 @@
...
+            ValueError: If both `prohibited_data_types` and `allowed_data_types` are set.
+        """
+        if name in {"prohibited_data_types", "allowed_data_types"}:
+            if not isinstance(value, list):
+                raise TypeError(f"'{name}' must be a list, got {type(value).__name__}")
+            if not all(isinstance(item, str) for item in value):
+                raise TypeError(f"All elements of '{name}' must be strings")
+
+            # Ensure both attributes are not set at the same time
+            other_name = (
```
scngupta-dsp, line 53: Both the attributes can be defined in following manner, but this will get enforced by `validate_data_types` - right ? 
```
LocalRuntime.allowed_data_types = []
LocalRuntime.prohibited_data_types = []
LocalRuntime.prohibited_data_types.append("str")
LocalRuntime.allowed_data_types.append("int")
```

yuliasherman:
> For the FederatedRuntime, how does the Aggregator become aware of the allowed or prohibited data types configured in the Jupyter Notebook, especially in a distributed infrastructure setup? Also, have these scenarios been tested on a distributed infrastructure?

prohibited_data_types and allowed_data_types are class attributes of the Runtime class. These attributes are configured in the aggregator code and can be reviewed in the PR. The scenario has been tested with both Local and Federated runtimes. The FederatedRuntime test is expected to behave equivalently to the distributed setup.

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
```diff
@@ -490,6 +490,8 @@
     "local_runtime = LocalRuntime(\n",
     "    aggregator=aggregator, collaborators=collaborators, backend=\"single_process\"\n",
     ")\n",
+    "# Set prohibited data types\n",
+    "LocalRuntime.prohibited_data_types = [\"bytes\", \"bytearray\"]\n",
```
scngupta-dsp: I am wondering if this feature is really useful for `LocalRuntime` where no data is transmitted over the network. Is it being offered to enable the user to simulate this feature and ensure that the flow works according to the policy

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
```diff
@@ -609,7 +611,9 @@
     "    collaborators=collaborator_names,\n",
     "    director=director_info, \n",
     "    notebook_path='./101_MNIST_FederatedRuntime.ipynb'\n",
-    ")"
+    ")\n",
+    "# Set allowed data types\n",
+    "FederatedRuntime.allowed_data_types = [\"str\", \"int\", \"float\", \"bool\"]"
```
scngupta-dsp, line 614: `101_MNIST_FederatedRuntime` may not be the right place to introduce this feature. These attributes are added without any documentation (or hints about their purpose / intended behavior) making it difficult to follow for the users. Even if some documentation is added, I feel that introducing many features in a single tutorial could make the tutorial complex

Due to this I would recommend adding a new tutorial with a valid / real world use case and then demonstrate how the infrastructure prohibits a data type (malicious function?) to be sent across the network

openfl/experimental/workflow/interface/fl_spec.py
```diff
@@ -343,10 +345,15 @@ def next(self, f, **kwargs) -> None:
...
+            # Remove unwanted attributes from next task
+            filter_attributes(self, f, **kwargs)
+            if self.runtime.prohibited_data_types or self.runtime.allowed_data_types:
```
scngupta-dsp, line 351: As the intent is to prohibit / allow certain data types to be transmitted over the network, this validation of whether certain attributes are prohibited or allowed could be done at a transition point (aggregator to collaborator or vice versa). WDYT?

openfl/experimental/workflow/interface/fl_spec.py
```diff
@@ -343,10 +345,15 @@ def next(self, f, **kwargs) -> None:
...
+                validate_data_types(
```
scngupta-dsp, line 352: I am not able to follow why `kwargs` are being validated to check if they are prohibited or allowed. Some examples

`self.next(self.aggregated_model_validation, foreach="collaborators", exclude="private")`
> kwargs = {'foreach': 'collaborators', 'exclude': 'private`}; validation occurs on 
> > attr_name = `foreach` and attr_type = `str`
> > attr_name = `exclude` and attr_type = `str`

`self.next(self.join)`
> kwargs = {}; no validation occurs

I do not think this validation will affect what is being sent over the network. Can you please clarify

openfl/experimental/workflow/runtime/runtime.py
```diff
@@ -10,7 +10,79 @@
 from openfl.experimental.workflow.interface.participants import Aggregator, Collaborator
 
 
-class Runtime:
+class AttributeValidationMeta(type):
+    """
+    Metaclass that enforces validation rules on class attributes.
```
scngupta-dsp, line 15: I am wondering how the policy of prohibiting certain data types to be transmitted over the network could be enforced. In the current design there is a potential to modify them via flow. For e.g. 

```
self.collaborators = self.runtime.collaborators
print(f"Current prohibited_data_types = {self.runtime.prohibited_data_types}")
self.runtime.prohibited_data_types = []
print(f"Modified prohibited_data_types = {self.runtime.prohibited_data_types}")
```

Is this a valid scenario that should be prevented?

scngupta-dsp (changes requested):
In addition to individual comments, a valid (close to real world use case) use case / Tutorial for this feature should be demonstrated. This would significantly enhance the understanding of this feature

openfl/experimental/workflow/runtime/runtime.py
yuliasherman: It will not, my bad. Thank you.

openfl/experimental/workflow/runtime/runtime.py
yuliasherman: Correct

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
yuliasherman: Correct, it is introduced only for the testing purpose.

openfl/experimental/workflow/interface/fl_spec.py
yuliasherman: What do you mean by transition point? protobuf? We want to limit data types injected remotely. At a protobuf level we don't have an information about where the data came from.

openfl/experimental/workflow/interface/fl_spec.py
yuliasherman: We rely on a notebook review conducted by the data owner to prevent malicious functions from being written. However, this review may not be effective when function arguments are provided as raw data and potentially contain executable code. To mitigate this risk, we allow functions written in the notebook to be used but enforce validation on their arguments.

openfl/experimental/workflow/runtime/runtime.py
yuliasherman: Here again we rely on a notebook review.

openfl/experimental/workflow/interface/fl_spec.py
yuliasherman: This configuration is intended to validate only kwargs passed to the "next" function.

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
teoparvanov: Thanks for pointing this out, @scngupta-dsp! While I agree that a dedicated tutorial for this feature could be added separately (esp. for advanced scenarios like Vertical FL), the basic usage is straightforward enough to be showcased via the 101 tutorial. In fact, we would likely want this to be explicitly stated in any notebook that would require a security review by collaborators/data owners.

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
teoparvanov: @yuliasherman, I would suggest removing this from the 101 tutorial, and only adding it to the appropriate notebook(s) used by Workflow interface unit tests.

openfl/experimental/workflow/runtime/runtime.py
```diff
@@ -10,7 +10,77 @@
...
+            transmitted over the network.
+
+    Notes:
+        - Either `prohibited_data_types` or `allowed_data_types` may be specified.
+        - If neither is specified, all data types are allowed to be transmitted.
+        - If both are specified, a `ValueError` will be raised.
+    """
+
+    prohibited_data_types = []
+    allowed_data_types = []
```
teoparvanov, line 82: As I mentioned in our offline discussion, it bothers me a little bit that those are class attributes, instead of well-encapsulated object properties. This could lead to strange behaviors, f.e. if a single notebook first uses a `LocalRuntime` object for simulation, and then a `FederatedRuntime` for a distributed run (both would share the same class-level prohibited/allow lists).

We may decide as a team to accept the risk and proceed with this approach for now. Later on, we could consider refactoring to a cleaner, object-oriented approach where each `Runtime` object has its own allow/prohibit lists (f.e. wired via the FL plan that's generated under the hood).

WDYT?

teoparvanov (commented):
Thanks @yuliasherman, I think this approach would functionally work - although I'd prefer an object-oriented implementation of the allow/prohibit lists (if at all possible). 

Here are some additional thoughts and comments:

openfl/experimental/workflow/runtime/runtime.py
kminhta: I generally do agree with you on this, @teoparvanov . Given that `LocalRuntime` and `FederatedRuntime` are initialized separately, using encapsulated object properties could prevent strange behaviors, especially with debugging if users expect different deny/allow lists for each runtime.
 
However, I actually don't anticipate that these lists would differ within the same experiment, and it seems almost more intuitive for the general Runtime class to apply these lists universally. Considering this, the risk seems low enough to proceed with the current approach.

That said, if refactoring to object properties or instance attributes isn't too much of a lift, it might be worth doing now to avoid technical debt.

scngupta-dsp: @teoparvanov: I agree with you that it could lead to strange behaviors and should be reevaluated. Using class attributes to define the properties and then enforcing them by [instance attributes](https://github.com/securefederatedai/openfl/blob/0b920744cebaaed456742e7c53c62f172d29d57c/openfl/experimental/workflow/interface/fl_spec.py#L351) could lead to another potential risk that is described [here](https://github.com/securefederatedai/openfl/pull/1429/files#r2011407171)

openfl/experimental/workflow/runtime/runtime.py
teoparvanov: I agree with @yuliasherman that the ultimate guarantee against malicious behaviors is the code review (manual or AI-assisted in the future).

openfl/experimental/workflow/runtime/runtime.py
teoparvanov: Hey @yuliasherman, I suggest we meet with @scngupta-dsp next week to discuss whether a fully instance-based implementation is even feasible. If not (or if too complex), we can proceed with the current approach.

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
yuliasherman: I agree that this does not worth a new notebook creation.

openfl-tutorials/experimental/workflow/FederatedRuntime/101_MNIST/workspace/101_MNIST_FederatedRuntime.ipynb
yuliasherman: OK

teoparvanov (approved):
Approving this PR, as this seems like the best currently known approach to including/excluding "next" function parameter types. This is likely the solution that will go into 1.9, but I suggest not merging it until the alternative approach has been fully considered.

PS: In parallel, we are exploring a solution where the allow/prohibit lists become Runtime instance variables, and are passed around via the `plan.yaml`. This method is conceptually cleaner, but it may be hindered by the limitations of the runtime object itself (esp. from the collaborator side).

PR #1433, opened 2025-03-07
 Introducing Task Runner workspace to enable interoperability with Flower workloads

kminhta:
## Summary
This PR introduces a Task Runner example workspace with some framework-level modification that enables interoperability with Flower experiment. In particular, a user can set up an experiment using Flower API and have it run through OpenFL's Task Runner.

## Type of Change (Mandatory)
Feature enhancement

## Description (Mandatory)
A major step toward FL standards is supporting functional interoperability between existing FL frameworks. This PR introduces a workspace to support interoperability with Flower. The workspace extends all the components needed (i.e. the aggregator, collaborator, task runner, etc.) in order to enable this interoperability as well as introduces a new `FlowerTaskRunner`, `FlowerDataLoader`, and `Connector` class. This is intentional in order to minimize framework-level changes at this stage while still offering an avenue to interoperate between the two. It is intended that in the future, we migrate functionality of the extended components to the framework-level counterparts for fully integrated support.

That said, there are some framework-level changes that need to be made that could not be encapsulated in the workspace.

**Framework-level changes**
1.  We introduce a new component called the `Connector` which is tasked with handling the server-side communication (i.e. the local gRPC client and the Flower SuperLink, ServerApp, and run commands). For this, we add a `get_connector` method to `openfl.federated.plan` in order to build the `Connector` object with configurations specified in `plan.yaml`
2.  We modify `openfl.federated.runner` in order to be able to accept `dataloader=None`. This is particularly important as part of initializing the `FlowerTaskRunner` is to install the Flower app bundle (for SGX) and does not use a dataloader object
3. We add a conditional to `interface.plan` in order to bypass initialization of tensors if the `connector` is present since the model is handled by Flower and not OpenFL (Flower has no concept of `init.pbuf` to start the experiment)
4. Introduce a new RPC `aggregator.proto` called `PelicanDrop` (open to name suggestions) that'll input and return a `DropPod` which is effectively just a gRPC message that carries the Flower message plus header information (for verification) and any metadata that OpenFL decides to add (i.e. an `end_experiment` flag for automatic shutdown)
5. Add a `send_message_to_server` method to the OpenFL `aggregator_client` that is used to communicate with the local gRPC server (which accepts requests from Flower SuperNode) and `aggregator_server`
6. Add a `PelicanDrop` method to OpenFL `aggregator_server` in order to accept Flower messages from the OpenFL client (converted to DropPod) and pass it to the local gRPC client (which will communicate with the Flower SuperLink)
   - also add self attribute to acquire local grpc client from the aggregator/connector
   - logic to start and stop server 
   - condition to unimplement `GetAggregatedTensor` when the connector is present since OpenFL is not responsible for aggregation in this current implementation. 

**Workspace changes (encapsulated in workspace)**
1. `FlowerTaskRunner` which extends the Task Runner base class. It has one primary task `start_client_adapter` which starts the local grpc server and Flower SuperNode
2. `FlowerDataLoader` which obtains the data path (as set in `data.yaml`) to be inputted into the `flower-supernode` command.
3.  New Connector component - briefly discussed above, it is a new component designed to manage connections between the OpenFL aggregator server and Flower (or any other) server. It is built into the aggregator.
4.  We extend the aggregator with methods to interact with the connector as well as conditionals to bypass initialization, aggregation, computing scores, etc. which are handled by flower
5.  We extend the collaborator's `do_task` command to enable it to set up the local grpc server, have the task runner run the `start_client_adapter` task, and call `send_local_task_results` directly after rather than acquiring tensor keys and metrics, which are handled by flower at this stage
6. under `grpc.connector`, we introduce logic to deserialize the flower messages, convert between flower and openfl messages, as well as set up and handle the local gRPC components.

**NOTE ON PATCHING**
We also include a patch on `flwr.cli.build` which:
1. saves the FAB file to `FLWR_HOME` directory
2. sorts the list of files in the FAB before hashing for reproducibility

Both of these patches (as well as building and installing the Flower FAB during task runner initialization) are necessary for running SGX but NOT necessarily required for running outside of SGX. The reason is that there are system calls that `flwr build` and `flwr install` call that are not supported in Gramine. Patch 2 has already been merged into Flower and will likely be able to be removed in the next stable release, Patch 1 being added to the Flower-level is still under discussion

## Testing
Below is a successful run:
image

and in SGX enclave:
image


openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
+# Open(FL)ower
```
MasterSkepticista, line 1: Nice

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+This workspace demonstrates a new functionality in OpenFL to interoperate with [Flower](https://flower.ai/). In particular, a user can now use the Flower API to run on OpenFL infrastructure. OpenFL will act as an intermediary step between the Flower SuperLink and Flower SuperNode to relay messages across the network using OpenFL's transport mechanisms.
```
MasterSkepticista, line 3: I suggest describing the main benefit to both side of the communities in this overview: they get a peek into what they hope to achieve by the end of this guide.

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+pip install virtualenv
+virtualenv ./venv
+source ./venv/bin/activate
+```
+
+Install OpenFL from source
+```sh
+git clone https://github.com/securefederatedai/openfl.git
+cd openfl
+pip install -e .
```
MasterSkepticista: Can we point to our official installation docs?

openfl/federated/plan/plan.py
```diff
@@ -382,6 +393,10 @@ def get_aggregator(self, tensor_dict=None):
         defaults[SETTINGS]["compression_pipeline"] = self.get_tensor_pipe()
         defaults[SETTINGS]["straggler_handling_policy"] = self.get_straggler_handling_policy()
 
+        connector = self.get_connector()
+        if connector is not None:
+            defaults[SETTINGS]["connector"] = connector
```
MasterSkepticista, line 398: Given connector is not used anywhere else, I suggest in-lining the contents of `get_connector()` here. This will contain the surface of changes, and reduce one state variable `self.connector_()`.

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Execution Methods
+
+There are two ways to execute this:
+
+1. Automatic shutdown which will spawn a `server-app` in isolation and trigger an experiment termination once the it shuts down. (Default/Recommended)
+2. Running `SuperLink` and `SuperNode` as [long-lived components](#long-lived-superlink-and-supernode) that will indefinitely wait for new runs. (Limited Functionality)
```
MasterSkepticista: Contents of this section feel too early for my taste. My suggestion would be to move this to post installation/setup

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+```yaml
+task_runner:
+  defaults: plan/defaults/task_runner.yaml
+  template: openfl.federated.task.runner_flower.FlowerTaskRunner
+  settings:
+    auto_shutdown: True
+```
+3. `FlowerDataLoader` with similar high-level functionality to other dataloaders.
+
+**IMPORTANT NOTE**: `aggregator.settings.rounds_to_train` is set to 1. __Do not edit this__. The actual number of rounds for the experiment is controlled by Flower logic inside of `./app-pytorch/pyproject.toml`. The entirety of the Flower experiment will run in a single OpenFL round. The aggregator round is there to stop the OpenFL components at the completion of the experiment.
```
MasterSkepticista: Instead of `__Do not edit this__`, we should mention `This parameter will be ignored by OpenFL. As with Flower applications, rounds to train is specified within pyproject.toml` or the likes.

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+# The CA signs collaborator2's certificate
+fx collaborator certify -n collaborator2 --silent
+
+##############################
+# Start to Run the Federation
+##############################
+
+# Run the Aggregator
+fx aggregator start
```
MasterSkepticista, line 124: How about we point to our [quickstart](https://openfl.readthedocs.io/en/latest/tutorials/taskrunner.html) guide?
The quickstart covers context behind certain commands (when and where to run them), is CI-verified, and reduces duplication. If this requires contextual updates to the quickstart guide, we can discuss.

openfl-workspace/flower-app-pytorch/src/aggregator.py
MasterSkepticista, line 1: Ouch, this will be quite a bit of future maintenance overhead. Let's discuss

openfl-workspace/flower-app-pytorch/src/collaborator.py
```diff
@@ -0,0 +1,108 @@
...
+            opt_treatment,
+            device_assignment_policy,
+            use_delta_updates,
+            compression_pipeline,
+            db_store_rounds,
+            log_memory_usage,
+            write_logs,
+            callbacks,
+            secure_aggregation
+        )
```
MasterSkepticista: Given this we can entirely skip `__init__`. It will be inherited as-is I think.

openfl-workspace/flower-app-pytorch/src/connector.py
MasterSkepticista, line 1: This is a connector proprietary to Flower and this workspace. We can inherit from the main implementation if required. Do we really need this ABC?

openfl-workspace/flower-app-pytorch/src/grpc/connector/__init__.py
```diff
@@ -0,0 +1,3 @@
+from src.grpc.connector.utils import get_local_grpc_server
+
+__all__ = ['get_local_grpc_server']
```
MasterSkepticista: `__all__` not necessary

MasterSkepticista (commented):
(partial review)

openfl-workspace/flower-app-pytorch/README.md
teoparvanov: AFAIU, in both cases we deploy the SuperLink/SuperNode infra. Or is the intention to only do so if `auto_shutdown` is set to `False`, and just run the Flower ClientApp/ServerApp otherwise?

PS: Regarding 2., it would be interesting to explore the interop between SuperNode/SuperLink and Director/Envoy (in the context of the Workflow API `FederatedRuntime`). Of course, this is out of scope for the current PR (just an idea).

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+```sh
+fx workspace create --template flower-app-pytorch --prefix my_workspace
+cd my_workspace
+```
+
+This will create a workspace in your current working directory called `./my_workspace` as well as install the Flower app defined in `./app-pytorch.` This will be where the experiment takes place.
+
+### Configure the Experiment
+Notice under `./plan`, you will find the familiar OpenFL YAML files to configure the experiment. `col.yaml` and `data.yaml` will be populated by the collaborators that will run the Flower client app and the respective data shard or directory they will perform their training and testing on.
```
teoparvanov: nit - `cols.yaml`

openfl-workspace/flower-app-pytorch/README.md
teoparvanov: I personally think the complexity an uniqueness of the OpenFL/Flower workspace warrants self-contained documentation, even at the cost of some duplication with the official docs.

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+```
+
+3. `FlowerTaskRunner` which will execute the `start_client_adapter` task. This task starts the Flower SuperNode and makes a connection to the OpenFL client. Additionally, the `FlowerTaskRunner` has an additional setting `FlowerTaskRunner.settings.auto_shutdown` which is default set to `True`. When set to `True`, the task runner will shut the SuperNode at the completion of an experiment, otherwise, it will run continuously.
+
+```yaml
+task_runner:
+  defaults: plan/defaults/task_runner.yaml
+  template: openfl.federated.task.runner_flower.FlowerTaskRunner
+  settings:
+    auto_shutdown: True
```
teoparvanov, line 54: So, if this is set to `False`, it would be possible to trigger multiple OpenFL runs against the same SuperNode/SuperLink Flower infra?

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+```yaml
+assigner:
+  defaults: plan/defaults/assigner.yaml
+  template: openfl.component.ConnectorAssigner
+  settings:
+    task_groups:
+      - name: Connector_Flower
+        tasks:
+          - start_client_adapter
```
teoparvanov: I wonder if we need a custom assigner for this. Couldn't the `RandomGroupedAssigner` do the job (as you in fact do in the `plan.yaml` provided as part of the workspace)?

PS: I also couldn't find the `ConnectorAssigner` class definition anywhere in this PR.

openfl-workspace/flower-app-pytorch/plan/plan.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+data_loader :
+  defaults : plan/defaults/data_loader.yaml
+  template : src.loader_flower.FlowerDataLoader
+  settings :
+    collaborator_count : 2
+
+task_runner :
+  defaults : plan/defaults/task_runner.yaml
+  template : src.runner_flower.FlowerTaskRunner
```
teoparvanov: nit - I suggest simplifying the naming to `src.runner.FlowerTaskRunner`, resp. `src.runner.FlowerDataLoader`.
Or even `src.FlowerTaskRunner`/`src.FlowerDataLoader`

WDYT?

openfl-workspace/flower-app-pytorch/src/connector.py
```diff
@@ -0,0 +1,43 @@
+import signal
+import sys
+from logging import getLogger
+from abc import ABC, abstractmethod
+
+class Connector(ABC):
+    """
+    Abstract base class for managing a server process of an external federated learning framework
+    and the connection with OpenFL's server.
+    """
```
teoparvanov: By the way, we could imagine using this type of `Connector` objects to interface with containerized _ML_ frameworks for better isolation from the core OpenFL components.

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/deserialize_message.py
```diff
@@ -0,0 +1,41 @@
+import importlib
+from google.protobuf.message import DecodeError
+
+def deserialize_flower_message(flower_message):
```
teoparvanov: Where is this used? Or is it just for illustration purposes, to show that OpenFL could audit the Flower messages?

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/local_grpc_server.py
```diff
@@ -0,0 +1,141 @@
...
+import psutil
+import time
+
+logger = logging.getLogger(__name__)
+
+
+class LocalGRPCServer(grpcadapter_pb2_grpc.GrpcAdapterServicer):
+    """
+    LocalGRPCServer is a gRPC server that handles requests from the Flower SuperNode
+    and forwards them to the OpenFL Client. It uses a queue-based system to
```
teoparvanov: What is the OpenFL Client in this case?

openfl-workspace/flower-app-pytorch/src/patch/patch_flwr_build.py
```diff
@@ -0,0 +1,150 @@
...
+def build(
+    app: Annotated[
+        Optional[Path],
+        typer.Option(help="Path of the Flower App to bundle into a FAB"),
+    ] = None,
+) -> tuple[str, str]:
+    """Build a Flower App into a Flower App Bundle (FAB).
+
+    You can run ``flwr build`` without any arguments to bundle the app located in the
+    current directory. Alternatively, you can you can specify a path using the ``--app``
```
teoparvanov: typo: `you can`

openfl-workspace/flower-app-pytorch/src/runner_flower.py
```diff
@@ -0,0 +1,183 @@
...
+        local_grpc_server.start_server(local_server_port)
+
+        local_server_port = local_grpc_server.get_port()
+
+        command = [
+            "flower-supernode",
+            "--insecure",
+            "--grpc-adapter",
+            "--superlink", f"127.0.0.1:{local_server_port}",
+            "--clientappio-api-address", f"127.0.0.1:{self.client_port}",
```
teoparvanov: Shouldn't those come from the FL plan?

openfl/transport/grpc/aggregator_server.py
```diff
@@ -269,10 +288,46 @@ def SendLocalTaskResults(self, request, context):  # NOQA:N802
             federation_uuid=self.aggregator.federation_uuid,
             single_col_cert_common_name=self.aggregator.single_col_cert_common_name,
         )
+
         return aggregator_pb2.SendLocalTaskResultsResponse(header=header)
 
+    def PelicanDrop(self, request, context):
```
teoparvanov, line 294: I have to admit I don't get the pelican metaphor 😅

teoparvanov (commented):
Great work, @kta-intel ! While I'm still wrapping my head around certain elements, I don't see any major issues. A high-level components diagram could help though, esp. to better understand the role of the gRPC client/server.

Here are a couple of questions from my first review pass:

openfl-workspace/flower-app-pytorch/README.md
```diff
@@ -0,0 +1,291 @@
...
+```SH
+flwr run ./src/app-pytorch
+```
+It will run another experiment. Once you are done, you can manually shut down OpenFL's `collaborator` and Flower's `SuperNode` with `CTRL+C`. This will trigger a task-completion by the task runner that'll subsequently begin the graceful shutdown process of the OpenFL and Flower components.
+
+### Running in SGX Enclave
+Gramine does not support all Linux system calls. Flower FAB is built and installed at runtime. During this, `utime()` is called, which is an [unsupported call](https://gramine.readthedocs.io/en/latest/devel/features.html#list-of-system-calls), resulting in error or unexpected behavior. To navigate this, when running in an SGX enclave, we opt to build and install the FAB during initialization and package it alongside the OpenFL workspace. To make this work, we introduce some patches to Flower's build command. In addition, since secure enclaves have strict read/write permissions, dictate by a set of trusted/allowed files
+
+To run these patches, simply add `patch: True` to the `Connector` and `Task Runner` settings. For the `Task Runner` also include the name of the Flower app for building and installation.
```
teoparvanov, line 244: What's the purpose of those patches, and in what case would the user want to set this parameter?
This either requires better documentation, or a different naming of the `patch` parameter.

PS: If specific to SGX, we could maybe call this parameter `tee_ready`, or something along those lines?

openfl-workspace/flower-app-pytorch/README.md
teoparvanov: @psfoley, this is Flower Summit slide material haha 😆

openfl/federated/plan/plan.py
kminhta: I'm not opposed to this, but this is true of a lot the `get_x` calls though isn't it? i.e. `self.get_straggler_handling_policy()`, `self.get_assigner()` etc.

edit: this method is called by openfl-security in the extended `plan.py`, so we need to keep it

openfl-workspace/flower-app-pytorch/README.md
kminhta: It's technically not ignored. The entire Flower experiment is encapsulated in a single OpenFL round. If a user set this to 2, then OpenFL would attempt to run another Flower experiment after the first one. Keeping it 1 ensures that OpenFL will shut down the experiment after the end of round check... it's not elegant

openfl-workspace/flower-app-pytorch/src/aggregator.py
kminhta: For sure. One thing to consider, though, we chose to keep it contained in a workspace so that we can completely dump it if we need to without it affecting the rest of the codebase. Ideally, for 1.9+ we can start pulling in robust and maintainable code to the framework-level aggregator, collaborator, etc. and I'm definitely open to discussion around the best way to do that

openfl-workspace/flower-app-pytorch/src/connector.py
kminhta: You're right, we don't need this. It's leftover from when I originally had all this code in the main `openfl` directory (and not `openfl-workspace`. I was trying to "future-proof" the work so that we could use the connector for other frameworks if we start to expand interoperability.

openfl-workspace/flower-app-pytorch/README.md
kminhta: Yeah, SuperLink and SuperNode are deployed in both cases, but they run indefinitely unless they're given a shut down command. Normally, `flwr run` will trigger the SuperLink to spawn the ServerApp as a child process. The ServerApp exists for the duration of a single experiment, so it was recommended as the process to track if we want to signal the end of an experiment, but we can only feasibly do that if we spawn the ServerApp ourselves. So with `auto_shutdown` set to True, we spawn the ServerApp ourselves alongside the SuperLink instead of letting the SuperLink spawn it.

SuperNode/ClientApp behaves the same way (we only spawn SuperNode and it'll spawn the ClientApp itself since those get spawned with each round).

Regarding your PS, I agree, I think that would be a very interesting direction and getting it working in the context of the Workflow API is definitely a future direction

openfl-workspace/flower-app-pytorch/README.md
kminhta: It would be possible to trigger multiple Flower runs on the same SuperNode/SuperLink. It'll be the same OpenFL run technically since the entire Flower experiment is contained in a single OpenFL round. This feature would be really limited in the cases where OpenFL is tracking metrics and other telemetry, but in the event that OpenFL is simply acting as a transport, then it would be feasible. For example, having long lived SuperLink and SuperNode in an SGX enclave. This feature was actually just a byproduct or the long-lived nature of the Flower components, so I consider it limited in practicality

openfl-workspace/flower-app-pytorch/README.md
kminhta: Oh, this is a good catch. It's a typo - it should be the `RandomGroupedAssigner`. I had a separate `ConnectorAssigner` at first, but realized it wasn't necessary. I'll fix this

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/deserialize_message.py
kminhta: Oh, it looks like this was also leftover from previous experiments. I used to deserialize the message to track when the experiment ends, but switched to tracking the ServerApp. That said, maybe it's good to keep for illustration purposes. If not, I can remove and readd it later. It is certainly a useful tool and will be something we'll likely use at some point when we need/want to introspect the Flower message, whether for security verification purposes or something else

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/local_grpc_server.py
kminhta: It's the AggregatorGRPCClient that the Task Runner uses

openfl-workspace/flower-app-pytorch/src/runner_flower.py
kminhta: Ideally, but since this is local communication (this is the ClientApp communicating with the SuperNode) we needed them to be separate ports or else we'd run into issues. There probably is a way to tie this to the plan though so that we don't run into port binding issues on local pocs, and I'd be open to that if/when we start moving this work into the core framework

openfl-workspace/flower-app-pytorch/README.md
kminhta: Yeah, it is specific to SGX. Basically, when a Flower experiment is run, the server will attempt to build and install the Flower App Bundle (FAB). The client will request the FAB then install it on their system (not too unlike how the OpenFL workspace export/import works). The issue is that Gramine does not support some of the system calls that is used when building or installing. The obvious solution is to build and install them in the workspace beforehand. This only resolves the installation issue since the server and client will skip installation if it already detects an installed FAB, but it'll still try to build. We've been in discussion with Flower about this and we'll likely be able to get rid of these patches within the next release

teoparvanov: Sounds good, thanks for the clarification, @kta-intel !

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/deserialize_message.py
teoparvanov: I think it makes sense to keep it around, but ideally show where it can be wired. I suppose this would be as part of the local gRPC server that runs on the collaborator?

To additionally emphasize the potential added transparency, I suggest renaming this function to `inspect_flower_message()`

WDYT?

openfl-workspace/flower-app-pytorch/README.md
kminhta: I think we can flesh that out in a whitepaper after the summit once the feature matures more. For now, this README, I just focused on functionality

openfl-workspace/flower-app-pytorch/src/collaborator.py
kminhta: oh you're right :) fixed it

openfl-workspace/flower-app-pytorch/README.md
kminhta: fixed

openfl-workspace/flower-app-pytorch/README.md
kminhta: updated.

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/deserialize_message.py
kminhta: decided to remove it for this version. it's used in openfl-security. it's also not used by the user, so I don't think it's too important to keep for now. we can add it back in later.

openfl-workspace/flower-app-pytorch/src/patch/patch_flwr_build.py
kminhta: updated.

openfl-workspace/flower-app-pytorch/README.md
kminhta: done.

openfl-workspace/flower-app-pytorch/README.md
kminhta: sounds reasonable to me. moved

openfl-workspace/flower-app-pytorch/plan/plan.yaml
kminhta: sounds good to me. it'll be more consistent with our other workspaces. updated

openfl-workspace/flower-app-pytorch/src/aggregator.py
```diff
@@ -0,0 +1,335 @@
...
+                self.model: base_pb2.ModelProto = utils.load_proto(self.init_state_path)
+                self._load_initial_tensors()  # keys are TensorKeys
+
+        self.collaborator_tensor_results = {}  # {TensorKey: nparray}}
+        self._secure_aggregation_enabled = secure_aggregation
+        if self._secure_aggregation_enabled:
+            self.secagg = secagg_setup(self.uuid, self.authorized_cols, self.tensor_db)
+
+        # Callbacks
+        self.callbacks = callbacks_module.CallbackList(
```
theakshaypant: Duplicate `callbacks` initialisation from line 122. 
Is being fixed in base `Aggregator` class in #1420.

openfl/protocols/aggregator.proto
```diff
@@ -69,3 +70,9 @@ message TaskResults {
 message SendLocalTaskResultsResponse {
   MessageHeader header = 1;
 }
+
+message DropPod {
+  MessageHeader header = 1;
+  DataStream message = 2;
+  map<string, string> metadata = 3;
+}
```
teoparvanov: I think `InteropPod` or `InteropWrapper`/`InteropMessage` would be more meaningful

openfl/protocols/aggregator.proto
```diff
@@ -12,6 +12,7 @@ service Aggregator {
   rpc GetTasks(GetTasksRequest) returns (GetTasksResponse) {}
   rpc GetAggregatedTensor(GetAggregatedTensorRequest) returns (GetAggregatedTensorResponse) {}
   rpc SendLocalTaskResults(stream DataStream) returns (SendLocalTaskResultsResponse) {}
+  rpc PelicanDrop(DropPod) returns (DropPod) {}
```
teoparvanov: Expanding on the previous comment, how about:
```
rpc InteropRelay(InteropPod) returns (InteropPod) {}
```
or simply:
```
rpc Relay(InteropPod) returns (InteropPod) {}
```

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/local_grpc_server.py
```diff
@@ -0,0 +1,141 @@
...
+from src.grpc.connector.flower.message_conversion import flower_to_openfl_message, openfl_to_flower_message
+from multiprocessing import cpu_count
+import signal
+import psutil
+import time
+
+logger = logging.getLogger(__name__)
+
+
+class LocalGRPCServer(grpcadapter_pb2_grpc.GrpcAdapterServicer):
```
teoparvanov: I think `FlowerInteropServer` better describes what this component actually does. WDYT?

openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/local_grpc_client.py
```diff
@@ -0,0 +1,54 @@
+import grpc
+from flwr.proto import grpcadapter_pb2_grpc
+from src.grpc.connector.flower.message_conversion import flower_to_openfl_message, openfl_to_flower_message
+from logging import getLogger
+
+class LocalGRPCClient:
```
teoparvanov: In the same vein as above, how about `FlowerInteropClient`?

teoparvanov (approved):
Thanks for addressing all my comments so far, @kta-intel! I'm following up with a few naming-related suggestions, to (hopefully) help improve code readability and assist the intuition.

Looking good, otherwise, so I'm approving as a "v0.1" OpenFL/Flower interop framework!

openfl-workspace/flower-app-pytorch/src/connector.py
kminhta: Connector ABC removed for this workspace. Can consider having one in the future when we start migrating to the core level

openfl/protocols/aggregator.proto
kminhta: I like it. Done

openfl/transport/grpc/aggregator_server.py
```diff
@@ -54,6 +54,19 @@ def __init__(
...
+            self.use_connector = self.aggregator.is_connector_available()
+        else:
+            self.use_connector = False
+
+        if self.use_connector:
+            self.local_grpc_client = (
+                self.aggregator.get_local_grpc_client()
+            )  # Initialize the local gRPC client
+        else:
+            self.local_grpc_client = None
```
teoparvanov: To complete the renaming, I suggest calling those resp. `interop_client` and `get_interop_client()`

openfl/transport/grpc/aggregator_server.py
```diff
@@ -269,10 +288,46 @@ def SendLocalTaskResults(self, request, context):  # NOQA:N802
...
+
+        Returns:
+            aggregator_pb2.PelicanDrop: The response to the
+            request.
+        """
+        if not self.use_connector:
+            context.abort(
+                grpc.StatusCode.UNIMPLEMENTED,
+                "PelicanDrop is only available in federated interopability mode.",
+            )
```
teoparvanov: nit - some pelican drop residue... 😅

openfl-workspace/flower-app-pytorch/src/aggregator.py
kminhta: good catch, thanks!

kminhta:
Thanks @teoparvanov , FYI I added a small change since you approved

In [Dockerfile.workspace](https://github.com/securefederatedai/openfl/blob/08bc48d2859ee969deadbf186f071635f423aedb/openfl-docker/Dockerfile.workspace#L22), since the flower app is installed via a relative path, you need to step into the workspace to properly execute. It doesn't break any other workspaces

teoparvanov (approved):
Approving again, following the latest set of changes. Thanks, @kta-intel !

openfl-workspace/flower-app-pytorch/src/aggregator.py
psfoley: @MasterSkepticista agree this is a valid point. This recommendation came at my request to limit short term changes to core OpenFL components while working on this PoC. As @kta-intel suggests, it gives time to consider how best to integrate these changes in OpenFL 1.9 and beyond.

openfl-workspace/workspace/plan/defaults/tasks_connector.yaml
```diff
@@ -0,0 +1,4 @@
+start_client_adapter:
+  function : start_client_adapter
+  kwargs   :
+    local_server_port : 0 # local grpc server, 0 to dynamically allocate
```
psfoley, line 4: This may be a small item, but for a future PR I would like to see consistency between how gRPC ports are assigned. We use [the `auto` keyword today](https://github.com/securefederatedai/openfl/blob/develop/openfl/federated/plan/plan.py#L312) and have our own logic to calculate the port based on the hash of the plan. Because this is a local port that won't effect other distributed components, I don't think this is a big deal, but consistency in setting the port parameter would result in one less lookup for users.

psfoley (approved):
Nice work, @kta-intel! Great to see the first phase of Open(FL)ower come to fruition 🚀

PR #1434, opened 2025-03-09
Verifiable Dataset

Efrat1:
## Summary
Detailed Python classes and interfaces design for managing multiple data sources within a verifiable dataset, supporting for now local filesystems.

## Type of Change (Mandatory)
New infrastructure for managing datasets and data sources.

## Description (Mandatory)
new versatile VerifiableDataset class in OpenFL comprising multiple data sources (also used by SFAI for validation).
The VerifiableDataset supports 2 validation routines:
verbose: a hash per file
concise: a hash of all hashes of all files

S3 datasource is planned to be supported in next steps.

## Testing
Unittests
Manual testing with SFAI


rahulga1:
Thanks @Efrat1 for your PR, here are few generic comments:
1. Can you please fill the template properly, also as this is opensource project, please give some details as other users wont be able to access Intel JIRA page.
2. For all function, please use pep-8 style function doc string 
3. Avoid using print, instead user logger.

openfl/utilities/verifdataset/local_data_source.py
```diff
@@ -0,0 +1,62 @@
...
+    """This class represents a local data source."""
+
+    def __init__(self, source_path: Path, hash_func=sha384, maxDatasetSize=0):
+        super().__init__(DataSourceType.LOCAL)
+        self.source_path = Path(source_path)
+        self.hash_func = hash_func
+        self.max_dataset_size = maxDatasetSize
+
+    def enumerate_objects(self, base_dir: Path):
+        """Walk through the data directory and collect all files."""
```
vrancurel: do you get the real path ? os.path.realpath()

tests/openfl/utilities/test_verifiable_dataset.py
```diff
@@ -0,0 +1,263 @@
...
+from openfl.utilities.verifdataset.local_data_source import LocalDataSource
+from openfl.utilities.verifdataset.verifiable_dataset import DatasetFormat, VerifiableDataset
+import pytest
+import os
+import shutil
+from pathlib import Path
+from typing import List, Tuple
+
+@pytest.fixture(scope="class")
+def data_sources(tmp_path_factory) -> Tuple[Path, Path]:
```
vrancurel: can you add a directory with a symlink to test the realpath ?

tests/openfl/utilities/test_verifiable_dataset.py
```diff
@@ -0,0 +1,263 @@
+import json
+from openfl.utilities.verifdataset.local_data_source import LocalDataSource
+from openfl.utilities.verifdataset.verifiable_dataset import DatasetFormat, VerifiableDataset
+import pytest
+import os
+import shutil
+from pathlib import Path
+from typing import List, Tuple
+
+@pytest.fixture(scope="class")
```
vrancurel: can you use pyfakefs in your fixture ? this will prevent creating real files and folders, which is never good in unit tests. This is in the same vein as mock_aws.

vrancurel (changes requested):

openfl/utilities/verifdataset/local_data_source.py
Efrat1: yes, it is done when register dataset before creating dataset_info.
The datasource itself does not hold the full path but only relative path, which is calculated before creating the datasources (against the real path of course), so it is not really testable but anyway I added some tests as we use in the test the same logic of getting the base path and relative paths.

tests/openfl/utilities/test_verifiable_dataset.py
Efrat1: see above

tests/openfl/utilities/test_verifiable_dataset.py
Efrat1: done

tests/openfl/utilities/test_verifiable_dataset.py
vrancurel: I mean as a unit test

tests/openfl/utilities/test_verifiable_dataset.py
```diff
@@ -0,0 +1,332 @@
+import json
+from openfl.utilities.verifdataset.local_data_source import LocalDataSource
+from openfl.utilities.verifdataset.verifiable_dataset import DatasetFormat, VerifiableDataset
+import pytest
+import os
+from pathlib import Path
+from typing import List, Tuple
+
```
vrancurel: I don't see where you import and use pyfakefs in the tests ?

Efrat1: The fs fixture that is provided to the data_sources fixture.
AFAIK there no need to import anything

openfl/utilities/verifdataset/data_source.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        self.datasource_type = datasource_type
+
+    @abstractmethod
+    def compute_object_hash(self, file_path: str) -> str:
+        """
+        Compute the hash of the object or file.
+
+        Args:
+            file_path (str): Path to the file.
```
teoparvanov: Instead of `file_path`, could we consider something more generic like `path` or `url` ? This way, the interface would generalize better beyond files - towards objects or other types of storage.

openfl/utilities/verifdataset/data_source.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def enumerate_objects(self, base_dir):
+        """
+        Enumerate all files in the data source.
+
+        Args:
+            base_dir (str): the base directory of the data source.
```
teoparvanov: Could we use the more general `base_path` or `base_url` notion here? Also, please add a type hint. Is this of type `str` or `Path` ?

openfl/utilities/verifdataset/data_source.py
```diff
@@ -0,0 +1,73 @@
...
+            list: A list of objects and labels
+        """
+        yield
+
+    @classmethod
+    @abstractmethod
+    def from_dict(cls, ds_dict: dict):
+        """
+        Create a DataSource from a dictionary. This method is used when deserializing a
+        VerifiableDataset from JSON.
```
teoparvanov: The `DataSource` class isn't aware of the `VerifiableDataset`, so I suggest not mentioning it here. IMO it's sufficient to say that this factory method is used for creating a `DataSource` from a dictionary.

openfl/utilities/verifdataset/__init__.py
teoparvanov, line 1: I suggest moving & renaming this package under `openfl/federated/data/sources`, to:
- highlight it as a core framework feature
- keep it close to the data loader modules

openfl/utilities/verifdataset/local_data_source.py
```diff
@@ -0,0 +1,62 @@
...
+from hashlib import sha384
+from pathlib import Path
+
+from openfl.utilities.verifdataset.data_source import DataSource, DataSourceType
+
+
+class LocalDataSource(DataSource):
+    """This class represents a local data source."""
+
+    def __init__(self, source_path: Path, hash_func=sha384, maxDatasetSize=0):
```
teoparvanov: let's be consistent: `max_data_size`

openfl/utilities/verifdataset/local_data_source.py
```diff
@@ -0,0 +1,62 @@
...
+            raise ValueError(f"skip local dataset: {full_path} is not a valid file or directory")
+
+        if self.max_dataset_size > 0:
+            total_size_gb = total_size_bytes / (1024 ** 3)
+            if total_size_gb > self.max_dataset_size:
+                raise ValueError(
+                    f"Total dataset size: {total_size_gb:.2f}GB exceeds {self.max_dataset_size}GB"
+                )
+
+        return dataset_files, relative_path
```
teoparvanov: I suggest refactoring this method as a [generator](https://www.geeksforgeeks.org/generators-in-python/), to accommodate for iterating over datasets with large amounts of entries. Relying on the fact that `glob` is a generator itself, it would roughly look something like:

```python
        if full_path.is_dir():
            for file_path in full_path.glob("**/*.*"):
                if self.max_dataset_size > 0:
                    total_size_bytes += file_path.stat().st_size
                    total_size_gb = total_size_bytes / (1024 ** 3)
                    if total_size_gb > self.max_dataset_size:
                        raise ValueError(
                            f"Total dataset size: {total_size_gb:.2f}GB exceeds {self.max_dataset_size}GB"
                        )
                yield file_path, file_path.relative_to(full_path).as_posix()

        elif full_path.is_file():
            if self.max_dataset_size > 0:
                total_size_bytes = full_path.stat().st_size
                total_size_gb = total_size_bytes / (1024 ** 3)
                if total_size_gb > self.max_dataset_size:
                    raise ValueError(
                        f"Total dataset size: {total_size_gb:.2f}GB exceeds {self.max_dataset_size}GB"
                    )
            yield full_path, full_path.name
```

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+            val = val.value  # Convert Enum to its value
+        elif isinstance(val, Path):
+            val = str(val)  # Convert Path to string
+        serializable_dict[key] = val
+    return serializable_dict
+
+
+class VerifiableDataset:
+    """
+    This class represents a data set that can be verified.
```
teoparvanov: ... a data set whose integrity can be verified

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+    It is containing multiple data sources and contains methods to compute the commitment,
```
teoparvanov: nit-grammar: 
```
It contains multiple data sources and methods to compute data commitments, verification, as well as utilities for serialization and deserialization.
```

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+    validation, serialization and deserialization
+    """
+
+    def __init__(
+        self,
+        data_sources: List[DataSource],
+        base_dir: Path,
```
teoparvanov: `base_path`, as discussed above?

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def create_dataset_hash(self):
+        """Create the hash of the dataset according to dataset format."""
+        if self.dataset_format == DatasetFormat.VERBOSE:
+            return self._create_verbose_dataset_hash()
+        if self.dataset_format == DatasetFormat.CONCISE:
+            return self._create_concise_dataset_hash()
+        raise ValueError("Unknown dataset format")
+
+    def _validate_verbose_dataset_info(self, valid_hash):
```
teoparvanov: Why do you need to pass `valid_hash` as a parameter, given that it's available as an object member under `self.hash`?

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+
+    def __init__(
+        self,
+        data_sources: List[DataSource],
+        base_dir: Path,
+        label: str,
+        metadata,
+        dataset_format: DatasetFormat,
+        valid_hash=None,
```
teoparvanov: I suggest simply `hash=None`

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+
+from openfl.utilities.verifdataset.data_source import DataSource, DataSourceType
+from openfl.utilities.verifdataset.local_data_source import LocalDataSource
+
+
+class DatasetFormat(Enum):
+    """Enum for the different dataset commitment formats."""
+
+    VERBOSE = "verbose_dataset"
+    CONCISE = "concise_dataset"
```
teoparvanov: Following the docstring, I suggest renaming this enum to `DatasetCommitment`. Down the road, we could imagine adding a MERKLE_TREE entry here.

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+            base_dir=base_dir,
+            label=data_dict["label"],
+            metadata=data_dict["metadata"],
+            dataset_format=DatasetFormat(data_dict["format"]),
+            valid_hash=data_dict["hash"],
+        )
+
+    @staticmethod
+    def validate_dataset(json_str, base_dir: Path):
+        """Deserialize the VerifiableDataset from JSON and validate it."""
```
teoparvanov: I suggest renaming this to `load_and_verify_dataset`, to reflect the actual implementation, and for consistency with the `verify_dataset` instance method.

teoparvanov (commented):
This is a good starting point, thanks for kicking this work off, @Efrat1 !
I have a few comments and questions from my first review pass:

openfl/utilities/verifdataset/verifiable_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+            continue  # Skip methods
+        if isinstance(val, Enum):
+            val = val.value  # Convert Enum to its value
+        elif isinstance(val, Path):
+            val = str(val)  # Convert Path to string
+        serializable_dict[key] = val
+    return serializable_dict
+
+
+class VerifiableDataset:
```
teoparvanov: @Efrat1 and @vrancurel, if we wanted to introduce Merkle tree-based verification (esp. for individual files). I suppose this class could be extended to a `MerkleVerifiableDataset` and override the `verify_dataser` and `verify_single_file` methods? What other changes would be involved?

openfl/utilities/verifdataset/local_data_source.py
Efrat1: done

openfl/utilities/verifdataset/data_source.py
Efrat1: done

openfl/utilities/verifdataset/data_source.py
Efrat1: renamed to `base_path` and add type hint (str)

openfl/utilities/verifdataset/data_source.py
Efrat1: renamed to `path`

tests/openfl/utilities/test_verifiable_dataset.py
Efrat1: currently there are the test cases:
test_two_local_datasource_concise_with_symlink
test_two_local_datasource_different_base_dir_concise_with_symlink
test_two_local_datasource_verbose_with_symlink
test_two_local_datasource_different_base_dir_verbose_with_symlink
 (in test__verifiable_dataset.py)

what do I miss?

openfl/utilities/verifdataset/verifiable_dataset.py
Efrat1: done

openfl/utilities/verifdataset/verifiable_dataset.py
Efrat1: done

openfl/utilities/verifdataset/verifiable_dataset.py
Efrat1: done

openfl/utilities/verifdataset/verifiable_dataset.py
Efrat1: right. done

openfl/utilities/verifdataset/verifiable_dataset.py
Efrat1: yeah. some linter I tried didn't like it but I see that ours is good with that so I renamed to `hash`. thanks

openfl/utilities/verifdataset/verifiable_dataset.py
Efrat1: maybe load may be bit confusing as loading is used when talk about loading the actual data, I renamed to `deserialize_and_verify`, let me know what you think

openfl/utilities/verifdataset/verifiable_dataset.py
vrancurel: We need to define an abstract interface "commitment" with 2 methods create and verify. We need to think this through. For now we can merge as is, as the focus will be S3. We can create the commitment interface after.

openfl/utilities/verifdataset/__init__.py
vrancurel: OK

openfl/utilities/verifdataset/verifiable_dataset.py
teoparvanov: Sounds good, thanks @Efrat1 !

openfl/utilities/verifdataset/verifiable_dataset.py
teoparvanov: Sounds good, thanks for the clarification!

openfl/utilities/verifdataset/__init__.py
Efrat1: Do you mean to rename verifdataset to sources? and inside have pytorch/tf/etc. dirs as now?

openfl/utilities/verifdataset/local_data_source.py
Efrat1: nice. done

vrancurel (changes requested):
- Please move the `verifdataset` folder in the location suggested by Teo.

openfl/utilities/verifdataset/__init__.py
Efrat1: moved

openfl/federated/data/sources/pytorch/local_image_folder.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Union
+
+from PIL import Image
+
+from openfl.federated.data.sources.pytorch.local_folder import LocalFolder
+
+# from openfl.federated.data.sources.pytorch.local_folder import LocalFolder
+
+
+class LocalImageFolder(LocalFolder):
```
teoparvanov: To illustrate (and validate) the new abstractions and APIs, I suggest refactoring one of the existing data loaders to the new data set/data source abstractions. For example, there's an opportunity to replace the current custom `HistologyDataset` class which does some custom integrity checks with a `VerifiableImageFolder`:
https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch/histology/src/dataloader.py#L60

Ideally, this would be included in the current PR in addition to the unit tests (which are very thorough, by the way, great job!).

WDYT?

openfl/federated/data/sources/verifiable_dataset_info.py
```diff
@@ -0,0 +1,167 @@
...
+            continue  # Skip methods
+        if isinstance(val, Enum):
+            val = val.value  # Convert Enum to its value
+        elif isinstance(val, Path):
+            val = str(val)  # Convert Path to string
+        serializable_dict[key] = val
+    return serializable_dict
+
+
+class VerifiableDatasetInfo:
```
teoparvanov, line 14: Why not simply `VerifiableDataset` ?

openfl/federated/data/sources/data_source.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class DataSourceType(Enum):
+    """Enum for the different types of data sources."""
+
+    LOCAL = "local"
+    S3 = "s3"
+
+
+class DataSource(ABC):
```
teoparvanov, line 20: I'd also suggest adding some documentation and examples on how the dataset/data source abstractions can be used in practice - for example under the [Developer Guide/OpenFL Utilities](https://openfl.readthedocs.io/en/latest/developer_guide/utilities.html) section. This could be a separate `.rst` document plugged in here:
https://github.com/securefederatedai/openfl/tree/develop/docs/developer_guide/utilities

PS: this could be done in a separate PR

openfl/federated/data/sources/local_data_source.py
```diff
@@ -0,0 +1,57 @@
...
+            for file_path in full_path.glob("**/*.*"):
+                if self.max_dataset_size > 0:
+                    total_size_bytes += file_path.stat().st_size
+                    total_size_gb = total_size_bytes / (1024**3)
+                    if total_size_gb > self.max_dataset_size:
+                        raise ValueError(
+                            f"Total dataset size: {total_size_gb:.2f} GB exceeds "
+                            f"{self.max_dataset_size} GB"
+                        )
+                yield file_path
```
timovirt: was this supposed to return `label` also as defined in the base class
```
        Returns:
            list: A list of objects and labels
```
or should we change the docstring to indicate that labels are part of the file/objectnames

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,71 @@
...
+    def get_label_name(self, index: int) -> str:
+        """Retrieves the original label name from an index."""
+        return self.idx_to_label.get(index, "Unknown")
+
+
+class LocalFolder(Dataset):
+    def __init__(self, base_path, label_mapper: LabelMapper, transform=None):
+        """
+        Args:
+            base_path (str or Path): Root directory containing labeled image subdirectories.
```
timovirt: `label_mapper` is missing

openfl/federated/data/sources/pytorch/local_image_folder.py
```diff
@@ -0,0 +1,25 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from pathlib import Path
+from typing import Union
+
+from PIL import Image
+
+from openfl.federated.data.sources.pytorch.local_folder import LocalFolder
+
+# from openfl.federated.data.sources.pytorch.local_folder import LocalFolder
```
timovirt:
```suggestion
```

openfl/federated/data/sources/data_source.py
```diff
@@ -0,0 +1,72 @@
...
+        Args:
+            path (str): Path to the file.
+
+        Returns:
+            str: The file's hash.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def enumerate_objects(self, base_path: str):
```
teoparvanov: @Efrat1 , could you add a return type hint?

openfl/federated/data/sources/local_data_source.py
```diff
@@ -0,0 +1,57 @@
...
+class LocalDataSource(DataSource):
+    """This class represents a local data source."""
+
+    def __init__(self, source_path: Path, hash_func=sha384, max_dataset_size=0):
+        super().__init__(DataSourceType.LOCAL)
+        self.source_path = Path(source_path)
+        self.hash_func = hash_func
+        self.max_dataset_size = max_dataset_size
+
+    def enumerate_objects(self, base_path: str):
```
MasterSkepticista: Does this need to be a generator? Generators are useful when function calls are expensive. To be coherent with function name, I suggest returning a fully formed list instead.

openfl/federated/data/sources/local_data_source.py
```diff
@@ -0,0 +1,57 @@
...
+        """Enumerate all files in the data source."""
+        total_size_bytes = 0
+        full_path = Path(base_path) / self.source_path
+        if full_path.is_dir():
+            for file_path in full_path.glob("**/*.*"):
```
MasterSkepticista: Isn't `os.walk` method more readable? I think the goal is to recursively scan through all files in a given directory (?)

openfl/federated/data/sources/local_data_source.py
```diff
@@ -0,0 +1,57 @@
...
+                total_size_bytes = full_path.stat().st_size
+                total_size_gb = total_size_bytes / (1024**3)
+                if total_size_gb > self.max_dataset_size:
+                    raise ValueError(
+                        f"Total dataset size: {total_size_gb:.2f} GB exceeds "
+                        f"{self.max_dataset_size} GB"
+                    )
+            yield full_path
+
+    def compute_object_hash(self, path: str) -> str:
```
MasterSkepticista: `compute_file_hash`?

openfl/federated/data/sources/local_data_source.py
```diff
@@ -0,0 +1,57 @@
...
+        """Compute the hash of the file. Return hash on hexstring format."""
+        hash_obj = self.hash_func()
+        with open(path, "rb") as file:
+            for byte_block in iter(lambda: file.read(65536), b""):
+                hash_obj.update(byte_block)
+            return hash_obj.hexdigest()
```
MasterSkepticista: Nit: We can indent the `return` outside `with` block, to ensure file is closed. I am sure python is intelligent enough, but we can't risk memory leaks on areas like these where it may go unnoticed!

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,71 @@
...
+        """Assigns or retrieves the index of a label."""
+        if label not in self.label_to_idx:
+            new_index = len(self.label_to_idx)
+            self.label_to_idx[label] = new_index
+            self.idx_to_label[new_index] = label
+        return self.label_to_idx[label]
+
+    def get_label_name(self, index: int) -> str:
+        """Retrieves the original label name from an index."""
+        return self.idx_to_label.get(index, "Unknown")
```
MasterSkepticista: Is `None` as default better? It is easier to catch `None` in exceptions than a string. Labels are also strings and this may actually be treated as a valid class.

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,71 @@
...
+                label_name = file_path.parent.name
+                label_idx = self.label_mapper.get_label_index(label_name)  # Use common mapping
+                self.samples.append((file_path, label_idx))
+
+    @abstractmethod
+    def load_file(self, file_path):
+        """Load a file from the dataset."""
+        return NotImplementedError
+
+    def __getitem__(self, index):
```
MasterSkepticista: Please add return type hints.

On another note, dictionary return types are extensible and backwards compatible when more data is to be returned in the future. This is common in TFDS/HF and advanced dataloaders.

openfl/federated/data/sources/pytorch/local_folder.py
Efrat1: Added, thanks.

openfl/federated/data/sources/local_data_source.py
Efrat1: You're right, description is wrong, fixed. Thanks

openfl/federated/data/sources/verifiable_dataset_info.py
Efrat1: Because it is not really a dataset, we can't use it for loading the data.

openfl/federated/data/sources/verifiable_dataset_info.py
```diff
@@ -0,0 +1,146 @@
...
+
+from openfl.federated.data.sources.data_source import DataSource, DataSourceType
+from openfl.federated.data.sources.local_data_source import LocalDataSource
+
+
+class DatasetFormat(Enum):
+    """Enum for the different dataset commitment formats."""
+
+    VERBOSE = "verbose_dataset"
+    CONCISE = "concise_dataset"
```
MasterSkepticista: Nit: Can be less redundant in naming ("verbose"/"concise" is ok)

openfl/federated/data/sources/verifiable_dataset_info.py
```diff
@@ -0,0 +1,146 @@
...
+        hashes = self._create_verbose_dataset_hash()
+        return sorted(hashes.values()) == sorted(self.all_hashes.values())
+
+    def _validate_concise_dataset_info(self):
+        concise_hash = self._create_concise_dataset_hash()
+        return concise_hash == self.root_hash
+
+    def verify_dataset(self, dataset_info=None):
+        """Verify the dataset root_hash."""
+        if dataset_info is None and self.root_hash is None:
```
MasterSkepticista: What does the `and` check? From the exception it seems the intent is to only check `if not dataset_info: raise ValueError`. Potential for simplification or clarification in the error message.

openfl/federated/data/sources/verifiable_dataset_info.py
```diff
@@ -0,0 +1,146 @@
...
+    VERBOSE = "verbose_dataset"
+    CONCISE = "concise_dataset"
+
+
+def filter_non_serializable(obj):
+    """Filter out methods and non-serializable objects."""
+    serializable_dict = {}
+    for key, val in obj.__dict__.items():
+        if callable(val):
+            continue  # Skip methods
```
MasterSkepticista: Strange... How would a callable make its way into a dataset source? Am I missing something?

openfl/federated/data/sources/pytorch/verifiable_map_style_image_folder.py
```diff
@@ -0,0 +1,35 @@
...
+                )
+                datasources.append(
+                    LocalImageFolder(
+                        datasource_full_path, self.label_mapper, transform=self.transform
+                    )
+                )
+            # elif data_source.datasource_type == DataSourceType.S3:
+            #     self.datasources.append(
+            #         S3ImageFolder(data_source.url, self.label_mapper, transform=self.transform)
+            #     )
```
MasterSkepticista: Commented code blocks are generally not recommended. Please either mark the code active with conditionals, or suggest removing it altogether. It becomes a source of confusion for future readers

openfl/federated/data/sources/local_data_source.py
Efrat1: I kept it as a generator for efficiency with large datasets and added a return type hint (Generator[str, None, None]) to clarify its behavior. Let me know if that works.

openfl/federated/data/sources/local_data_source.py
Efrat1: yes, nice. thanks

openfl/federated/data/sources/local_data_source.py
Efrat1: nice catch!

openfl/federated/data/sources/verifiable_dataset_info.py
Efrat1: The verification can be against dataset_info or already saved root_hash (for when we don't have the dataset_info anymore). I'll fix the error message for clarification, thanks.

openfl/federated/data/sources/verifiable_dataset_info.py
Efrat1: because we serialize DataSources class objects, so the default is to include the methods.

openfl/federated/data/sources/pytorch/verifiable_map_style_image_folder.py
Efrat1: done

openfl/federated/data/sources/pytorch/local_folder.py
Efrat1: done

openfl/federated/data/sources/local_data_source.py
vrancurel: In S3, files are called "objects" so it could be more generic. But agree maybe for the data scientist community it might be confusing, let's rename it to file.

Efrat1: NP. So same for `enumerate_objects` -> `enumerate_files`

openfl/federated/data/sources/local_data_source.py
MasterSkepticista: How about an iterator method like `__getitem__(self, idx)`? This is programmatically simpler and may enhance readability:

```python
ds = LocalDataSource(...)
for f in ds:  # __getitem__ is implicit
  ...

# instead of
ds = LocalDataSource(...)
for f in ds.enumerate_objects(base_path):
  ...
```

If there are other reasons to not use iterators, current implementation is OK too

openfl/federated/data/sources/verifiable_dataset_info.py
MasterSkepticista: That makes sense, but then can't we filter them at source (i.e. during the serialization step)?

Efrat1: With that function we didn't have to implement serialization in the sources.
However, following your comment I think that it makes more sense to enable each data source to control its serialization,
so now this logic is in DataSource (`_serialize_fields`) and can be overridden in subclasses if needed.

openfl/federated/data/sources/local_data_source.py
Efrat1: The thing is that we don't have access by index.
Maybe we can use `__iter__`, do you think it would be better?

MasterSkepticista: I am leaning towards `__getitem__`. Now that I think about it, it links to our previous comments on generators vs materialized lists.

Indexed fetching of samples is very common in data science workflows (e.g. selecting specific samples for evaluation/inference, or index-based dataset sharding for multi-GPU training - a strong possibility)

I think it bears importance to revisit - how costly is it to fully materialize the list? Based on the function definition, the operations seem trivial and wouldn't take up significant memory. Thoughts?

Edit: To cover both cases, `__item__` could just `return iter(self._file_paths)` or equivalent. It gives the flexibility of a generator as well as indexed fetching.

openfl/federated/data/sources/data_source.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    def __init__(self, datasource_type: DataSourceType):
+        """
+        Initialize a DataSource.
+
+        Args:
+            datasource_type (DataSourceType): The storage type of the data source.
+        """
+        self.datasource_type = datasource_type
```
MasterSkepticista: Nit: `self.type` is cognitively simpler.

```python
dataset_source = DataSource(...)
print(dataset_source.type)  # more readable
print(dataset_source.datasource_type)
```

openfl/federated/data/sources/local_data_source.py
Efrat1: Actually this class is not really used for loading the data items, but the classes that inherit from `torch.utils.data.Dataset`, which do implement `__getitem__` (e.g. `VerifiableMapStyleDataset`) so I think we don't have to stick to the indexed fetching in our case.

vrancurel: As @Efrat1 said the VerifiableDataset is not supposed to be loaded directly. I think it could be interesting to support this method at some point, but I would prefer focusing on the functionality for now.

teoparvanov (approved):
Looking good, thanks @Efrat1 and @vrancurel !

I approve, assuming my comments [here](https://github.com/securefederatedai/openfl/pull/1434#discussion_r2001284881) and [here](https://github.com/securefederatedai/openfl/pull/1434#discussion_r2001498770) are going to be addressed in subsequent PRs (esp. when S3 data loading becomes available in OpenFL).

openfl/federated/data/sources/pytorch/local_image_folder.py
teoparvanov: PS: as discussed with @vrancurel , this can be illustrated with an example workspace where data is loaded from S3 (in a subsequent PR)

openfl/federated/data/sources/data_source.py
```diff
@@ -0,0 +1,90 @@
...
+from enum import Enum
+from pathlib import Path
+from typing import Any, Dict, Generator
+
+
+class DataSourceType(Enum):
+    """Enum for the different types of data sources."""
+
+    LOCAL = "local"
+    S3 = "s3"
```
teoparvanov: nit - this can be written more elegantly as:
```python
    LOCAL = auto()
    S3 = auto()
```

openfl/federated/data/sources/verifiable_dataset_info.py
```diff
@@ -0,0 +1,129 @@
...
+from typing import List
+
+from openfl.federated.data.sources.data_source import DataSource, DataSourceType
+from openfl.federated.data.sources.local_data_source import LocalDataSource
+
+class DatasetCommitment(Enum):
+    """Enum for the different dataset commitment formats."""
+
+    VERBOSE = "verbose"
+    CONCISE = "concise"
```
teoparvanov: Same here:
```python
    VERBOSE = auto()
    CONCISE = auto()
```

test-requirements.txt
```diff
@@ -8,3 +8,5 @@ defusedxml==0.7.1
 matplotlib==3.10.1
 fpdf2==2.8.2
 papermill==2.6.0
+pyfakefs
+torch
```
teoparvanov: It may not be a bad idea to "pin" torch to a specific version. In Task Runner API workspaces, we currently support up to `2.4.1`.

PS: same goes for `pyfakefs` - ideally you would pin to a specific version for improved test stability

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,73 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from abc import abstractmethod
+from pathlib import Path
+from typing import Any, Dict
+
+from torch.utils.data import Dataset
+
+
+class LabelMapper:
```
teoparvanov: could you add a docstring for this class?

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self):
+        self.label_to_idx = {}
+        self.idx_to_label = {}
```
teoparvanov: shouldn't there be a way to provide those when initializing the `LabelMapper` object?

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,73 @@
...
+            if file_path.is_file():
+                # Get parent directory as label
+                label_name = file_path.parent.name
+                label_idx = self.label_mapper.get_label_index(label_name)  # Use common mapping
+                self.samples.append((file_path, label_idx))
+
+    @abstractmethod
+    def load_file(self, file_path):
+        """Load a file from the dataset."""
+        return NotImplementedError
```
teoparvanov: I think you can simply `pass` here, as the `@abstractmethod` decorator takes care of enforcing implementation

teoparvanov (commented):
\+ a couple of minor comments on code quality:

openfl/federated/data/sources/pytorch/local_folder.py
Efrat1: The labels are dynamically discovered during the data loading process by different data sources, inferred from the folder hierarchy, similar to how PyTorch's `DatasetFolder` and `ImageFolder` work. This class is shared among all data sources, allowing them to use the same mapper while loading samples, ensuring consistent label mapping across them

openfl/federated/data/sources/pytorch/local_folder.py
Efrat1: thanks. fixed also in other abstract methods

test-requirements.txt
Efrat1: sure.
I used `2.4.1` for `torch` and `5.8.0` for `pyfakefs` (the version I run and works...)

openfl/federated/data/sources/data_source.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @abstractmethod
+    def enumerate_files(self) -> Generator[str, None, None]:
+        """
+        Enumerate all files in the data source.
+
+        Returns:
+            list: A list of objects
+        """
+        yield
```
ishaileshpant: Query: why not 'pass' or 'NotImplementedError' to make it more explicit and not produce a generator that yields None 
AFAIU this has to be implemented by every concrete source

openfl/federated/data/sources/pytorch/local_folder.py
```diff
@@ -0,0 +1,79 @@
...
+        self.transform = transform
+        self.samples = []
+        self.label_mapper = label_mapper
+
+        # Build the dataset
+        self._load_samples()
+
+    def _load_samples(self):
+        """Recursively find all image files and assign labels based on the directory name."""
+        for file_path in self.base_path.rglob("*.*"):  # Search for all files in subdirectories
```
ishaileshpant: Nit: Better to make it explicit that only file(s) with extensions are included or if its possible to even cater to that edge use-case by using a pattern like "*"

tests/openfl/federated/data/test_verifiable_dataset.py
```diff
@@ -0,0 +1,196 @@
...
+            fs.create_file(new_path, contents=file_content)  # Create file in new location
+
+def test_one_local_datasource(local_data_sources):
+    ds1, _ = local_data_sources
+    base_path, relative_paths = split_to_base_and_relative_paths([ds1])
+    datasources = [LocalDataSource(source_path=rel_path, base_path=base_path) for rel_path in relative_paths]
+    verifiable = VerifiableDatasetInfo(data_sources=datasources, label="my_dataset", metadata="md")
+    hash = verifiable.create_dataset_hash()
+    assert isinstance(hash, str), f"Expected str, got {type(hash)}"
+    verifaible_json = verifiable.to_json()
```
ishaileshpant: Minor : Typo : verifaible_json 
```suggestion
    verifiable_json = verifiable.to_json()
```

tests/openfl/federated/data/test_verifiable_dataset.py
```diff
@@ -0,0 +1,196 @@
...
+    assert VerifiableDatasetInfo.deserialize_and_verify(verifaible_json, base_path)
```
ishaileshpant: Minor : Typo : verifaible_json 
```suggestion
    verifiable_json
```

tests/openfl/federated/data/test_verifiable_dataset.py
```diff
@@ -0,0 +1,196 @@
...
+
+def test_two_local_datasource(local_data_sources):
+    ds1, ds2 = local_data_sources
+    base_path, relative_paths = split_to_base_and_relative_paths([ds1, ds2])
+    datasources = [LocalDataSource(source_path=rel_path, base_path=base_path) for rel_path in relative_paths]
+    verifiable = VerifiableDatasetInfo(data_sources=datasources, label="my_dataset", metadata="md")
+    hash = verifiable.create_dataset_hash()
+    assert isinstance(hash, str), f"Expected str, got {type(hash)}"
+    verifaible_json = verifiable.to_json()
```
ishaileshpant: Minor : Typo : verifaible_json 
```suggestion
    verifiable_json = verifiable.to_json()
```

tests/openfl/federated/data/test_verifiable_dataset.py
ishaileshpant, line 1: Minor : Typo : verifaible_json 
```suggestion
    verifiable_json
``` potentially replicated across test case(s)

File: openfl/federated/data/sources/pytorch/verifiable_map_style_dataset.py [hunk 47]
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, vds: VerifiableDatasetInfo, transform=None, verify_dataset=False):
+        self.transform = transform
+        self.verifiable_dataset_info = vds
+        self.verify_dataset = verify_dataset
+        self.datasources = self.create_datasets()
+
+        # create indices for fast lookup
+        lengths = list(map(len, self.datasources))
+        self.indices = list(chain(*[range(length) for length in lengths]))
```
ishaileshpant: May be i am not reading it right but where is this indices list used?
AFAIU its to be used in __getitem__ but there we are using cumulative_sizes?

Would it be more memory-efficient , specifically for large datasets, to drop holding giant list of indices - wdyt?

openfl/federated/data/sources/pytorch/verifiable_map_style_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+
+from openfl.federated.data.sources.verifiable_dataset_info import (
+    VerifiableDatasetInfo,
+)
+
+
+class VerifiableMapStyleDataset(torch.utils.data.Dataset):
+    """Base class for different types of map style datasets."""
+
+    def __init__(self, vds: VerifiableDatasetInfo, transform=None, verify_dataset=False):
```
ishaileshpant: As per the test case, for verify_dataset=True to work the intended workflow would be to ensure that the VerifiableDatasetInfo hashes are populated right? I suggest to make it either explicitly call that internally when flag is set or atleast document this intended usage / workflow - wdyt?

ishaileshpant: Reason : I suspect verify_single_file will raise a ValueError due to missing hashes if user forgets to call the verification step first - does that make sense

openfl/federated/data/sources/pytorch/local_folder.py
Efrat1: Good question.
@vrancurel / @teoparvanov : do we want to include files with no extension or is it too risky because we might include unwanted files, like hidden system files or directory metadata?

openfl/federated/data/sources/pytorch/verifiable_map_style_dataset.py
Efrat1: Right, just a leftover, thank you!

openfl/federated/data/sources/pytorch/verifiable_map_style_dataset.py
Efrat1: Yes, I see what you mean, good point.
The thing is that we need a trusted hash to compare to when we validate (root hash is good enough).
I modify `verify_single_file` to compute the hashes in case we have the root_hash, otherwise raise error.
Naturally the VerifiableDatasetInfo instance that we get in the datasets were created from dataset_info so they should have root_hash. I think that now we are good, what do you think?

openfl/federated/data/sources/data_source.py
Efrat1: OK, I modified it to `pass`. The return type hint already indicates that we expect a generator. Thanks

openfl/federated/data/sources/pytorch/local_folder.py
vrancurel: I think we should use "*"

openfl/federated/data/sources/pytorch/verifiable_map_style_dataset.py
```diff
@@ -0,0 +1,49 @@
...
+        data_idx = idx - self.cumulative_sizes[dataset_idx - 1] if dataset_idx > 0 else idx
+        item = self.datasets[dataset_idx][data_idx]
+        data_path = item["path"]
+
+        if self.verify_dataset:
+            item_hash = self.verifiable_dataset_info.data_sources[dataset_idx].compute_file_hash(
+                data_path
+            )
+            if not self.verifiable_dataset_info.verify_single_file(data_path, item_hash):
+                raise ValueError(f"Data integrity check failed for {data_path}")
```
teoparvanov: @vrancurel, @Efrat1, I'm wondering if this `verify_dataset` flag is actually useful... If the dataset doesn't need to be verified, then why not simply instantiate a `torch.utils.data.Dataset`? On the other hand, load-time verification would be enforced for the `VerifiableMapStyleDataset`. Otherwise why go through the trouble of setting up this complex class hierarchy?

Efrat1: The `verify_dataset` flag was added to support a feature toggle in openfl-security. For example, `HistologyDataset` inherits from `VerifiableImageFolder` (which extends `VerifiableMapStyleDataset`), so if a user wants to skip per-item verification there, we'd have to restructure things to inherit directly from `ImageFolder` — which isn’t ideal.

BTW, dataset-wide validation (via `VerifiableDatasetInfo`) still happens either way, and we use that info for both serialization and deserialization. It’s also used to initialize `VerifiableImageFolde`r, so the whole setup is still relevant even when `verify_dataset` is off.

To make things clearer, I’ll rename the flag to `verify_dataset_items` (or something along those lines) to better reflect that it controls per-item checks during loading.
Do you think it makes sense to default it to `True`?

teoparvanov: Well, an `if` statement in a method implementation is sometimes (often) a "smell" of some problem with the class hierarchy... Could you consider removing the flag which is confusing IMO and inherit directly from `ImageFolder` if per-item verification isn't required?

teoparvanov: Understood the purpose of the `self.verify_dataset` flag, and the fact that the `VerifiableMapStyleDataset` also adds the ability to use multiple data sources.

openfl/federated/data/sources/torch/local_folder.py
```diff
@@ -0,0 +1,79 @@
...
+            file_data = self.transform(file_data)
+
+        return {"data": file_data, "label": label, "path": str(file_path)}
+
+    def __len__(self):
+        return len(self.samples)
+
+    def get_label_map(self):
+        """Returns the mapping of class names to label indices."""
+        return self.label_to_idx
```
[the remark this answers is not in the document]
Efrat1: fixed

teoparvanov (approved):
Looks great, @Efrat1! Thanks for addressing my final set of comments.

vrancurel (approved):

psfoley (approved):

PR #1435, opened 2025-03-10
Use Single Function For Logging With Rich Text Handler

noopurintel:
## Summary
Single logging mechanism throughout the OpenFL repository.

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
Currently, we have logging functions at three places:
1. `setup_logging` - openfl/interface/cli.py
2. `setup_loggers` - openfl/utilities/logs.py  _(not used anywhere)_
3. `configure_logging` - tests/end_to_end/utils/logger.py   _(used in E2E pytest framework only)_

Intention of this PR is to retain the one in `openfl/utilities/logs.py` (as the file name would be easier to guess) as base logging file and use it wherever needed.

## Testing
Part of PR pipeline itself.

## Additional Information

Current logging (example from E2E test run)
image

Post code change
image


openfl/utilities/logs.py
MasterSkepticista, line 1: Suggest renaming this to `logging.py`.

openfl/utilities/logs.py
```diff
@@ -5,12 +5,15 @@
 """Logs utilities."""
 
 import logging
-
 from rich.console import Console
 from rich.logging import RichHandler
 
+from openfl.utilities import add_log_level
+
+logger = logging.getLogger()
```
MasterSkepticista: Is this required here? Plus we do not want to import this module-level logger from a different module (I saw `from openfl.utilities.logs import logger` somewhere)

Recommended approach is to have `logger = getLogger(__name__)` on top of each module. It assumes logging has been set up globally once

openfl/utilities/logs.py
```diff
@@ -5,12 +5,15 @@
...
+
 
-def setup_loggers(log_level=logging.INFO):
+def setup_loggers(log_level=logging.INFO, log_file=None):
```
MasterSkepticista: `setup_logger` now that we have an opportunity to change.

noopurintel: Modified the function name to `setup_logger`

openfl/utilities/logs.py
noopurintel: I had actually planned to rename it to `logging.py`. Good that you mentioned it. Done now.

openfl/utilities/logs.py
noopurintel: It's not required here. Removed.

MasterSkepticista (approved):
Nice to see a line change ratio <1
@noopurintel Please verify if this has a downstream impact on openfl-security.

noopurintel:
> Nice to see a line change ratio <1 @noopurintel Please verify if this has a downstream impact on openfl-security.

I have checked the codebase of OpenFL security, it isn't using our logging. So, we are good.

rahulga1 (approved):

teoparvanov (approved):

PR #1438, opened 2025-03-11
Deleting AggregatorBasedWorkflow tutorials and examples

teoparvanov:
## Summary
Deleting `AggregatorBasedWorkflow` tutorials and examples, which are currently not well maintained, and not tested as part of CI. At the same time, we preserve the core OpenFL code that enables the `AggregatorBasedWorkflow` (incl. workspace export), which can be updated later.

## Type of Change (Mandatory)
- Cleanup [removing unmaintained examples]

## Description (Mandatory)
The [experimental aggregator and collaborator](https://github.com/securefederatedai/openfl/tree/develop/openfl/experimental/workflow/component) are used under the hood to implement the `FederatedRuntime`. The `AggregatorBasedWorkflow` also uses those components to provide a way to transform Workflow API experiments to TaskRunner-style FL workspaces that can be run in a distributed setup. With the introduction of `FederatedRuntime` though, the `AggregatorBasedWorkflow` is becoming less relevant. In fact, it doesn't seem to be used by anyone.
 
So, my suggestion would be to deprecate (ideally completely remove) the Aggregator-based workflow approach in favor of a production-ready `FederatedRuntime` in OpenFL 1.9. 

The current PR is a first step in this process - removing the example workspaces for `AggregatorBasedWorkflow`. 

## Testing
CI


MasterSkepticista (approved):

kminhta (approved):

psfoley:
@teoparvanov thanks for submitting this - this was a stopgap approach for OpenFL 1.6 before `FederatedRuntime` was implemented, so I'm certainly in support to removing these

psfoley (approved):

PR #1449, opened 2025-03-13
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


openfl-docker/licenses.sh
```diff
@@ -33,8 +33,8 @@ if [ "$INSTALL_SOURCES" = "yes" ]; then
     rm -rf /var/lib/apt/lists/*
 fi
 
-mkdir -p thirdparty
-cd thirdparty
+mkdir -p third-party
+cd third-party
```
yarikoptic: I guess could be reverted/skipped here but I thought "why not" if it is just a bundle of licenses?

teoparvanov:
Hi @yarikoptic, thank you for the contribution! Before integrating [codespell](https://github.com/codespell-project/codespell) into the OpenFL CI pipelines though we have to assess its security and trustworthiness.

In the meantime, to make progress with this PR, I'd suggest removing the _codespell_ CI job for now, and only preserving the typo fixes as part of this PR. The CI integration can be done at a later stage.

Could you update your PR accordingly?

yarikoptic:
dropped commit with CI workflow (kept pre-commit since was not instructed to remove), signed off, force pushed

teoparvanov:
> dropped commit with CI workflow (kept pre-commit since was not instructed to remove), signed off, force pushed

My apologies for the omission, the pre-commit hook will have to be removed too at this stage.

pyproject.toml
```diff
@@ -32,4 +32,10 @@ ignore = [
...
 force-single-line = false
-known-first-party = ["openfl"]
\ No newline at end of file
+known-first-party = ["openfl"]
+[tool.codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = '.git*,*.svg,*.css'
+check-hidden = true
+ignore-regex = '^\s*"image/\S+": ".*'
+ignore-words-list = 'fpr,astroid'
```
teoparvanov: please remove

.pre-commit-config.yaml
```diff
@@ -33,4 +33,11 @@ repos:
...
-        additional_dependencies: ["bandit[toml]"]
\ No newline at end of file
+        additional_dependencies: ["bandit[toml]"]
+  - repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in pyproject.toml
+    rev: v2.4.1
+    hooks:
+    - id: codespell
+      additional_dependencies:
+      - tomli
```
teoparvanov: To be removed too, as discussed in a previous comment. Thanks!

openfl-docker/licenses.sh
teoparvanov: Yes, let's revert this change as it may have an impact on downstream projects that use OpenFL

teoparvanov (changes requested):

pyproject.toml
yarikoptic: will do, although this is "your loss" happen you decide to run the tool to catch fresh typos

MasterSkepticista:
Thanks for the contribution @yarikoptic

PR #1453, opened 2025-03-18
Do not review: Task Runner E2E - Cleanup the participant log file during verification step

noopurintel:
## Summary
Remove the ANSI escape codes and unwanted patterns from participant log files.

## Type of Change (Mandatory)
- E2E Test artifact improvement/correction

## Description (Mandatory)
Currently, the participant log files have ANSI escape codes and various unwanted patterns (after rich text handler was enabled at root level)

image

As part of this PR, I have added a cleanup function which runs as part of verification process and importantly doesn't fail the test in case of an exception.

image

## Testing
PR pipeline contains the impacted tests.


MasterSkepticista:
I think `NO_COLOR=1` will do the trick.

PR #1455, opened 2025-03-18
TaskRunner E2E: Remove ANSI escape codes and unwanted patterns from participant log files with LOG_FILE usage

noopurintel:
## Summary
1. To make use of LOG_FILE environment variable for creating and logging into participant log files during Task Runner federation experiment. This in turn expects changes in federation completion logic as `✔️ OK` is not written to the log file and instead `Experiment Completed` message is to be considered.
2. To keep the participant log files inside `logs` folder present inside `workspace` (instead of directly inside `workspace`).
3. To fix the problem of escape characters making it to the participant log files (in E2E test artifacts).
4. To skip csv files from getting uploaded under artifacts for GaNDLF workflow.
5. To fix typo in one of the workflow step names.

**IMPORTANT** - The format of aggregator/collaborators output log remains same at OpenFL level. It is modified/fixed only in E2E artifacts.

## Type of Change (Mandatory)
- Feature enhancement  
- Test framework improvement

## Description (Mandatory)
During Task Runner federation experiment, we were redirecting the command output of `fx start participant` to respective files and because rich text handler doesn't apply to the file handler, these files had ANSI escape codes and unwanted patterns.

Corrected the same by using LOG_FILE to log the participant federation run output and having a temp file running in background just to keep the process alive.

In `logging.py`, defined a separate formatter for file_handler to display time and filename:lineno details along with content.

## Testing
All the tests are part of PR pipeline.
Successful testing with Dockerized approach - https://github.com/noopurintel/openfl/actions/runs/13924245796

### Before code fix

**Participant log file format (in E2E artifacts):**
image

### After code fix

**Participant log file format (in E2E artifacts):**
image

**GitHub workflow code run:**
image

**Direct command run on terminal (locally):**
image

**Redirected file output on command run (locally):** (for e.g. `fx aggregator start > a.log`)
image


payalcha (approved):

tests/end_to_end/utils/federation_helper.py
```diff
@@ -1131,4 +1147,19 @@ def remove_stale_processes(num_collaborators=0, envoys=[], director=False):
...
+
+    Args:
+        path (str): The path to the workspace to be deleted.
+    """
+    if os.path.islink(path) or os.path.isfile(path):
+        subprocess.run(['sudo', 'rm', '-f', path], check=True)
+    elif os.path.isdir(path):
+        for entry in os.scandir(path):
+            remove_workspace(entry.path)
+        subprocess.run(['sudo', 'rmdir', path], check=True)
```
MasterSkepticista, line 1165: Why not `shutil.rmtree(path)`?

openfl/utilities/logging.py
```diff
@@ -23,26 +23,34 @@ def setup_logger(log_level=logging.INFO, log_file=None):
...
+    # Console handler includes date and log level, do not add it again
+    console_handler.setFormatter(logging.Formatter("%(message)s"))
+    logger.addHandler(console_handler)
+
+    # Create a file handler if log_file is provided
     if log_file:
+        file_formatter = logging.Formatter(
+            "[%(asctime)s] [%(filename)s:%(lineno)d] %(levelname)s %(message)s",
+            datefmt="%H:%M:%S",
+        )
```
MasterSkepticista, line 51: Thanks!

MasterSkepticista (approved):
I assume we tried re-using `setup_logger` within tests and we decided to override handlers.

LGTM if this is the optimal solution in your view

tests/end_to_end/utils/federation_helper.py
payalcha: We were using shutil.rmtree(path) but for docker some files seems to be created as root and its giving permission error.

noopurintel: That's right.

noopurintel:
> I assume we tried re-using `setup_logger` within tests and we decided to override handlers.
> 
> LGTM if this is the optimal solution in your view

Yes, we wanted to override the main/existing rich handler (`force_terminal=False`) with a new one (`force_terminal=True`) so that color codes are made available during GitHub workflow run display as well. 
Appending handlers would have resulted in duplicate display of content.

PR #1459, opened 2025-03-19
Optimize Aggregator Memory Usage

MasterSkepticista:
**Summary**
* Eliminates two leaking objects on the aggregator.
* Eliminates unwanted copies/allocations while maintaining functional correctness.
* Certain allocations have been made deterministic.
* Forces `np.float32` typecast to prevent `np.float64` allocations, which depending on input dtypes is default numpy behavior.
* Executes `SendLocalTaskResults` RPC in a queued manner. This does not affect the fact that multiple clients can still call RPCs as many threads are allotted on the server.

**Outcomes (assuming 16-collaborator histology experiment):**
* Average memory usage of aggregator down by about **40%**.
* Experiments finish in about half the time (yes, a lot more is at play in real world, but this outcome is more to say that aggregator executes calls faster (_ceteris paribus_)).

**Tests:** CI

**Sample run:**

* Blue: This PR (**+9MB/round** on steady state)
* Orange: `develop` tip at the time of run (**+54MB/round** on steady state)

image


rahulga1:
@payalcha can we run the experiment which you folks did earlier on this PR? we have a baseline already and this optimization will give us much better picture.

openfl/transport/grpc/aggregator_server.py
```diff
@@ -16,6 +17,17 @@
 logger = logging.getLogger(__name__)
 
 
+def synchronized(func):
```
rahulga1: I feel this could be part of a utility file.

MasterSkepticista:
> @payalcha can we run the experiment which you folks did earlier on this PR? we have a baseline already and this optimization will give us much better picture.

Added one plot from my local tests in the PR description, would be good to have one more independent test

rahulga1:
> @payalcha can we run the experiment which you folks did earlier on this PR? we have a baseline already and this optimization will give us much better picture.

As Karan has shared the plot now, I think its not required.

porteratzo (approved):
LGTM

payalcha:
As asked by @rahulga1 ran the comparison between Base code and this PR for histology for 25 rounds

image

 Seems significant improvement.

As confirmed by @MasterSkepticista There is no changes done by him for Collaborator.

openfl/component/aggregator/aggregator.py
```diff
@@ -973,82 +960,17 @@ def _prepare_trained(self, tensor_name, origin, round_number, report, agg_result
...
-        # First insert the aggregated model layer with the
-        # correct tensorkey
+        # The aggregated tensorkey tags should have the form of 'trained' or
+        # 'trained.lossy_decompressed' they need to be relabeled to 'aggregated' and reinserted.
+
+        # First insert the aggregated model layer with the correct tensorkey
         agg_tag_tk = TensorKey(tensor_name, origin, round_number + 1, report, ("aggregated",))
         self.tensor_db.cache_tensor({agg_tag_tk: agg_results})
 
-        # Create delta and save it in TensorDB
```
psfoley: This block should not be deleted entirely. If lossy compression is used and/or deltas are calculated between models, I expect this will lead to problems. I understand removing this results in significant memory savings. I would suggest making this entire block conditional for now to reduce memory for existing examples / SFAI.

psfoley (changes requested):
@MasterSkepticista these memory saving are significant - looking forward to bringing this into OpenFL. I have one comment around a breaking change that removes support for deltas / lossy compression. Once that is addressed my view is it's ready to merge.

openfl/component/aggregator/aggregator.py
MasterSkepticista: From the lines, it seemed like a one-way compress/decompress to make the tensor sparse. Wouldn't it be mathematically same regardless of flags enabled?

I did verify that the decompressed keys were not used elsewhere in the code. I will rewrite this block in the `TensorCodec` migration.

Reverted in https://github.com/securefederatedai/openfl/pull/1459/commits/21eedc29d7ad68be385b538a57cacde052ce3fbc

tanwarsh (approved):

openfl/transport/grpc/aggregator_server.py
MasterSkepticista: Migrated https://github.com/securefederatedai/openfl/pull/1459/commits/a5ff312e387d9eb86f3c1a3e7d8a2a51ec6009e7

PR #1469, opened 2025-03-21
Task callback enhancement / Bugfix

bsenthilr:
## Summary
This PR addresses Round number simplification and adjustments required for proper tensor key retrievals. 
When a given round completes aggregator builds a new set of Tensorkey for next round. The collaborator then retrieves these tensor keys on next round when they process the tasks for next round. This PR addresses this problem and streamlines the change on recovery path as well/

To support custom application extensions additional task begin & end callbacks are supported.

## Type of Change (Mandatory)
- Feature enhancement  
- Bug fix  
- Documentation update  
- Security improvement

## Testing
Regressed at application level and also CI pipeline end to end test level.

### DIFF
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -1139,9 +1139,9 @@ class Aggregator:
         self._end_of_round_check_done[self.round_number] = True
 
         # Save the latest model
-        logger.info("Saving round %s model...", self.round_number + 1)
-        self._save_model(self.round_number + 1, self.last_state_path)
         self.round_number += 1
+        logger.info("Saving round %s model...", self.round_number)
+        self._save_model(self.round_number, self.last_state_path)
 
         # End of round callbacks.
         # todo handle case when aggregator restarted before callback was successful
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -1135,9 +1135,6 @@ class Aggregator:
         for task_name in self.assigner.get_all_tasks_for_round(self.round_number):
             logs.update(self._compute_validation_related_task_metrics(task_name))
 
-        # End of round callbacks.
-        self.callbacks.on_round_end(self.round_number, logs)
-
         # Once all of the task results have been processed
         self._end_of_round_check_done[self.round_number] = True
 
@@ -1146,6 +1143,11 @@ class Aggregator:
         self._save_model(self.round_number, self.last_state_path)
 
         self.round_number += 1
+
+        # End of round callbacks.
+        # todo handle case when aggregator restarted before callback was successful
+        self.callbacks.on_round_end(self.round_number, logs)
+
         # resetting stragglers for task for a new round
         self.stragglers = []
         # resetting collaborators_done for next round
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -237,8 +237,8 @@ class Aggregator:
                 to_proto_tensor_dict, committed_round_number, self.compression_pipeline
             )
             # round number is the current round which is still in process
-            #  i.e. committed_round_number + 1
-            self.round_number = committed_round_number + 1
+            #  i.e. committed_round_number
+            self.round_number = committed_round_number
             logger.info(
                 "Recovery - loaded round number %s and best score %s",
                 self.round_number,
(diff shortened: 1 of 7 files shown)


MasterSkepticista (commented):
Does `TensorDB` access cover what you need for the callback?

bsenthilr:
> Does `TensorDB` access cover what you need for the callback?

lets sync up on it. looks like we infer keys from task runner based on tasks.

openfl/component/collaborator/collaborator.py
```diff
@@ -248,6 +248,9 @@ def do_task(self, task, round_number) -> dict:
             input_tensor_dict=input_tensor_dict,
             **kwargs,
         )
+
+        self.callbacks.on_task_end(round_number)
+
```
MasterSkepticista, line 257: Think it is neater to call this in the task loop.

```python
for task in tasks:
  self.callbacks.on_task_begin(round_number)
  metrics = self.do_task(task)
  self.callbacks.on_task_end(round_number)
```

It also gives confidence that any TensorDB transactions are done and dusted within the `do_task` scope

openfl/databases/tensor_db.py
```diff
@@ -165,6 +172,12 @@ def get_tensors_by_round_and_tags(self, fl_round: int, tags: tuple) -> dict:
         df = self.tensor_db[
             (self.tensor_db["round"] == fl_round) & (self.tensor_db["tags"] == tags)
         ]
+        if origin_filter is not None:
+            df = df[df["origin"].str.contains(origin_filter)]
+        if skip_optimizer_tensors:
+            df = df[~df["tensor_name"].str.startswith("__opt")]
+        if skip_report:
+            df = df[~df["report"]]
```
MasterSkepticista: I suggest applying these filters outside this function and retaining the API. In other words, one can fetch with round number and tags and apply filters on tensor keys later. 

Eventually, we even want to remove `get_tensors_by_round_and_tags` from tensor db with the reason of keeping it only for CRUD operations. Filtering gets complex very quickly, and it is better to keep it outside a database scope.

openfl/component/aggregator/aggregator.py
```diff
@@ -1135,17 +1135,18 @@ def _end_of_round_check(self):
...
 
         # Save the latest model
-        logger.info("Saving round %s model...", self.round_number)
-        self._save_model(self.round_number, self.last_state_path)
-
+        logger.info("Saving round %s model...", self.round_number + 1)
+        self._save_model(self.round_number + 1, self.last_state_path)
         self.round_number += 1
+
+        # End of round callbacks.
```
MasterSkepticista: Continuing our offline discussion, I think saving _after_ round number increment helps in two ways:
* It makes it explicit that a global model is saved at the beginning of new round.
* `+1` is not mistaken for zero-indexing offset.

Also a comment on the whole save logic is recommended, because this will be overlooked and has critical consequences on open/closed source

openfl/databases/tensor_db.py
bsenthilr: yes, lets discuss on this. If we can eliminate iteration to single time that would be even better. But doing it as user level makes sense to push out all complexity higher up and keep the core simple and strong.

openfl/databases/tensor_db.py
```diff
@@ -151,7 +151,40 @@ def get_tensor_from_cache(self, tensor_key: TensorKey) -> Optional[np.ndarray]:
...
+                origin=row["origin"],
+                round_number=row["round"],
+                report=row["report"],
+                tags=row["tags"],
+            )
+            tensor_dict[tensor_key] = np.array(row["nparray"])
+
+        return tensor_dict
+
+    def get_tensors_by_round_and_tags(
```
bsenthilr, line 183: use helper filter function

openfl/component/collaborator/collaborator.py
bsenthilr: current proposal is to wrap precisely around the task call. There was a timing issue and trying to recall. Lets start with current proposal and tune it in next iteration.

openfl/databases/tensor_db.py
bsenthilr: done

openfl/component/aggregator/aggregator.py
bsenthilr: fixed restart will not matter since the round number will be reinitialized from persistent storage.

psfoley (approved):

openfl/component/aggregator/aggregator.py
bsenthilr: Think there was a scenario where aggregator callback extension would try to access round number and assume the model was saved. Trying to recall..

bsenthilr: noticed a scenario where callbacks are still invoked for a completed task due to asynchronous aggregator. We can discuss on this offline addressed the PR and adjusted callback to handle it. We will need to rationalize exactly why this occurs and make adjustments as required.

PR #1476, opened 2025-03-24
Add `CollaboratorSerialiser` middleware for serialisation/deserialisation between collaborator and aggregator client

theakshaypant:
## Summary
Moving serialisation/deserialisation logic out of collaborator into a separate middleware. 

### Why?
By clearly separating concerns across layers, it aims to enhance code readability for contributors and improves debuggability in case of failures.
This middleware introduction is similar to "middleware chaining" paradigm followed widely in [Go](https://medium.com/@chrisgregory_83433/chaining-middleware-in-go-918cfbc5644d) where the same function is implemented across multiple "layers", each handling a distinct aspect of the functionality. In the context of this PR, this is done such that
- Layer 1 (`Collaborator`): Performs operations central to FL.
- Layer 2 (`CollaboratorSerialiser`): Performs serialisation/desrialisation of numpy array.
- Layer 3 (Transport): Interacts with the aggregator server to send/receive data. 

## Type of Change (Mandatory)
- Feature enhancement  
- Restructuring

## Description (Mandatory)
- The collaborator uses `TensorCodec` to convert tensors to protobuf format and vice versa in order to communicate with the aggregator.
- This conversion logic is moved out from the collaborator and it focuses solely on the core functionality while the conversion and client call is abstracted from the user.
- With the introduction of other transport implementations in the future, the functionality can be retained by passing the new (say REST) client to the middleware. 

## Testing
Manual and CI.
[Eden compression CI](https://github.com/theakshaypant/openfl/actions/runs/14307859854/job/40095684077)

## Additional Information
- Will be adding a serialisation middleware to the aggregator in a subsequent PR.


openfl/transport/serialiser/collaborator.py
MasterSkepticista, line 1: While I understand the motivation behind adding a new component, this adds significant LOC for what is effectively adding `deserialize/serialize` calls into each RPC.

IMO let us migrate those calls to grpc client to land a minimum API change version.

When we realize the need for a serializer to do more than what is being done today, I clearly see value in having a dedicated component that forwards across RPC methods like we do here.

theakshaypant:
The changes in tests mostly include moving tests between the now appropriate modules and/or renaming them.

openfl/pipelines/tensor_codec.py
```diff
@@ -146,98 +145,99 @@ def decompress(
...
+        """Convert named tensor to a numpy array.
 
         Args:
-            tensor_key: This is the tensor_key associated with the nparray.
-                Should have a tag of 'trained' or 'aggregated'
-            nparray: The nparray that corresponds to the tensorkey.
-            base_model_nparray: The base model tensor that will be subtracted
-                from the new weights.
+            named_tensor (protobuf): The tensor to convert to nparray.
+            collaborator_name (str): Name of teh collaborator for which the named tensor is to
```
rahulga1, line 153: typo `teh`

openfl/pipelines/tensor_codec.py
```diff
@@ -146,98 +145,99 @@ def decompress(
...
         """
-        tensor_name, origin, round_number, report, tags = tensor_key
-        if not np.isscalar(nparray):
-            assert nparray.shape == base_model_nparray.shape, (
-                f"Shape of updated layer ({nparray.shape}) is not equal to base "
-                f"layer shape of ({base_model_nparray.shape})"
-            )
-        assert "model" not in tags, (
-            "The tensorkey should be provided from the layer with new weights, not the base model"
+        # do the stuff we do now for decompression and frombuffer and stuff
```
rahulga1: This comment is not very clear, can you rewrite this?

openfl/pipelines/tensor_codec.py
```diff
@@ -146,98 +145,99 @@ def decompress(
...
+                tensor_key,
+                str.encode(json.dumps(nparray, cls=NumpyEncoder)),
+            )
+        else:
+            compressed_tensor_key, compressed_nparray, metadata = self.compress(
+                tensor_key, nparray, require_lossless=lossless
+            )
 
-        tensor_name, origin, round_number, report, tags = tensor_key
+        named_tensor = utils.construct_named_tensor(
```
rahulga1, line 220: wont it throw exception as metadata not defined for secagg?

theakshaypant: The "secagg" tag handling is a remnant of testing I was doing for aggregator side of similar changes. As for the collaborator serialisation/deserialisation, this does not affect the e2e flow. I will remove the code relevant to aggregator-side serialisation. 

Edit: Removed these lines in https://github.com/securefederatedai/openfl/commit/bdd5f433cddba7cf94229ce370533770eafff67c

openfl/pipelines/tensor_codec.py
theakshaypant: These comments were moved from `Collaborator.named_tensor_to_nparray`, since they have been moved to the relevant component, I will remove them. 

Edit: Removed these comments in https://github.com/securefederatedai/openfl/commit/bdd5f433cddba7cf94229ce370533770eafff67c

theakshaypant: Removed these comments in bdd5f433

openfl/pipelines/tensor_codec.py
theakshaypant: Resolved in https://github.com/securefederatedai/openfl/commit/bdd5f433cddba7cf94229ce370533770eafff67c

rahulga1 (approved):

openfl/pipelines/tensor_codec.py
```diff
@@ -146,98 +145,80 @@ def decompress(
...
+                require_lossless=lossless,
+            )
+
+        return decompressed_tensor_key, decompressed_nparray
 
-        return new_model_tensor_key, base_model_nparray + delta
+    def serialise(self, tensor_key, nparray, lossless=True):
+        """Construct the NamedTensor Protobuf.
 
-    def find_dependencies(self, tensor_key, send_model_deltas):
```
theakshaypant, line 212: This method has been moved to `Collaborator` after removing [these](https://github.com/securefederatedai/openfl/blob/develop/openfl/pipelines/tensor_codec.py#L235-L241) lines
```
                if self.compression_pipeline.is_lossy():
                    new_tags = ("aggregated", "delta", "lossy_compressed")
                else:
                    new_tags = ("aggregated", "delta", "compressed")
                tensor_key_dependencies.append(
                    TensorKey(tensor_name, origin, round_number, report, new_tags)
                )
```
As can be seen from the usage of the returned dependencies in the `Collaborator.get_data_for_tensorkey`, only the first tensor dependency is used, which is always
```
TensorKey(tensor_name, origin, round_number - 1, report, tags)
```
[Ref.](https://github.com/securefederatedai/openfl/blob/develop/openfl/component/collaborator/collaborator.py#L317)

openfl/component/collaborator/collaborator.py
```diff
@@ -69,12 +68,12 @@ def __init__(
         opt_treatment="RESET",
         device_assignment_policy="CPU_ONLY",
         use_delta_updates=False,
-        compression_pipeline=None,
         db_store_rounds=1,
         log_memory_usage=False,
         write_logs=False,
         callbacks: Optional[List] = [],
         secure_aggregation=False,
+        serialisation_middleware: CollaboratorSerialiser = None,
```
theakshaypant, line 76: We can also consider removing `client` completely from the collaborator since the middleware already has the client object. 
The client object is currently used by the secure aggregation callback which could potentially also work with the middleware given the changes proposed in #1518

tests/openfl/component/collaborator/test_collaborator.py
```diff
@@ -225,8 +175,73 @@ def test_get_data_for_tensorkey_dependencies(collaborator_mock, tensor_key):
...
+#     tensor_key = TensorKey(
+#         tensor_name, origin, round_number, report, ('model',)
+#     )
+#     tensor_key_dependencies = collaborator_mock._find_dependencies(tensor_key)
+
+#     assert len(tensor_key_dependencies) == 2
+#     tensor_key_dependency_0, tensor_key_dependency_1 = tensor_key_dependencies
+#     assert tensor_key_dependency_0.round_number == round_number - 1
+#     assert tensor_key_dependency_0.tags == tensor_key.tags
+#     assert tensor_key_dependency_1.tags == ('aggregated', 'delta', 'lossy_compressed')
```
theakshaypant, line 247: These tests are commented out and planned to be removed, refer https://github.com/securefederatedai/openfl/pull/1476#discussion_r2030982671

openfl/component/collaborator/collaborator.py
teoparvanov: On the contrary, I'm wondering whether the collaborator shouldn't just have a `client` object, while any "middlewares" are added as decorators to the `AggregatorGRPCClient` class. Those decorators could receive certain configurations via the FL plan (such as the compression algo). 

WDYT?

openfl/transport/serialiser/collaborator.py
```diff
@@ -0,0 +1,95 @@
...
+
+"""
+This file contains the CollaboratorSerialiser which is used as a middleware between the
+collaborator component and aggregator client.
+"""
+
+from openfl.pipelines import TensorCodec
+
+
+class CollaboratorSerialiser:
```
teoparvanov, line 12: This class indeed looks like a decorator of `AggregatorGRPCClient`, as it implements the same interface, but adds some orthogonal aspects like encoding/decoding via the tensor codec.

teoparvanov (commented):
Hi @theakshaypant , thanks for starting this effort.

To begin with, could you add a "why" section in the PR description? In addition to the obvious encapsulation benefits, what framework capabilities do you expect this refactoring to enable down the road? 

Additional comments based on the code itself:

theakshaypant:
> To begin with, could you add a "why" section in the PR description? In addition to the obvious encapsulation benefits, what framework capabilities do you expect this refactoring to enable down the road?

@teoparvanov Thanks for the comments! Updated the PR description. 
Adding a bit more descriptive answer here. This PR doesn't directly support any upcoming features/capabilities. However, by clearly separating concerns across layers, it enhances code readability for contributors and improves debuggability in case of failures. For instance, future developers won’t need to worry about low-level details like converting NumPy arrays to named tensors, as such operations are abstracted away and don’t impact the core functionality of the component. 
This is a small but hopefully meaningful step toward the larger improvements we’re aiming for in the repo.

openfl/component/collaborator/collaborator.py
theakshaypant: Ah, I see your point, I hadn't considered using decorators here!
I went with "middleware" approach as I feel it is easier to "chain" - multiple layers can be used and each of them can be easily added/removed without affecting any of the other layers. 
With the middleware approach, the only change required to switch between gRPC and REST would be in how the client is initialized. Similarly, in the future if we decide to make changes to the compression pipeline or such, we can look to make changes only the serialisation layer.

Gotta admit, I’m a bit biased toward this approach 😁 but totally open to other perspectives!

MasterSkepticista: I am leaning towards having entire approach encapsulated _within_ the client (@teoparvanov's approach is one of the ways to do that). 

The current implementation removes compression pipeline from the collaborator and it adds a new serializer object. From an abstraction standpoint it would be worth asking: _does a collaborator need to know about serialization_? Having it all in the client marks that separation very well.

I see the benefit in your approach of having a middleware when a different use case leverages it. Could you list any planned features that may benefit from this approach?

openfl/pipelines/tensor_codec.py
psfoley: I don't think that's quite right. When `use_delta_updates` is set to true, there are two dependencies: the prior version of the model, and the delta that should be applied (either compressed with a lossless or lossy pipeline) to that model to bring it to the current version. The 2nd tensor dependency (`tensor_dependency[1]`) is used later in the function that you reference [here](https://github.com/securefederatedai/openfl/blob/develop/openfl/component/collaborator/collaborator.py#L326-L331). If you're only seeing the first tensor dependency for every run, then what's likely happening is that `db_store_rounds` for the collaborator is set to 1 (it needs to be set to 2 in order to retain the prior version of the model locally). Before these lines are removed, you should add tests for `use_delta_updates=True` and use a lossy compression pipeline (such as [this one](https://github.com/securefederatedai/openfl/blob/develop/openfl/pipelines/skc_pipeline.py#L233)).

openfl/component/collaborator/collaborator.py
```diff
@@ -427,128 +419,43 @@ def send_task_results(self, tensor_dict, round_number, task_name) -> dict:
...
+                # Model should already be stored in the TensorDB
+                model_nparray = self.tensor_db.get_tensor_from_cache(
+                    TensorKey(tensor_name, origin, round_number, report, ("model",))
+                )
+
+                # The original model will not be present for the optimizer on the
+                # first round.
+                if model_nparray is not None:
+                    tensor, nparray = generate_delta(tensor, tensor_dict[tensor], model_nparray)
+                    # Second element of value indicates whether a lossless transofmation is
```
psfoley, line 433:
```suggestion
                    # Second element of value indicates whether a lossless transformation is
```

psfoley (commented):
I haven't gotten to the point of reviewing the architectural changes proposed, but I see some changes that appear to break the delta update functionality. I see some positive steps taken overall - reducing code within the collaborator is always a good thing. 2nd half of the review to come tomorrow.

openfl/component/collaborator/collaborator.py
theakshaypant: My original point was that, once we have sorted out the secure aggregation serialisation, we can fully replace the client object with this proposed class object. 
A key argument for keeping this logic separate from the client is the potential addition of a REST client. With the approach proposed in this PR, the serialiser would simply reference the enabled transport client, REST or gRPC.
An argument can be made for going with the decorator approach as mentioned by Teo but in my opinion, the middleware approach provides us more flexibility.

Issue #1475, opened 2025-03-24
Taskrunner workspace torch/mnist gets stuck post round 1.


tanwarsh:
**Describe the bug**
While running torch/mnist taskrunner workspace, experiments gets stuck post first round.

**To Reproduce**
1. Follow https://openfl.readthedocs.io/en/latest/tutorials/taskrunner.html
2. Instead of keras/mnist, run torch/mnist workspace. 
3. Observe the output.

**Expected behavior**
The experiment should complete with a trained model as output.

**Screenshots**

![Image](https://github.com/user-attachments/assets/ec27eedd-523e-4c54-a60a-d448d263f56a)

**Desktop (please complete the following information):**
 - NAME="Ubuntu"
 - VERSION="20.04.6 LTS (Focal Fossa)"
 - Python 3.11.11
 - torch==2.4.1
 - torchvision==0.19.1

teoparvanov:
Hi @tanwarsh, thanks for noticing this issue! From the logs, it looks like both collaborators have completed their tasks...

PS: the [torch/mnist job with Python 3.11](https://github.com/securefederatedai/openfl/actions/runs/14019986739/job/39250629754) seems to succeed in the PQ pipeline - could you compare what you would be doing differently?

noopurintel:
By default, when workspace is created, files `data.yaml` and `cols.yaml` already have two entries - for `collaborator1` and `collaborator2`.

If you are running the experiment manually, you need to remove these entries, IF the collaborators are going to be different than these.

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/769fbab3-c66c-40b0-9f03-a713058ce0e2" />

Refer - https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch/mnist/plan/data.yaml
https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch/mnist/plan/cols.yaml

Better solution is not to have any collaborator entry in these files. 

Please confirm if this was the case. 

Also, this is taken care of in the automation, thus the PR/PQ pipelines are successful.

payalcha:
With PR #1479, above issue will be resolved. Please check.

tanwarsh:
closing this as it is resolved by https://github.com/securefederatedai/openfl/pull/1479.

PR #1478, opened 2025-03-24
To streamline Collaborator yaml for torch/mnist with all other cols.yaml file

payalcha:
Fix - https://github.com/securefederatedai/openfl/issues/1475


openfl-workspace/torch/mnist/plan/cols.yaml
```diff
@@ -1,6 +1,4 @@
 # Copyright (C) 2020-2024 Intel Corporation
 # Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
 
-collaborators:
-- collaborator1
-- collaborator2
\ No newline at end of file
+collaborators:
```
teoparvanov, line 4: nit - please add a new line for a cleaner diff

teoparvanov (commented):
I also suggest updating the non-TLS section of the quickstart guide to mention `bob` and `charlie` instead of `collaborator1` and `collaborator2`:
https://openfl.readthedocs.io/en/latest/tutorials/taskrunner.html

Otherwise the non-TLS flavor of the tutorial would likely not work either. WDYT?

PR #1487, opened 2025-03-25
OpenFL v1.8 release notes

teoparvanov:
## Summary
Release notes for OpenFL 1.8.

## Type of Change (Mandatory)
- Documentation update  

## Description (Mandatory)
Summarizing the framework enhancements included in the 1.8 release.

## Testing
N/A

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,4 +1,20 @@
 # Releases
+## 1.8
+[Full Release Notes](https://github.com/securefederatedai/openfl/releases/tag/v1.8)
+
+### New Features
+- [**Secure Aggregation**](https://openfl.readthedocs.io/en/latest/about/features_index/secure_aggregation.html): A privacy-preserving aggregation algorithm for Federated Learning, based on secure multiparty computation (MPC) that helps protect intermediate model updates from introspection. OpenFL now supports Secure Aggregation via both Task Runner API and Workflow API.
+
+- [**OpenFL/Flower Interop**](https://github.com/securefederatedai/openfl/tree/develop/openfl-workspace/flower-app-pytorch): Showcasing how FL experiments defined using the Flower API can be run as OpenFL federations via TaskRunner API. This enables combining Flower's extensive library of FL algorithms with OpenFL's advanced security features, including trusted execution, secure communication, and explicit safeguards against data exfiltration.
+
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -1,4 +1,20 @@
...
+
+- [**Federated Evaluation**](https://openfl.readthedocs.io/en/latest/about/features_index/fed_eval.html): Enabling seamless switching from learning to evaluation in Task Runner API without redistributing the FL plan, which is particularly advantageous in large, geo-distributed federations. Furthermore, Federated Evaluation is now available via the [Workflow API](https://github.com/securefederatedai/openfl/blob/develop/openfl-tutorials/experimental/workflow/405_MNIST_FederatedEvaluation.ipynb).
+
+### Enhanced Developer Experience
+- **Removing Legacy APIs**: The Python Native API and the Interactive API have been removed, along with the accompanying examples and documentation.
+
+- **ML Frameworks Integration**: Upgraded PyTorch-based FL workspaces to 2.4.1, and provided additional Keras 3 back-ends (Jax and PyTorch).
+
+- **Enhanced Resilience**: OpenFL Task Runner API experiments can now recover from Collaborator and Aggregator restarts.
```
payalcha, line 17: Persistent Database for Task runner also got introduced in 1.8 which actually helps in making aggregator resilient. Do we need to add the same in Release notes.

teoparvanov: I think the SQLite DB in the Aggregator is an implementation detail that contributes to the ability to support restarts. Not sure whether we should go into that level of detail in the release notes. WDYT?

rahulga1:
We should be explicit about memory improvements, I think that is a significant improvement, wdyt?

teoparvanov:
> We should be explicit about memory improvements, I think that is a significant improvement, wdyt?

This is a good point, @rahulga1! However, I thought those changes didn't make it to the 1.8 branch?

rahulga1:
> > We should be explicit about memory improvements, I think that is a significant improvement, wdyt?
> 
> This is a good point, @rahulga1! However, I thought those changes didn't make it to the 1.8 branch?

you are right, I mixed up the branch while going through the merged PRs. We are good then.

rahulga1 (approved):

PR #1493, opened 2025-03-26
Federated Analytics and Histogram Taskrunner Workspace 

tanwarsh:
## Summary
The PR contains changes for Federated Analytics and Histogram Taskrunner Workspace
as discussed in https://github.com/securefederatedai/openfl/discussions/1385#discussioncomment-12548664

**without/minimum core components changes and dummy model.**

## Type of Change (Mandatory)
- New Feature.  

## Description (Mandatory)
The PR includes following changes:

- Used Dummy model to minimise core component changes till they are refactored.
- Collaborators will query data and return analysis results.
- The aggregator will aggregate the analysis results.
- The Base Taskrunner class will be used for FA.
- Taskrunner Workspace: Calculate a histogram (an estimate of the probability distribution of a continuous variable) on specific columns of the [Iris](https://huggingface.co/datasets/scikit-learn/iris)[ dat](https://huggingface.co/datasets/scikit-learn/iris)[aset](https://huggingface.co/datasets/scikit-learn/iris).

## Testing
Screenshot 2025-04-25 at 1 29 30 PM

Screenshot 2025-04-25 at 1 32 52 PM


openfl/protocols/utils.py
```diff
@@ -297,6 +300,26 @@ def dump_proto(model_proto, fpath):
...
+    """
+    with open(fpath, "w", newline="") as csvfile:
+        writer = csv.writer(csvfile)
+        for tensorkey, values in result.items():
+            if isinstance(values, list):
+                writer.writerow([tensorkey.tensor_name] + values)
+            elif isinstance(values, np.ndarray):
+                writer.writerow([tensorkey.tensor_name] + values.tolist())
+            else:
+                writer.writerow([tensorkey.tensor_name, values])
```
MasterSkepticista: Can we simply write to file as formatted JSON and avoid all these edge cases?

MasterSkepticista: Also - I don't think this function is related to 'protocols' in any way, so if the goal is only to write `result` to `fpath`, this can be done inline where this function is called, and not keep it here.

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,237 @@
...
+    def __init__(self, **kwargs):
+        """Initializes the FederatedAnalyticsTaskRunner instance.
+
+        Args:
+            **kwargs: Additional parameters to pass to the function
+        """
+        super().__init__(**kwargs)
+
+        # Dummy model initialization
+        self.model = self.build_model((28, 28, 1), 10, **kwargs)
```
MasterSkepticista: Why do we need a dummy model, that too for MNIST? Is this the same dummy/placeholder model @teoparvanov suggested?

openfl/protocols/utils.py
tanwarsh: Thank you for the review @MasterSkepticista. Moved this code inline and changed file to json instead of csv but I am afraid we cannot get rid of all the edge cases.

openfl/federated/task/runner_fa.py
tanwarsh: I'm not sure which dummy/placeholder model @teoparvanov suggested. From my understanding, the idea was to use a dummy model with dummy weights, which is what we are achieving here. I have removed the extra layers, and this model is solely used as a placeholder for core component code related to models.

tests/end_to_end/utils/federation_helper.py
```diff
@@ -718,7 +718,7 @@ def download_higgs_data(collaborators, local_bind_path):
         bool: True if successful, else False
     """
     common_download_for_higgs_and_flower(collaborators, local_bind_path)
-    
+
```
noopurintel: Thanks for this.

openfl/federated/__init__.py
```diff
@@ -13,7 +13,7 @@
 
 if util.find_spec("keras") is not None:
     from openfl.federated.data import KerasDataLoader
-    from openfl.federated.task import KerasTaskRunner
+    from openfl.federated.task import FederatedAnalyticsTaskRunner, KerasTaskRunner
```
kminhta: If this feature is at a stage where it needs a dummy model that is tied to a specific framework, I don't think we should introduce a dedicated task runner class outside of the workspace, yet.

payalcha:
@tanwarsh PR seems different from design https://github.com/securefederatedai/openfl/discussions/1385  in discussion. Can we update the design document as well or we can create a different design document in word or wiki.

openfl-workspace/federated_analytics/histogram/plan/plan.yaml
```diff
@@ -0,0 +1,52 @@
...
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+aggregator :
+  defaults : plan/defaults/aggregator.yaml
+  template : openfl.component.Aggregator
+  settings :
+    init_state_path : save/init.pbuf
+    best_state_path : save/best.pbuf
+    last_state_path : save/result.json
+    rounds_to_train : 1
```
payalcha: Comment mentioning that rounds_to_train can't be changed will be helpful.

openfl-workspace/federated_analytics/histogram/plan/cols.yaml
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2020-2025 Intel Corporation
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+collaborators:
```
payalcha, line 4: README.md document for this will be helpful as purpose of this is different from normal workspace.

openfl/federated/__init__.py
theakshaypant: Agree with this! The current changes would require the users to install keras for running a federated analytics workspace which does not even use it.

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+with catch_warnings():
+    simplefilter(action="ignore")
+    import keras
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class FederatedAnalyticsTaskRunner(TaskRunner):
```
teoparvanov: As suggested previously by @kta-intel and @theakshaypant , I suggest moving the `FederatedAnalyticsTaskRunner` under the `src` package in the workspace. At least until we figure out a cleaner way of integrating the FA use case with the core framework.

openfl/federated/data/loader_fa.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2020-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""DataLoader module."""
+
+
+class FederatedAnalyticsDataLoader:
```
teoparvanov: same here, let's move `FederatedAnalyticsDataLoader` under `src` in the example workspace, while the interfaces are still evolving

openfl-workspace/federated_analytics/histogram/src/histogram.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+"""Histogram module."""
+
+import numpy as np
+
+from openfl.interface.aggregation_functions.core import AggregationFunction
+
+
+class Histogram(AggregationFunction):
```
teoparvanov: Let's call this `AggregateHistogram`

openfl-workspace/federated_analytics/histogram/plan/plan.yaml
```diff
@@ -0,0 +1,52 @@
...
+  template : src.taskrunner.IrisHistogram
+
+network :
+  defaults : plan/defaults/network.yaml
+
+assigner :
+  template : openfl.component.RandomGroupedAssigner
+  settings :
+    task_groups  :
+      - name       : analysis
```
teoparvanov: nit - suggestion to rename analysis to analytics, to better align with the concept of Federated Analytics

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+        Args:
+            col_name (str): The column name for the analysis.
+            round_num (int): The current round number.
+            **kwargs: Additional parameters for analysis.
+
+        Returns:
+            dict: A dictionary of analysis results.
+        """
+        results = self.analysis_task(**kwargs)
+        tags = ("analysis",)
```
teoparvanov: Let's also rename the tag to `analytics`

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+                activation="relu",
+                input_shape=input_shape,
+            )
+        )
+
+        model.compile(loss="categorical_crossentropy", optimizer="adam", metrics=["accuracy"])
+
+        return model
+
+    def analysis(self, col_name, round_num, **kwargs):
```
teoparvanov: Let's rename the method to `query`

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+        results = self.analysis_task(**kwargs)
+        tags = ("analysis",)
+        origin = col_name
+        output_metric_dict = {
+            TensorKey(metric_name, origin, round_num, False, tags): metric_value
+            for metric_name, metric_value in results.items()
+        }
+        return output_metric_dict, output_metric_dict
+
+    def analysis_task(self, **kwargs):
```
teoparvanov: Same here - for symmetry, let's call this `query_task`

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+        model.compile(loss="categorical_crossentropy", optimizer="adam", metrics=["accuracy"])
+
+        return model
+
+    def analysis(self, col_name, round_num, **kwargs):
+        """
+        Return analytics result as tensors.
+
+        Args:
+            col_name (str): The column name for the analysis.
```
teoparvanov: By the way, where does the `col_name` parameter come from? I didn't see it in the FL plan task definition:
```yaml
tasks :
  analysis:
    function : analysis
    aggregation_type: 
      template: src.histogram.Histogram

    kwargs   :
      columns: ['sepal length (cm)', 'sepal width (cm)']
```

openfl/component/aggregator/aggregator.py
```diff
@@ -342,6 +344,17 @@ def _save_model(self, round_number, file_path):
         Returns:
             None
         """
+        analysis_result = self.tensor_db.get_tensors_by_round_and_tags(round_number, ("analysis",))
```
teoparvanov: -> analytics_result

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+                for i, weight in enumerate(copy.deepcopy(obj.variables))
+            }
+        else:
+            weight_name_index = 0
+            for layer in obj.layers:
+                if weight_name_index < len(weight_names) and len(layer.get_weights()) > 0:
+                    for weight in layer.get_weights():
+                        weights_dict[weight_names[weight_name_index] + suffix] = weight
+                        weight_name_index += 1
+        return weights_dict
```
teoparvanov: Instead of dynamically deriving the weights from a model object, couldn't we just set static ones?
Same goes for `_get_weights_names()` and `get_tensor_dict()`

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
...
+    def __init__(self, **kwargs):
+        """Initializes the FederatedAnalyticsTaskRunner instance.
+
+        Args:
+            **kwargs: Additional parameters to pass to the function
+        """
+        super().__init__(**kwargs)
+
+        # Dummy model initialization
+        self.model = self.build_dummy_model((28, 28, 1), **kwargs)
```
teoparvanov: If my suggestion from above works, wouldn't it even be possible to set `self.model = None` ?

teoparvanov (commented):
Thanks, @tanwarsh, I think it's looking better and more self-contained already! I especially like the fact that you've almost eliminated the changes to the core framework.

I have a couple of comments and suggestions from my first read-through.

PS: I see that the introduction of a dummy model introduces some complexity, for which I suggest an alternative to explore.

openfl/component/aggregator/aggregator.py
psfoley: These changes should be avoided as it bloats the scope of the save model function. There's a great opportunity to refactor this bespoke federated analytics code as a callback (i.e. [`on_round_end`](https://github.com/securefederatedai/openfl/blob/cb79a5c7da8321661bce8938322e37225dadf83f/openfl/callbacks/callback.py#L40-L44)).

openfl/federated/task/runner_fa.py
```diff
@@ -0,0 +1,220 @@
+# Copyright 2020-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""
+Base classes for Federated Analytics.
+
+You may copy this file as the starting point of your own keras model.
```
psfoley: These files should be deleted IMO. You can inherit directly from `TaskRunner` in the example. You can (and should) avoid the dependency on Keras. Part of the point of Federated Analytics is there is no machine learning model to be trained; you are just running a statistical query, etc. I expect you'll just need to return empty lists / dictionaries for `get_tensor_dict`, `set_tensor_dict`, `get_tensorkeys_for_function`, etc. to plug into the existing logic

psfoley (changes requested):
@tanwarsh Thanks for adding this. I like that you've added a Histogram aggregation function. I do think there's an opportunity to simplify some areas (remove dedicated analytics taskrunner, dataloader), and make use of callbacks for bespoke aggregator functionality. I see your intention of trying to limit what developers see in the example `src` directory, which is good, but if you remove the keras dependency I think you can cleanly trim this down to one file and inherit directly from the `TaskRunner` class.

openfl/component/aggregator/aggregator.py
teoparvanov: Great idea, @psfoley - this also bothered me during my review, but I wasn't sure what alternative to suggest to Shailesh... The callback approach would indeed do the job in this case!

openfl/federated/task/runner_fa.py
tanwarsh: removed keras mnist model.

openfl/federated/__init__.py
tanwarsh: Moved this code to src folder and removed keras mnist model.

openfl-workspace/federated_analytics/histogram/plan/plan.yaml
tanwarsh: resolved.

tanwarsh:
> @tanwarsh PR seems different from design #1385 in discussion. Can we update the design document as well or we can create a different design document in word or wiki.

I will update the design in https://github.com/securefederatedai/openfl/discussions/1385. I am also working on documentation. We have moved from work/wiki design doc in favour of github discussion.

openfl-workspace/federated_analytics/histogram/plan/cols.yaml
tanwarsh: I will update documentation instead of adding README.md file.

openfl/federated/task/runner_fa.py
tanwarsh: Moved FederatedAnalyticsTaskRunner to src folder in workspace.

openfl/federated/data/loader_fa.py
tanwarsh: removed FederatedAnalyticsDataLoader.

openfl-workspace/federated_analytics/histogram/src/histogram.py
tanwarsh: renamed to AggregateHistogram.

openfl-workspace/federated_analytics/histogram/plan/plan.yaml
tanwarsh: renamed analysis to analytics.

openfl/federated/task/runner_fa.py
tanwarsh: done.

openfl/federated/task/runner_fa.py
tanwarsh: I have renamed this to analytics_task instead of query_task. The idea is to not get confused between the query (the information requested which is histogram in this workspace) with query (a request for specific information or data from a database).

openfl/federated/task/runner_fa.py
tanwarsh: sorry for the confusion. col_name is actually collaborator name here. the doc string is updated.

openfl/component/aggregator/aggregator.py
tanwarsh: done.

openfl/federated/task/runner_fa.py
tanwarsh: removed model and added static tensors.

openfl/federated/task/runner_fa.py
tanwarsh: in order to make self.model = None work I had to change code mentioned below in openfl/protocols/utils.py.

```
if round_number is None:
# For Federated Analytics, we are using a placeholder model with no tensors
    round_number = 0
```

Alternative to this can be to handle this in _load_initial_tensors() in openfl/component/aggregator/aggregator.py where encounters error as round number is None.

let me know your thoughts on this.

openfl/component/aggregator/aggregator.py
tanwarsh: moved this code to on_round_end callback.

tanwarsh:
> @tanwarsh Thanks for adding this. I like that you've added a Histogram aggregation function. I do think there's an opportunity to simplify some areas (remove dedicated analytics taskrunner, dataloader), and make use of callbacks for bespoke aggregator functionality. I see your intention of trying to limit what developers see in the example `src` directory, which is good, but if you remove the keras dependency I think you can cleanly trim this down to one file and inherit directly from the `TaskRunner` class.

Thank you for the detailed review.

I will summarise the changes done below as per comments by @psfoley @teoparvanov @theakshaypant @kta-intel.

1. Removed dedicated FA dataloader to simply workspace.
2. Removed keras model and used static/empty tensors to bypass core related changes (please review the changes in openfl/protocols/utils.py  and in openfl/component/aggregator/aggregator.py in order to make this work).
3. Moved FederatedAnalyticsTaskRunner code to src folder in workspace. Although it is advised to remove this but IMO we should keep it as this separates the model related changes done to make FA work with openfl and these changes will be no longer required once we refactor the framework to enable it to work with both FL and FA. Once refactoring is done FederatedAnalyticsTaskRunner  will be left with code required for FA which can be moved back to framework instead of keeping it as part of workspace. This will help in simplify what user should focus on which is analytics and query logic in analytics_task in taskrunner.py and query function in dataloader respectively.
4. Moved the code from save model function to FederateAnalyticsCallback. (please review the changes done in aggregator to register the callback as we want the framework to work weather it is FL or FA. If we want to only use this callback for FA we will have to reintroduce mode/config in plan which was done earlier with core component changes, this will help in registering the callback only for FA.)

openfl/federated/task/runner_fa.py
tanwarsh: removed keras dependency and added code to return static/empty list for tensors.

openfl-workspace/federated_analytics/histogram/src/runner_fa.py
```diff
@@ -0,0 +1,150 @@
...
+        Get the model weights as a tensor dictionary.
+
+        Args:
+            with_opt_vars (bool): If we should include the optimizer's status.
+            suffix (str): Universally.
+
+        Returns:
+            model_weights (dict): The tensor dictionary.
+        """
+        return self._get_weights_dict(self.model, suffix)
```
teoparvanov: I suggest removing the "static" `_get_weights_dict()` method, as it doesn't seem to add much value. Also, here you could simply return a dummy tensor, to simulate the presence of a model:
```python
def get_tensor_dict(self, with_opt_vars):
        return {'dummy_tensor': np.float32(1)}
```

Thanks, @kta-intel for the hint!

openfl-workspace/federated_analytics/histogram/src/runner_fa.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+        self.required_tensorkeys_for_function["analytics"] = [
+            TensorKey(tensor_name, "GLOBAL", 0, False, ("model",))
+            for tensor_name in global_model_dict
+        ]
+        self.required_tensorkeys_for_function["analytics"] += [
+            TensorKey(tensor_name, "LOCAL", 0, False, ("model",))
+            for tensor_name in local_model_dict
+        ]
```
teoparvanov: Why is all of this needed, given that your analytics implementation does not rely on any model input tensors? I'd suggest trying with:
```python
    def initialize_tensorkeys_for_functions(self):
        pass
```
?

openfl-workspace/federated_analytics/histogram/src/taskrunner.py
```diff
@@ -0,0 +1,53 @@
...
+    def compute_hist(self, data, col_name):
+        """
+        Compute the histogram of a specified column in a DataFrame.
+        Args:
+            data (pandas.DataFrame): The DataFrame containing the data.
+            col_name (str): The name of the column for which to compute the histogram.
+        Returns:
+            numpy.ndarray: The computed histogram as an array.
+        """
+        histogram, _ = np.histogram(data[col_name], bins=np.linspace(2.0, 10.0, 10))
```
teoparvanov: Just for my understanding, could you explain the parameters passed here:
```python
bins=np.linspace(2.0, 10.0, 10)
```

openfl/callbacks/federate_analytics.py
```diff
@@ -0,0 +1,46 @@
...
+import logging
+
+import numpy as np
+
+from openfl.callbacks.callback import Callback
+
+logger = logging.getLogger(__name__)
+
+
+class FederateAnalyticsCallback(Callback):
```
teoparvanov: I'd rename this to the more self-explanatory `SaveAnalyticsResults`. WDYT?

openfl-workspace/federated_analytics/histogram/plan/plan.yaml
```diff
@@ -0,0 +1,52 @@
+# Copyright (C) 2020-2025 Intel Corporation
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+aggregator :
+  defaults : plan/defaults/aggregator.yaml
+  template : openfl.component.Aggregator
+  settings :
+    init_state_path : save/init.pbuf
+    best_state_path : save/best.pbuf
```
teoparvanov: Do we need to explicitly define the `init_state_path` and `best_state_path` here?

openfl/protocols/utils.py
```diff
@@ -44,6 +44,9 @@ def model_proto_to_bytes_and_metadata(model_proto):
                 f"Round numbers in model are inconsistent: {round_number} "
                 f"and {tensor_proto.round_number}"
             )
+    if round_number is None:
+        # For Federated Analytics, we are using a placeholder model with no tensors
+        round_number = 0
```
teoparvanov: This would likely not be needed if you implemented the suggestion above regarding `get_tensor_dict`...

openfl/component/aggregator/aggregator.py
```diff
@@ -1153,7 +1157,10 @@ def _end_of_round_check(self):
         self.callbacks.on_round_end(self.round_number, logs)
 
         # Save the latest model
-        if not self.assigner.is_task_group_evaluation():
+        analytics_result = self.tensor_db.get_tensors_by_round_and_tags(
+            self.round_number, ("analytics",)
+        )
+        if len(analytics_result) == 0 and not self.assigner.is_task_group_evaluation():
```
teoparvanov: you should also update the `logger.info` statement in the `else` case

openfl/component/aggregator/aggregator.py
```diff
@@ -1153,7 +1157,10 @@ def _end_of_round_check(self):
         self.callbacks.on_round_end(self.round_number, logs)
 
         # Save the latest model
-        if not self.assigner.is_task_group_evaluation():
+        analytics_result = self.tensor_db.get_tensors_by_round_and_tags(
+            self.round_number, ("analytics",)
+        )
```
teoparvanov: I suggest extracting this in a small private method called `self._has_analytics_results()`. In this case the condition below would become:
```python
if self._has_analytics_results() and not self.assigner.is_task_group_evaluation():
```

openfl-workspace/federated_analytics/histogram/plan/cols.yaml
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2020-2025 Intel Corporation
```
teoparvanov: nit - all newly added files related to FA should simply state `Copyright (C) 2025`. If they undergo significant changes next year, this would be extended to `Copyright (C) 2025-2026`.

teoparvanov (commented):
Looking good @tanwarsh, thanks for the hard work that has gone into this!
I have a couple of minor comments and suggestions:

openfl-workspace/federated_analytics/histogram/plan/cols.yaml
tanwarsh: done.

openfl/component/aggregator/aggregator.py
tanwarsh: done.

openfl/component/aggregator/aggregator.py
tanwarsh: done

openfl/protocols/utils.py
tanwarsh: removed.

openfl-workspace/federated_analytics/histogram/plan/plan.yaml
tanwarsh: removed.

openfl/callbacks/federate_analytics.py
tanwarsh: kept it as FederateAnalyticsCallback as it can be used to add other callbacks in future but for now renamed it SaveAnalyticsResults.

openfl-workspace/federated_analytics/histogram/src/runner_fa.py
tanwarsh: done

openfl-workspace/federated_analytics/histogram/src/runner_fa.py
tanwarsh: Thank you for the detailed review. I have implemented changes as per you suggestions.

openfl-workspace/federated_analytics/histogram/src/taskrunner.py
tanwarsh: sure, bins=np.linspace(2.0, 10.0, 10) specifies the bin edges. np.linspace(2.0, 10.0, 10) creates an array of 10 evenly spaced values between 2.0 and 10.0 (inclusive). These values serve as the boundaries for the histogram bins. the result you see in the description is the number of data points falling within each bin.

openfl/component/aggregator/aggregator.py
```diff
@@ -182,11 +182,15 @@ def __init__(
         self.model = None  # Initialize the model attribute to None
 
         # Callbacks
+        federate_analytics_callback = callbacks_module.SaveAnalyticsResults()
+        callbacks.append(federate_analytics_callback)
```
theakshaypant: Nit: It would be nice to add a comment explaining why no flag is required to run this callback, i.e., non analytics experiments would not be affected even with the callback.

openfl-workspace/federated_analytics/histogram/src/taskrunner.py
theakshaypant: Shouldn't these bins be part of the saved result (as headings) to make the histogram easier to understand?

openfl-workspace/federated_analytics/histogram/src/dataloader.py
```diff
@@ -0,0 +1,81 @@
...
+        Loads data from a CSV file.
+        This method reads the data from a CSV file located at './data/client.csv'
+        and returns it as a pandas DataFrame.
+        Returns:
+            pd.DataFrame: The data loaded from the CSV file.
+        """
+
+        return pd.read_csv('./data/client.csv')
+
+    def load_mnist_data_shard(self, shard_num, collaborator_count, **kwargs):
```
theakshaypant: Method name should be changed as MNIST database is not used here.

theakshaypant (commented):
Minor comments.

openfl-workspace/federated_analytics/histogram/src/dataloader.py
tanwarsh: done

openfl/component/aggregator/aggregator.py
tanwarsh: done

openfl/component/aggregator/aggregator.py
```diff
@@ -1181,6 +1191,18 @@ def _end_of_round_check(self):
...
+        """
+        Check if the current round has analytics results.
+
+        Returns:
+            bool: True if the current round has analytics results, False otherwise.
+        """
+        analytics_result = self.tensor_db.get_tensors_by_round_and_tags(
+            self.round_number, ("analytics",)
+        )
+        return len(analytics_result) != 0
```
teoparvanov: nit - `len(analytics_result) > 0`

teoparvanov (approved):
Thanks @tanwarsh, this is a great first iteration of the FA capability for OpenFL!

openfl/callbacks/federate_analytics.py
teoparvanov: Thanks, @tanwarsh - I think callbacks should be modular and should focus on a specific action. If more actions are needed, we can register additional callbacks.

openfl-workspace/federated_analytics/histogram/src/aggregatehistogram.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def call(self, local_tensors, *_) -> np.ndarray:
+        """
+        Aggregates a list of local tensors into a single histogram.
+        Args:
+            local_tensors (list): A list of objects, each containing a tensor attribute which is a numpy array.
+            *_: Additional arguments (unused).
+        Returns:
+            np.ndarray: The aggregated histogram as a numpy array. If the input list is empty, returns an empty numpy array.
+        """
```
MasterSkepticista: Not quite understood what this aggregation function does. Please elaborate from a data scientist's perspective what aggregating histograms is mathematically like. Example [optimizer](https://github.com/keras-team/keras/blob/9067d8060992924757e48492ee73dfa5df87c0ee/keras/src/optimizers/rmsprop.py#L7) description.

openfl-workspace/federated_analytics/histogram/src/aggregatehistogram.py
```diff
@@ -0,0 +1,31 @@
...
+
+        if not local_tensors:
+            return np.array([])
```
MasterSkepticista: If nothing is passed, it must be an error, not an empty result, isn't it?

openfl-workspace/federated_analytics/histogram/src/dataloader.py
```diff
@@ -0,0 +1,81 @@
...
+            raise ValueError("Columns parameter must be a list")
+        return self.data_shard[columns]
+
+    def get_feature_shape(self):
+        """Returns the shape of an example feature array.
+
+        Returns:
+            tuple: The shape of an example feature array.
+        """
+        pass
```
MasterSkepticista, line 82: If the intent was to tell the user it is not applicable, it is better to raise an error or let the docstring mention the same

openfl-workspace/federated_analytics/histogram/src/runner_fa.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self, **kwargs):
+        """Initializes the FederatedAnalyticsTaskRunner instance.
+
+        Args:
+            **kwargs: Additional parameters to pass to the function
+        """
+        super().__init__(**kwargs)
+
+        # Dummy model initialization
+        self.model = None
```
MasterSkepticista: A reader needs to know why it is a dummy model and how setting it to `None` helps. Bonus points if we can link an issue/discussion or even this PR to provide additional context.

openfl/callbacks/federate_analytics.py
```diff
@@ -0,0 +1,46 @@
...
+    SaveAnalyticsResults is a custom callback class for federated analytics.
+    This callback is triggered at the end of federated analytics to perform
+    analytics-related operations, such as retrieving tensors tagged with "analytics"
+    from a tensor database and saving them to a JSON file.
+
+    Methods:
+        __init__(self):Initializes the callback with an optional log directory.
+        end_of_round(round_number, metrics, context): process and save analytics data
+            after round completion.
+    """
```
MasterSkepticista: An entire new framework callback class isn't required. Since you need a single `on_round_end` trigger, pass a lambda function to [`LambdaCallback`](https://github.com/securefederatedai/openfl/blob/06d833543db969ad51e310af0548befcda917521/openfl/callbacks/lambda_callback.py#L6) instead.

MasterSkepticista (changes requested):

openfl-workspace/federated_analytics/histogram/src/aggregatehistogram.py
tanwarsh: added details in docstring.

openfl-workspace/federated_analytics/histogram/src/aggregatehistogram.py
tanwarsh: done

openfl-workspace/federated_analytics/histogram/src/dataloader.py
tanwarsh: done

openfl-workspace/federated_analytics/histogram/src/runner_fa.py
tanwarsh: added comment.

openfl/component/aggregator/aggregator.py
tanwarsh: code changed.

openfl/callbacks/federate_analytics.py
tanwarsh: Used [LambdaCallback](https://github.com/securefederatedai/openfl/blob/06d833543db969ad51e310af0548befcda917521/openfl/callbacks/lambda_callback.py#L6) instead of new callback.

openfl/component/aggregator/aggregator.py
```diff
@@ -1204,6 +1219,44 @@ def _end_of_round_check(self):
...
+        )
+        if len(analytics_result) > 0 and self.last_state_path:
+            with open(self.last_state_path, "w") as jsonfile:
+                analytics_result_json = {}
+                for tensorkey, values in analytics_result.items():
+                    if isinstance(values, np.ndarray):
+                        values = values.tolist()
+                    analytics_result_json[tensorkey.tensor_name] = values
+                json.dump(analytics_result_json, jsonfile, indent=4)
+            logger.info(f"Analytics result: {analytics_result_json}")
```
rahulga1: keep it debug log level.

teoparvanov (approved):

rahulga1 (approved):

openfl/component/aggregator/aggregator.py
tanwarsh: done.

PR #1500, opened 2025-03-27
Enable REST protocol between Aggregator and Collaborator

ishaileshpant:
- add new AggregatorClientInterface to allow switching b/w grpc and rest
- endhance existing AggregatorGRPCClient to start using AggregatorClientInterface
- added new transport package for rest with AggregatorRESTClient implementing AggregatorClientInterface
- added streaming api support with custom content-type
- added various connection flag for streaming request
- send additional header key "Sender" for better request logging at server side
- aligned Rest and gRPC client for most of the init params
- added AggregatorRESTServer and necesary changes in aggregator cli and federated/plan get_server method
- added transport_protocol settings in defaults/network.yaml, defaulted the same to 'grpc'
- reduced cyclomatic complexity of Rest Server
- fixed protobuf streaming issue for v1/task/results API
- added more detailed logging for task progression and metadata for each api calls
- pinned Flask version to latest stable 3.1.0

- addressing review comments - 13th-May
- added ping api and `collaborato` constructor hint for `AggregatorClientInterface`
- added send_message_to_server in client and AggregatorClientInterface, Rest Server is already at parity
- changed base uri for REST server to 'experimental/v1', adjusted the client and tests accordingly
- fixed issue related to mTLS in REST server/client
- disabled TLS 1.2 in both server/client
rebased 21st.May.1

- [x] Add AggregatorRESTServer
- [x] Remove collaborator cli support for protocol switch, this should be done via Plan/Aggregator
- [x] Add extensive logging for both REST server and client
- [x] Transparently create either REST or gRPC server and client binding when running a federation based on Plan config
- [x] Add Unit level test for existing modules
- [x] Add unit test for new modules that are added
- [x] Remove debug logs added for testing purposed
- [x] Manual testing of various config and status update in PR
- [x] Refactor verbose debug/info level log messages and metadata that gets logged
- [x] Known issue - mTLS communication is not working - when tls and client_auth both are enabled - Debug in progress
- [x] Validate that FedEval feature works as well with REST protocol
- [x] Validate with at-least two workspaces manually for REST : 
   - https://github.com/securefederatedai/openfl/tree/develop/openfl-workspace/torch/mnist
   - https://github.com/securefederatedai/openfl/tree/develop/openfl-workspace/torch/histology

## Summary
Enable rest protocol support for TaskRunner API

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This PR adds REST protocol support for TaskRunner API there by allowing running federation with REST protocol for seamless integration with managed/public API gateways
## Unit Test report
image

## Full Unit test report
image
image

### DIFF
--- a/openfl-workspace/workspace/plan/defaults/network.yaml
+++ b/openfl-workspace/workspace/plan/defaults/network.yaml
@@ -8,3 +8,4 @@ settings:
     require_client_auth        : True
     cert_folder                : cert
     enable_atomic_connections  : False
+    transport_protocol         : grpc
--- a/openfl/interface/aggregator.py
+++ b/openfl/interface/aggregator.py
@@ -92,8 +92,8 @@ def start_(plan, authorized_cols, task_group):
         logger.info(f"Setting aggregator to assign: {task_group} task_group")
 
     logger.info("🧿 Starting the Aggregator Service.")
-
-    parsed_plan.get_server().serve()
+    server = parsed_plan.get_server()
+    server.serve()
 
 
 @aggregator.command(name="generate-cert-request")
--- a/openfl/protocols/aggregator_client_interface.py
+++ b/openfl/protocols/aggregator_client_interface.py
@@ -1,0 +1,72 @@
+# Copyright 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""AggregatorClientInterface module."""
+
+from abc import ABC, abstractmethod
+from typing import Any, List, Tuple
+
+
+class AggregatorClientInterface(ABC):
+    @abstractmethod
+    def ping(self):
+        """
+        Ping the aggregator to check connectivity.
+        """
+        pass
+
+    @abstractmethod
+    def get_tasks(self) -> Tuple[List[Any], int, int, bool]:
+        """
+        Retrieves tasks for the given collaborator client.
+        Returns a tuple: (tasks, round_number, sleep_time, time_to_quit)
+        """
+        pass
+
+    @abstractmethod
+    def get_aggregated_tensor(
+        self,
+        tensor_name: str,
+        round_number: int,
+        report: bool,
+        tags: List[str],
+        require_lossless: bool,
+    ) -> Any:
+        """
+        Retrieves the aggregated tensor.
+        """
+        pass
+
+    @abstractmethod
+    def send_local_task_results(
+        self,
+        round_number: int,
+        task_name: str,
+        data_size: int,
+        named_tensors: List[Any],
+    ) -> Any:
+        """
+        Sends local task results.
+        Parameters:
+          collaborator_name: Name of the collaborator.
+          round_number: The current round.
+          task_name: Name of the task.
+          data_size: Size of the data.
+          named_tensors: A list of tensors (or named tensor objects).
+        Returns a SendLocalTaskResultsResponse.
+        """
+        pass
+
+    @abstractmethod
+    def send_message_to_server(self, openfl_message: Any, collaborator_name: str) -> Any:
+        """
+        Forwards a converted message from the local client to the OpenFL server and returns the
+        response.
+        Args:
+            openfl_message: The converted message to be sent to the OpenFL server (InteropMessage
+                proto).
+            collaborator_name: The name of the collaborator.
+        Returns:
+            The response from the OpenFL server (InteropMessage proto).
+        """
+        pass
--- a/openfl/transport/grpc/aggregator_client.py
+++ b/openfl/transport/grpc/aggregator_client.py
@@ -11,6 +11,7 @@ from typing import Optional, Tuple
 import grpc
 
 from openfl.protocols import aggregator_pb2, aggregator_pb2_grpc, utils
+from openfl.protocols.aggregator_client_interface import AggregatorClientInterface
 from openfl.transport.grpc.common import create_header, create_insecure_channel, create_tls_channel
 
 logger = logging.getLogger(__name__)
@@ -165,8 +166,10 @@ def _resend_data_on_reconnection(func):
     return wrapper
 
 
-class AggregatorGRPCClient:
+class AggregatorGRPCClient(AggregatorClientInterface):
     """Collaborator-side gRPC client that talks to the aggregator.
+
+    This class implements a gRPC client for communicating with an aggregator.
 
     Attributes:
         agg_addr (str): Aggregator address.
--- a/openfl/transport/rest/__init__.py
+++ b/openfl/transport/rest/__init__.py
@@ -1,0 +1,6 @@
+# Copyright 2020-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+from openfl.transport.rest.aggregator_client import AggregatorRESTClient
+from openfl.transport.rest.aggregator_server import AggregatorRESTServer
--- a/openfl/transport/rest/aggregator_client.py
+++ b/openfl/transport/rest/aggregator_client.py
@@ -1,0 +1,638 @@
+# Copyright 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""AggregatorRESTClient module."""
+
+# Standard library imports
+import logging
+import ssl
+import struct
+import time
+from typing import Any, List, Tuple
+
+# Third-party libraries
+import requests
+from google.protobuf import json_format
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+# Internal modules
+from openfl.protocols import aggregator_pb2, base_pb2
+from openfl.protocols.aggregator_client_interface import AggregatorClientInterface
+
+logger = logging.getLogger(__name__)
+
+
+class SecurityError(Exception):
+    """Security-related error."""
+
+    pass
+
+
+class AggregatorRESTClient(AggregatorClientInterface):
+    def __init__(
+        self,
+        agg_addr,
+        agg_port,
+        aggregator_uuid: str,
+        federation_uuid: str,
+        collaborator_name: str,
+        use_tls=True,
+        require_client_auth=True,
+        root_certificate=None,
+        certificate=None,
+        private_key=None,
+        single_col_cert_common_name=None,
+        refetch_server_cert_callback=None,
+        **kwargs,
+    ):
+        """
+        Initialize the AggregatorRESTClient with proper security settings.
+
+        Args:
+            agg_addr: Aggregator address
+            agg_port: Aggregator port
+            aggregator_uuid: UUID of the aggregator
+            federation_uuid: UUID of the federation
+            collaborator_name: Name of the collaborator
+            use_tls: Whether to use TLS
+            require_client_auth: Whether to require client authentication
+            root_certificate: Path to root certificate
+            certificate: Path to client certificate
+            private_key: Path to client private key
+            single_col_cert_common_name: Common name for single collaborator certificate
+            refetch_server_cert_callback: Callback to refetch server certificate
+        """
+        self.use_tls = use_tls
+        self.require_client_auth = require_client_auth
+        self.root_certificate = root_certificate
+        self.certificate = certificate
+        self.private_key = private_key
+        self.aggregator_uuid = aggregator_uuid
+        self.federation_uuid = federation_uuid
+        self.collaborator_name = collaborator_name
+        self.single_col_cert_common_name = single_col_cert_common_name
+        self.refetch_server_cert_callback = refetch_server_cert_callback
+
+        # Determine scheme and TLS verification
+        scheme = "https" if self.use_tls else "http"
+
+        # Configure certificate verification
+        self.cert_verification = self._configure_cert_verification(
+            self.use_tls, self.root_certificate
+        )
+
+        # Configure client certificates if required
+        if self.use_tls and self.require_client_auth:
+            if not self.certificate or not self.private_key:
+                raise ValueError(
+                    "Both certificate and private key are required for mTLS "
+                    "(client authentication). "
+                    "Please provide both certificate and private key paths."
+                )
+            self.cert = (self.certificate, self.private_key)
+        else:
+            self.cert = None
+
+        # Configure session with proper settings
+        self.session = requests.Session()
+
+        # Set default headers
+        self.session.headers.update(
+            {
+                "Connection": "keep-alive",
+                "Keep-Alive": "timeout=300",
+                "Accept": "application/json",
+                "X-Content-Type-Options": "nosniff",
+                "X-Frame-Options": "DENY",
+                "X-XSS-Protection": "1; mode=block",
+            }
+        )
+
+        # Configure timeouts with longer duration for large payloads
+        self.timeout = (30, 300)  # (connect timeout, read timeout) in seconds
+
+        # Configure retries with backoff
+        retry_strategy = Retry(
+            total=3,
+            backoff_factor=1,
+            status_forcelist=[408, 429, 500, 502, 503, 504],
+            allowed_methods=["GET", "POST"],
+            raise_on_status=True,
+        )
+
+        # Configure the adapter with the retry strategy
+        adapter = HTTPAdapter(
+            max_retries=retry_strategy, pool_connections=10, pool_maxsize=10, pool_block=False
+        )
+
+        # Mount the adapter for both HTTP and HTTPS
+        self.session.mount("http://", adapter)
+        self.session.mount("https://", adapter)
+
+        # Build the base URL
+        self.base_url = f"{scheme}://{agg_addr}:{agg_port}/experimental/v1"
+
+        # Log warning about experimental API
+        logger.warning(
+            "Initializing Aggregator REST Client (EXPERIMENTAL API - Not for production use)"
+        )
+
+        # Verify certificates if TLS is enabled
+        if self.use_tls:
+            try:
+                self._verify_certificates()
+            except Exception as e:
+                logger.error(f"Certificate verification failed: {e}")
+                raise
+
+    @classmethod
+    def _configure_cert_verification(
+        cls, use_tls: bool, root_certificate: str = None
+    ) -> bool | str:
+        """
+        Configure certificate verification settings for requests.
+
+        Args:
+            use_tls: Whether TLS is enabled
+            root_certificate: Optional path to root certificate file
+
+        Returns:
+            Union[bool, str]: Either True for system CA bundle, False for no verification,
+                             or path to root certificate file
+        """
+        if not use_tls:
+            return False
+
+        if root_certificate:
+            return root_certificate
+
+        return True  # Use system's default CA bundle
+
+    def _verify_certificates(self):
+        """Verify SSL certificates and configuration."""
+        import socket
+        import ssl
+
+        # Try to establish a test connection
+        try:
+            hostname = self.base_url.split("://")[1].split(":")[0]
+            port = int(self.base_url.split(":")[2].split("/")[0])
+
+            # Create SSL context with specific options
+            context = ssl.create_default_context()
+            context.verify_mode = ssl.CERT_REQUIRED
+            context.check_hostname = True
+
+            # Set secure cipher suites
+            context.set_ciphers("ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384")
+
+            # Disable older TLS versions
+            context.options |= (
+                ssl.OP_NO_TLSv1
+                | ssl.OP_NO_TLSv1_1
+                | ssl.OP_NO_TLSv1_2
+                | ssl.OP_NO_COMPRESSION
+                | ssl.OP_NO_TICKET
+            )
+
+            if self.root_certificate:
+                context.load_verify_locations(cafile=self.root_certificate)
+
+            if self.certificate and self.private_key:
+                context.load_cert_chain(certfile=self.certificate, keyfile=self.private_key)
+
+            # Use context managers for proper resource cleanup
+            with socket.create_connection((hostname, port)) as sock:
+                with context.wrap_socket(sock, server_hostname=hostname) as _:
+                    pass  # Connection successful if we get here
+
+        except ssl.SSLError as e:
+            if "CERTIFICATE_UNKNOWN" in str(e):
+                logger.error(
+                    "Certificate unknown error - this usually means the "
+                    "server's certificate is not trusted"
+                )
+                logger.error("Please verify that:")
+                logger.error(
+                    "1. The root certificate contains all necessary intermediate certificates"
+                )
+                logger.error("2. The server's certificate is properly signed by a trusted CA")
+                logger.error("3. The hostname matches the certificate's subject")
+            raise
+        except Exception as e:
+            logger.error(f"Connection verification failed: {e}")
+            raise
+
+    def _build_header(self) -> dict:
+        """Build and return a header dictionary with security headers."""
+        headers = {
+            "Receiver": self.aggregator_uuid,
+            "Federation-UUID": self.federation_uuid,
+            "Single-Col-Cert-CN": self.single_col_cert_common_name or "",
+            "X-Content-Type-Options": "nosniff",
+            "X-Frame-Options": "DENY",
+            "X-XSS-Protection": "1; mode=block",
+            "Sender": self.collaborator_name,
+        }
+        if self.use_tls:
+            headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains"
+        return headers
+
+    def _make_request(
+        self, method, url, data=None, params=None, headers=None, stream=False, timeout=None
+    ):
+        """Make a request with proper security settings."""
+        start_time = time.time()
+        try:
+            self._validate_url_scheme(url)
+            request_headers = self._prepare_headers(headers)
+            response = self._execute_request(
+                method, url, request_headers, data, params, stream, timeout
+            )
+            self._validate_response(response)
+            logger.debug(f"Request completed in {time.time() - start_time:.2f} seconds")
+            return response
+
+        except requests.exceptions.Timeout:
+            logger.error(f"Request timed out after {time.time() - start_time:.2f} seconds")
+            raise
+        except requests.exceptions.ConnectionError as e:
+            self._handle_connection_error(e)
+            raise
+        except requests.exceptions.RequestException as e:
+            self._handle_request_error(e)
+            raise
+
+    def _validate_url_scheme(self, url):
+        """Validate URL scheme matches TLS setting."""
+        if self.use_tls and not url.startswith("https://"):
+            raise ValueError("TLS required but URL is not HTTPS")
+        elif not url.startswith("http://") and not url.startswith("https://"):
+            raise ValueError("URL must use either HTTP or HTTPS scheme")
+
+    def _prepare_headers(self, headers):
+        """Prepare request headers with security settings."""
+        request_headers = self._build_header()
+        if headers:
+            request_headers.update(headers)
+        return request_headers
+
+    def _execute_request(self, method, url, headers, data, params, stream, timeout):
+        """Execute the HTTP request with retry logic."""
+        max_retries = 3
+        for attempt in range(max_retries):
+            try:
+                session = requests.Session()
+                if self.use_tls:
+                    # Extract hostname from URL for verification
+                    hostname = url.split("://")[1].split(":")[0].split("/")[0]
+
+                    # Create a custom SSL context for this request
+                    context = ssl.create_default_context(
+                        cafile=self.root_certificate if self.root_certificate else None
+                    )
+                    context.verify_mode = ssl.CERT_REQUIRED
+
+                    # Configure session with SSL context and hostname verification
+                    session.verify = self.cert_verification
+                    session.cert = self.cert
+
+                    # Configure adapter with proper SSL settings
+                    adapter = HTTPAdapter(
+                        pool_connections=1,
+                        pool_maxsize=1,
+                        max_retries=Retry(
+                            total=3,
+                            backoff_factor=1,
+                            status_forcelist=[408, 429, 500, 502, 503, 504],
+                            allowed_methods=["GET", "POST"],
+                        ),
+                    )
+                    session.mount("https://", adapter)
+
+                    # Build the complete headers with security information
+                    base_headers = self._build_header()
+                    if headers:
+                        # Merge user-provided headers with base headers
+                        base_headers.update(headers)
+                    headers = base_headers
+                    headers["Host"] = hostname
+
+                    # Add certificate info to request kwargs
+                    request_kwargs = {
+                        "method": method,
+                        "url": url,
+                        "headers": headers,
+                        "data": data,
+                        "params": params,
+                        "stream": stream,
+                        "verify": self.cert_verification,
+                        "timeout": timeout or self.timeout,
+                    }
+
+                    # Add client certificate if mTLS is enabled
+                    if self.require_client_auth:
+                        if not self.certificate or not self.private_key:
+                            raise ValueError(
+                                "Both certificate and private key are required for mTLS "
+                                "(client authentication). "
+                                "Please provide both certificate and private key paths."
+                            )
+                        # Use proper cert format
+                        request_kwargs["cert"] = (self.certificate, self.private_key)
+
+                    response = session.request(**request_kwargs)
+                else:
+                    # For non-TLS requests, still use the security headers
+                    base_headers = self._build_header()
+                    if headers:
+                        base_headers.update(headers)
+
+                    response = session.request(
+                        method=method,
+                        url=url,
+                        headers=base_headers,
+                        data=data,
+                        params=params,
+                        stream=stream,
+                        timeout=timeout or self.timeout,
+                    )
+                return response
+            except requests.exceptions.SSLError as e:
+                self._handle_ssl_error(e, attempt, max_retries)
+                if attempt == max_retries - 1:
+                    raise
+
+    def _handle_ssl_error(self, e, attempt, max_retries):
+        """Handle SSL errors with retry logic."""
+        if "CERTIFICATE_UNKNOWN" in str(e):
+            logger.error(
+                "Certificate unknown error - this usually means the "
+                "server's certificate is not trusted"
+            )
+            logger.error("Please verify that:")
+            logger.error("1. The root certificate contains all necessary intermediate certificates")
+            logger.error("2. The server's certificate is properly signed by a trusted CA")
+            logger.error("3. The hostname matches the certificate's subject")
+            if attempt < max_retries - 1 and self.refetch_server_cert_callback:
+                logger.debug("Attempting to refetch server certificate")
+                self.root_certificate = self.refetch_server_cert_callback()
+                # Update the cert_verification with the new root certificate
+                self.cert_verification = self._configure_cert_verification(
+                    self.use_tls, self.root_certificate
+                )
+                # Re-verify certificates
+                try:
+                    self._verify_certificates()
+                except Exception as verify_error:
+                    logger.error(f"Certificate re-verification failed: {verify_error}")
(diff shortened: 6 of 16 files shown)


openfl/protocols/aggregator_client_interface.py
```diff
@@ -0,0 +1,46 @@
+from abc import ABC, abstractmethod
+from typing import Any, List, Tuple
+
+
+class AggregatorClientInterface(ABC):
```
teoparvanov, line 10: what about `ping()`, which was recently introduced in [this PR](https://github.com/securefederatedai/openfl/pull/1516)?

openfl/transport/rest/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2020-2024 Intel Corporation
```
teoparvanov: nit - Copyright 2025

openfl/transport/grpc/aggregator_client.py
```diff
@@ -165,9 +166,12 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
-class AggregatorGRPCClient:
+class AggregatorGRPCClient(AggregatorClientInterface):
     """Collaborator-side gRPC client that talks to the aggregator.
 
+    This class implements a gRPC client for communicating with an aggregator
+    over a secure (TLS) connection.
```
teoparvanov: TLS is optional, so I'd just write here:
```
This class implements a gRPC client for communicating with an aggregator
```

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+        scheme = "https" if self.use_tls else "http"
+
+        # Configure certificate verification
+        if self.use_tls:
+            if self.root_certificate:
+                self.verify = self.root_certificate
+            else:
+                self.verify = True  # Use system default CA bundle
+        else:
+            self.verify = False
```
teoparvanov: Is `self.verify` a flag or an object?

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+            "Receiver": self.aggregator_uuid,
+            "Federation-UUID": self.federation_uuid,
+            "Single-Col-Cert-CN": self.single_col_cert_common_name or "",
+            "X-Content-Type-Options": "nosniff",
+            "X-Frame-Options": "DENY",
+            "X-XSS-Protection": "1; mode=block",
+        }
+        if self.use_tls:
+            headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains"
+        return {k: v for k, v in headers.items() if v is not None}
```
teoparvanov: why not simply `return headers` ?

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+        except requests.exceptions.RequestException as e:
+            self._handle_request_error(e)
+            raise
+
+    def _validate_url_scheme(self, url):
+        """Validate URL scheme matches TLS setting."""
+        if self.use_tls and not url.startswith("https://"):
+            raise ValueError("TLS required but URL is not HTTPS")
+        elif not self.use_tls and not url.startswith("http://"):
+            raise ValueError("Non-TLS mode but URL is not HTTP")
```
teoparvanov: ... but URL protocol is ...

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,615 @@
...
+        app = Flask(__name__)
+
+        # Set session and file age defaults
+        app.config["SEND_FILE_MAX_AGE_DEFAULT"] = 1800  # 30 minutes
+        app.config["PERMANENT_SESSION_LIFETIME"] = 1800  # 30 minutes
+
+        # Configure logging to be minimal
+        import logging
+
+        pass
```
teoparvanov: why is the `pass` needed here?

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,615 @@
...
+
+    def _setup_interop_client(self):
+        """Set up inter-federation connector client."""
+        try:
+            return self.aggregator.get_interop_client()
+        except AttributeError:
+            return None
+
+    @synchronized
+    def _validate_collaborator_identity(self, collaborator_id, federation_id):
```
teoparvanov: I'd suggest simply `_is_authorized()`

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,615 @@
...
+            except (TypeError, ValueError):
+                abort(400, "Invalid round number")
+            report = request.args.get("report", "").lower() == "true"
+            tags = request.args.getlist("tags")
+            require_lossless = request.args.get("require_lossless", "").lower() == "true"
+
+            # Get the tensor from aggregator
+            named_tensor = self.aggregator.get_aggregated_tensor(
+                tensor_name, round_number, report, tuple(tags), require_lossless
+            )
```
teoparvanov: Make sure to align with the security fix from [this PR](https://github.com/securefederatedai/openfl/pull/1570) when it gets merged (likely before yours).

tests/openfl/transport/rest/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (C) 2020-2023 Intel Corporation
```
teoparvanov, line 1: nit - Copyright (C) 2025

tests/openfl/transport/rest/conftest.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (C) 2020-2023 Intel Corporation
```
teoparvanov, line 1: nit - Copyright (C) 2025

openfl-workspace/workspace/plan/defaults/network.yaml
```diff
@@ -7,4 +7,5 @@ settings:
     client_reconnect_interval  : 5
     require_client_auth        : True
     cert_folder                : cert
-    enable_atomic_connections  : False
\ No newline at end of file
+    enable_atomic_connections  : False
+    transport_protocol         : grpc
```
teoparvanov, line 11: I suppose you've tested this E2E too, by setting the `transport_protocol` to `rest` for one of the existing workspaces, and running a full federation locally?

If so, please mention it in the PR description (in addition to the unit tests).

teoparvanov (commented):
Nice work, @ishaileshpant ! I've left a few minor comments, but overall it's looking good:

openfl-workspace/workspace/plan/defaults/network.yaml
ishaileshpant: Yes, will put details on status of these test, thanks for pointing it out

openfl/transport/rest/aggregator_client.py
ishaileshpant: Good point - will fix this

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+
+
+class SecurityError(Exception):
+    """Security-related error."""
+
+    pass
+
+
+class AggregatorRESTClient(AggregatorClientInterface):
+    def _verify_certificates(self):
```
theakshaypant: Nit: I think we should keep the constructor as the first method as part of the class definition.

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+        self.single_col_cert_common_name = single_col_cert_common_name
+        self.refetch_server_cert_callback = refetch_server_cert_callback
+
+        # Determine scheme and TLS verification
+        scheme = "https" if self.use_tls else "http"
+
+        # Configure certificate verification
+        if self.use_tls:
+            if self.root_certificate:
+                self.verify = self.root_certificate
```
theakshaypant: Is this the expected usage of verify?

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+
+        # Configure timeouts with longer duration for large payloads
+        self.timeout = (30, 300)  # (connect timeout, read timeout) in seconds
+
+        # Configure retries with backoff
+        retry_strategy = Retry(
+            total=3,
+            backoff_factor=1,
+            status_forcelist=[408, 429, 500, 502, 503, 504],
+            allowed_methods=["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE", "POST"],
```
theakshaypant: Are we using all these methods in the REST endpoints? 
If not, should we not restrict this list to the ones that are being used.

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+        return request_headers
+
+    def _execute_request(self, method, url, headers, data, params, stream, timeout):
+        """Execute the HTTP request with retry logic."""
+        max_retries = 3
+        for attempt in range(max_retries):
+            try:
+                session = requests.Session()
+                if self.use_tls:
+                    session.verify = self.verify
```
theakshaypant: Follow up on an earlier comment regarding the usage of verify. Is it a boolean or not?

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+                    session.cert = self.cert
+
+                response = session.request(
+                    method=method,
+                    url=url,
+                    data=data,
+                    headers=headers,
+                    timeout=timeout or self.timeout,
```
theakshaypant: Question: Why are we using both, timeout and self.timeout?

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,457 @@
...
+            if attempt < max_retries - 1:
+                logger.warning(f"SSL error (attempt {attempt + 1}/{max_retries}): {str(e)}")
+                time.sleep(1)
+            else:
+                raise
+
+    def _validate_response(self, response):
+        """Validate response headers and security settings."""
+        if self.use_tls:
+            if not response.url.startswith("https://"):
```
theakshaypant: Question: When would this condition be true?

openfl/transport/rest/aggregator_server.py
theakshaypant, line 1: Missing copyright header.

openfl/transport/rest/aggregator_server.py
theakshaypant: `pass` was added by the linter, I believe. Should be removed.

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,615 @@
...
+            self.ssl_context = None  # Explicitly set to None when TLS is disabled
+
+        # Set up Flask app
+        self.app = self._setup_flask_app()
+
+        # Set up interop client
+        self.interop_client = self._setup_interop_client()
+        self.use_connector = self.interop_client is not None
+
+    def __init__(
```
theakshaypant: Nit: Same as earlier, would suggest moving constructor earlier in the class.

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,615 @@
...
+                round_number=round_number,
+                tensor=named_tensor
+                if named_tensor is not None
+                else aggregator_pb2.NamedTensorProto(),
+            )
+
+            logger.debug(f"Tensor retrieval completed in {time.time() - start_time:.2f} seconds")
+            return jsonify(json_format.MessageToDict(response_proto))
+
+    def _setup_relay_route(self):
```
theakshaypant, line 800: I would question the addition of interop endpoint in the REST server while it is still not part of the core openfl gRPC server yet.

theakshaypant (commented):
Would also suggest modifying the type hint in the Collaborator constructor [here](https://github.com/securefederatedai/openfl/blob/develop/openfl/component/collaborator/collaborator.py#L66).

openfl/transport/rest/aggregator_server.py
teoparvanov: In fact, the `InteropRelay` is already implemented in the gRPC aggregator server (see [here](https://github.com/securefederatedai/openfl/blob/develop/openfl/transport/grpc/aggregator_server.py#L312)).

On the other hand, I agree with @theakshaypant that this functionality is early-stage, so I suggest leaving this endpoint as "not implemented" in the REST server (for now).

@kta-intel, WDYT?

kminhta: @teoparvanov, is the primary motivation for leading it as "no implemented" maintainability? If so, I think integrating it into the REST server at this stage is feasible. While `InteropRelay` _may_ be subject to change over time, its core function is to send a packaged Flower (or other framework) message across the network. Message conversion, message introspection, limiting info that'll be packaged/send, etc will be managed within the interop components, so that is where I am expecting most changes to happen. 

My main concern is whether this integration will affect the `interop_server` and `interop_client`, which are just a local gRPC server/client. If Flower and OpenFL can maintain gRPC communication via these interop  components, and the interop components can seamlessly interface with the OpenFL REST server/client, then implementing the `InteropRelay` in the REST server should be safe.

edit: `interop_server` does actually call the OpenFL grpc client directly to relay a message to the server [[ref]](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/interop_server.py#L89)]. This is currently contained in the flower workspace, but is intended to be pulled back into the core framework

openfl/protocols/aggregator_client_interface.py
kminhta, line 10: also what about, `send_message_to_server()`: [[ref]](https://github.com/securefederatedai/openfl/blob/develop/openfl/transport/grpc/aggregator_client.py#L425)

which is called by the `interop_server` to send a message to OpenFL server through OpenFL client [[ref]](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/flower-app-pytorch/src/grpc/connector/flower/interop_server.py#L89)

realize this may be tougher to get parity with since it directly calls the OpenFL grpc client, though

openfl/transport/rest/aggregator_client.py
ishaileshpant: `self.verify` is used as a flag or path passed to the requests library for SSL verification. It can be a boolean (True/False) or a string (path to a CA cert file). Essentially, it’s a direct mapping to requests.verify behavior. 

I agree it might be a bit unclear at first glance and will clarify this in the code comments. 

Specifically, if use_tls is True and we have a custom CA file, self.verify holds the path to that file (thus, an “object” in the sense of a filepath string). If use_tls is True without a custom CA, self.verify remains a boolean True (meaning use the default system CAs). And if TLS is disabled, we set it to False (no server cert verification). 

BTW This approach aligns with how `requests` expects these parameters https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification. 
I’ll make sure this logic is robust and clearly documented.

openfl/transport/rest/aggregator_server.py
ishaileshpant: Linter :(

openfl/transport/rest/aggregator_client.py
ishaileshpant: Yes - this is the intended usage afaiu from https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification

openfl/transport/rest/aggregator_server.py
ishaileshpant: This portion is still in discussion, will resolve when i feel we can either keep it for parity or completely remove it - AFAIU for now it should be OK to keep as long as it does

ishaileshpant:
> Nice work, @ishaileshpant ! I've left a few minor comments, but overall it's looking good:

Thank you for the positive feedback and for reviewing the PR thoroughly. I appreciate the guidance on the minor issues — I’ll address each of those

openfl/transport/rest/aggregator_server.py
ishaileshpant: agree with @kta-intel that we can keep it for parity as per gRPC server, however the full rest integration, @kta-intel pointed out can be achieved if the interop client / server also adapt to it afaiu, right @kta-intel ?

openfl/transport/rest/aggregator_client.py
ishaileshpant: The intention is to keep the usage of self.verify in line with 'requests' packages usage of verify as noted here https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification

openfl/transport/rest/aggregator_client.py
ishaileshpant: I added the dictionary comprehension to filter out any header values that might be None, but in our case that’s probably unnecessary having a second look at how server is adjusting to it. 
We can indeed simplify that to just return headers after conditionally adding the Strict-Transport-Security header.

ishaileshpant: the client-side filtering of None values is actually redundant because:
- The server will reject requests with missing required headers
- The server's security headers are always set with valid values
- The server validates header values before processing them

openfl/transport/rest/aggregator_client.py
ishaileshpant: this pattern is useful because different API calls need different timeouts:
- get_aggregated_tensor() uses extended_timeout = (30, 600) for longer operations
- send_local_task_results() uses (30, 60) for shorter operations
- Other operations fall back to the default self.timeout = (30, 300)

openfl/transport/rest/aggregator_client.py
ishaileshpant: Nice catch - the safeguards before ensure that we never reach this check / this condition never arises due to previous checks wherein request would fail before reaching this point if there was any attempt to downgrade to HTTP.

openfl/protocols/aggregator_client_interface.py
ishaileshpant: Done @teoparvanov

ishaileshpant:
> Would also suggest modifying the type hint in the Collaborator constructor [here](https://github.com/securefederatedai/openfl/blob/develop/openfl/component/collaborator/collaborator.py#L66).

Done @theakshaypant

openfl/transport/rest/aggregator_server.py
ishaileshpant: Done

openfl/transport/rest/aggregator_server.py
ishaileshpant: @psfoley as discussed offline, shall we retain this for grpc parity and implement the interop server/client changes in a separate PR? This way the rest server will still be at parity even if the feature is not fully baked in e2e

openfl/transport/rest/aggregator_server.py
ishaileshpant: I just validate that the above PR anyways have implemented this check inside the aggregators' get_aggregated_tensor method and thats the same called - for parity with grpc - even inside the REST server and hence the check is already in place AFAIU - wdyt @teoparvanov ?

openfl/transport/rest/aggregator_client.py
ishaileshpant: @teoparvanov tried to reduce the confusion and cognitive load by renaming this variable while still leveraging requests.verify internally

openfl/transport/rest/aggregator_server.py
theakshaypant: Reopening this.

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,631 @@
...
+        return app
+
+    def _setup_interop_client(self):
+        """Set up inter-federation connector client."""
+        try:
+            return self.aggregator.get_interop_client()
+        except AttributeError:
+            return None
+
+    def _is_authorized(self, collaborator_id, federation_id, cert_common_name=None):
```
theakshaypant, line 301: Can we move the timing attack protection sleep outside the conditionals since it is used for all of them (apart from the valid request)?

openfl/transport/rest/aggregator_server.py
ishaileshpant: This linter issue is becoming a pain :(

theakshaypant: Will need to remove it manually after running the linter or move the import to the top of the file. 😓

openfl/interface/aggregator.py
```diff
@@ -92,8 +92,9 @@ def start_(plan, authorized_cols, task_group):
         logger.info(f"Setting aggregator to assign: {task_group} task_group")
 
     logger.info("🧿 Starting the Aggregator Service.")
-
-    parsed_plan.get_server().serve()
+    # Instantiate either gRPC or REST server based on transport configuration
```
teoparvanov: I suggest removing the comment here, as it may become outdated (in case we introduce other protocols in the future f.e.).

openfl/protocols/aggregator_client_interface.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2020-2025 Intel Corporation
```
teoparvanov: nit - Copyright 2025, as this is a new file
(same comment for all other newly introduced files throughout the PR)

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,500 @@
...
+
+            if self.certificate and self.private_key:
+                context.load_cert_chain(certfile=self.certificate, keyfile=self.private_key)
+
+            # Create socket and wrap with SSL
+            sock = socket.create_connection((hostname, port))
+            ssl_sock = context.wrap_socket(sock, server_hostname=hostname)
+
+            ssl_sock.close()
+            sock.close()
```
teoparvanov: To ensure proper cleanup, I suggest acquiring/disposing those via context management:
```python
with socket.create_connection((hostname, port)) as sock:
    with context.wrap_socket(sock, server_hostname=hostname) as ssl_sock:
        pass
```

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,500 @@
...
+                f"Single collaborator certificate common name in response header does not match. "
+                f"Expected: {self.single_col_cert_common_name}, "
+                f"Actual: {header.get('singleColCertCommonName')}"
+            )
+
+            logger.info("Aggregator pong!")
+            return True
+        except Exception as e:
+            logger.warning(f"Aggregator ping failed: {str(e)}")
+            return False
```
teoparvanov: `ping()` either succeeds, or it fails with an exception - no return object is expected

openfl/transport/rest/aggregator_server.py
teoparvanov: As @kta-intel suggested in an earlier comment, we can keep the REST API just at the Collaborator/Aggregator level. The interop client and server are purely back-end entities (no app gateway in-between), so those can still communicate via gRPC (ref. https://github.com/securefederatedai/openfl/discussions/1440).

setup.py
```diff
@@ -94,6 +94,7 @@ def run(self):
         'tensorboardX',
         'protobuf>=4.21,<6.0.0',
         'grpcio>=1.56.2,<1.66.0',
+        'Flask==3.1.0',
```
teoparvanov, line 97: @rahulga1 this is an additional dependency for the core openfl framework, so it'll likely need to be mentioned as part of the security review for the 1.9 release.

openfl/transport/rest/aggregator_client.py
teoparvanov: So, actually this is due to the 3rd party `requests.verify` object that either accepts a boolean, or a certificate?
I'm wondering whether there's a way to better "isolate" this oddity to the boundary between `openfl` and `requests`, such as encapsulating it in a function that calculates the value for `self.verify`?

teoparvanov (approved):
Looking good, thanks @ishaileshpant !

openfl/transport/rest/aggregator_client.py
ishaileshpant: Yeah mainly to keep the shim simple and thin - can we attack this in subsequent PR if that works @teoparvanov ?

ishaileshpant: i agree that it helps to isolate and no locking to a particular library however, IMO, requests is something so intrinsic to API client integration that ROI of that abstraction might never materialize - wdyt?

ishaileshpant:
@psfoley / @teoparvanov is it fine if this refactor / task 'Retry mechanism config pullout, remove the hardcoding' be taken in a subsequent PR ?

openfl/transport/rest/aggregator_client.py
ishaileshpant: Done, thanks for catching this gap

openfl/transport/rest/aggregator_server.py
ishaileshpant: Done - simplified some checks too

payalcha:
@ishaileshpant 
https://github.com/securefederatedai/openfl/pull/1612/files#diff-13d501ef02760f7d8b8de7d82adc709762a4d2bb56aef8cdbbf8ddf06b4daafbR54 
Please uncomment the test once PR get merged. So that basic no-op connectivity will get tested with your PR.

openfl/transport/rest/aggregator_client.py
ishaileshpant: Done

openfl/transport/rest/aggregator_client.py
teoparvanov: I'm ok keeping it as-is if others don't mind either.

ishaileshpant: on second thought @teoparvanov have fixed it by moving it out a function to limit the sprawl of the dependency effect

teoparvanov: sounds good, at least now the "beast" is isolated in `_configure_cert_verification()` 😉

openfl/protocols/aggregator_client_interface.py
ishaileshpant: Thanks for catching this @kta-intel - Implemented

kminhta:
Thanks @ishaileshpant ! This is looking good. Btw, can you manually run https://github.com/ishaileshpant/openfl/actions/workflows/task_runner_flower_e2e.yml on your branch? I don't anticipate any major issues since it'll just run gRPC, but it'll be good to catch any unforeseen issues early

ishaileshpant:
> Thanks @ishaileshpant ! This is looking good. Btw, can you manually run https://github.com/ishaileshpant/openfl/actions/workflows/task_runner_flower_e2e.yml on your branch? I don't anticipate any major issues since it'll just run gRPC, but it'll be good to catch any unforeseen issues early

Thanks for the inputs @kta-intel - have just validated and the default workflow works https://github.com/ishaileshpant/openfl/actions/runs/15036703510/job/42259707667 better to validate this with protocol set to 'rest' - wdyt?

kminhta:
> Thanks for the inputs @kta-intel - have just validated and the default workflow works https://github.com/ishaileshpant/openfl/actions/runs/15036703510/job/42259707667 better to validate this with protocol set to 'rest' - wdyt?

thanks for checking, and yes I agree it would be good to validate with `rest`

ishaileshpant:
> > Thanks for the inputs @kta-intel - have just validated and the default workflow works https://github.com/ishaileshpant/openfl/actions/runs/15036703510/job/42259707667 better to validate this with protocol set to 'rest' - wdyt?
> 
> thanks for checking, and yes I agree it would be good to validate with `rest`

Validated with 'rest' protocol (w/o client auth) https://github.com/ishaileshpant/openfl/actions/runs/15099298504/job/42438127648

kminhta (approved):

payalcha:
RestAPI is not working as expected - **Collaborator** is not able to connect with **Aggregator**
```
[15:54:02] [plan.py:240] INFO Building `openfl.pipelines.NoCompressionPipeline` Module.
[15:54:02] [aggregator_client.py:130] WARNING Initializing Aggregator REST Client (EXPERIMENTAL API - Not for production use)
[15:54:02] [plan.py:240] INFO Building `openfl.component.Collaborator` Module.
[15:54:02] [collaborator.py:130] WARNING Argument `device_assignment_policy` is deprecated and will be removed in the future.
[15:54:02] [aggregator_client.py:295] ERROR Certificate unknown error - this usually means the server's certificate is not trusted
[15:54:02] [aggregator_client.py:299] ERROR Please verify that:
[15:54:02] [aggregator_client.py:300] ERROR 1. The root certificate contains all necessary intermediate certificates
[15:54:02] [aggregator_client.py:301] ERROR 2. The server's certificate is properly signed by a trusted CA
[15:54:02] [aggregator_client.py:302] ERROR 3. The hostname matches the certificate's subject
[15:54:02] [aggregator_client.py:332] ERROR Connection error: HTTPSConnectionPool(host='localhost', port=49467): Max retries exceeded with url: /experimental/v1/tasks?collaborator_id=collaborator1&federation_uuid=plan.yaml_1cb6b3e5 (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] sslv3 alert certificate unknown (_ssl.c:2578)')))
[15:54:02] [aggregator_client.py:334] ERROR Connection error details: HTTPSConnectionPool(host='localhost', port=49467): Max retries exceeded with url: /experimental/v1/tasks?collaborator_id=collaborator1&federation_uuid=plan.yaml_1cb6b3e5 (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] sslv3 alert certificate unknown (_ssl.c:2578)')))

```

ishaileshpant:
> RestAPI is not working as expected - **Collaborator** is not able to connect with **Aggregator**
> 
> ```
> [15:54:02] [plan.py:240] INFO Building `openfl.pipelines.NoCompressionPipeline` Module.
> [15:54:02] [aggregator_client.py:130] WARNING Initializing Aggregator REST Client (EXPERIMENTAL API - Not for production use)
> [15:54:02] [plan.py:240] INFO Building `openfl.component.Collaborator` Module.
> [15:54:02] [collaborator.py:130] WARNING Argument `device_assignment_policy` is deprecated and will be removed in the future.
> [15:54:02] [aggregator_client.py:295] ERROR Certificate unknown error - this usually means the server's certificate is not trusted
> [15:54:02] [aggregator_client.py:299] ERROR Please verify that:
> [15:54:02] [aggregator_client.py:300] ERROR 1. The root certificate contains all necessary intermediate certificates
> [15:54:02] [aggregator_client.py:301] ERROR 2. The server's certificate is properly signed by a trusted CA
> [15:54:02] [aggregator_client.py:302] ERROR 3. The hostname matches the certificate's subject
> [15:54:02] [aggregator_client.py:332] ERROR Connection error: HTTPSConnectionPool(host='localhost', port=49467): Max retries exceeded with url: /experimental/v1/tasks?collaborator_id=collaborator1&federation_uuid=plan.yaml_1cb6b3e5 (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] sslv3 alert certificate unknown (_ssl.c:2578)')))
> [15:54:02] [aggregator_client.py:334] ERROR Connection error details: HTTPSConnectionPool(host='localhost', port=49467): Max retries exceeded with url: /experimental/v1/tasks?collaborator_id=collaborator1&federation_uuid=plan.yaml_1cb6b3e5 (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_CERTIFICATE_UNKNOWN] sslv3 alert certificate unknown (_ssl.c:2578)')))
> ```

This has been fixed in latest push - the above error comes only in mTLS mode - TLS w/o client auth worked 

Now with the fix all the cases should be fine

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,634 @@
...
+            context = ssl.create_default_context()
+            context.verify_mode = ssl.CERT_REQUIRED
+            context.check_hostname = True
+
+            # Set secure cipher suites
+            context.set_ciphers("ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384")
+
+            # Disable older TLS versions
+            context.options |= (
+                ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1 | ssl.OP_NO_COMPRESSION | ssl.OP_NO_TICKET
```
psfoley: @ishaileshpant I believe TLS 1.2 should be disabled as well (check SDLe)

openfl/transport/rest/aggregator_server.py
```diff
@@ -0,0 +1,886 @@
...
+        """Set up SSL context for TLS/mTLS."""
+        ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+
+        # Set secure cipher suites
+        ssl_context.set_ciphers("ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384")
+
+        # Disable older TLS versions and set security options
+        ssl_context.options |= (
+            ssl.OP_NO_TLSv1
+            | ssl.OP_NO_TLSv1_1
```
psfoley, line 96: Same comment here about TLS 1.2

openfl/transport/rest/aggregator_client.py
```diff
@@ -0,0 +1,634 @@
...
+        data_bytes = data_stream.SerializeToString()
+        end_bytes = end_stream.SerializeToString()
+
+        # Create length-prefixed stream format
+        stream_data = (
+            struct.pack(">I", len(data_bytes))  # Length prefix for first message
+            + data_bytes  # First message
+            + struct.pack(">I", len(end_bytes))  # Length prefix for second message
+            + end_bytes  # Second message (empty message signals end)
+        )
```
psfoley, line 532: Is it necessary to encode the data stream this way? If so that's fine. I'm asking because it would be nice to have more symmetry with the gRPC implementation.

theakshaypant (approved):

psfoley (changes requested):
This is looking good, @ishaileshpant. I've tested this locally with mTLS and I think this is in pretty good shape overall. I have a few minor comments from my review. Two other things I would request before merge:

- **A REST E2E test for CI**. This is a big enough feature I want to be confident this is in a continuous working state.
- **Basic documentation**. There should either be a minor page that explains how to toggle between gRPC / REST in the plan, or a note of this in the real world federation page (related to proxies that don't support gRPC).

After this is merged for OpenFL 1.9 I see an opportunity for:
- More function / constant reuse between gRPC and REST implementations. I called out the possibility of more consistency between the gRPC / REST datastream messages, but other areas we could streamline are disallowed TLS versions. This feels like it should be defined once somewhere else and imported. 
- Additional tests for TLS / mTLS plan configurations. You are pretty well covered for functional tests for mTLS, but it would be good to test around just TLS as well (no client auth)

ishaileshpant:
> * **A REST E2E test for CI**. This is a big enough feature I want to be confident this is in a continuous working state.

We already have this in place as @payalcha has pointed out - we can uncomment this post the PR is merged

ishaileshpant:
> After this is merged for OpenFL 1.9 I see an opportunity for:

Totally agree, already have some ideas on doing this post the PR merge

openfl/transport/rest/aggregator_client.py
ishaileshpant: Done

openfl/transport/rest/aggregator_server.py
ishaileshpant: Done

openfl/transport/rest/aggregator_client.py
ishaileshpant: The current implementation is more like a batched RPC call with a custom binary protocol format rather than true streaming and i kept it this way to support both HTTP/1.1 and enhance for HTTP/2 true streaming later on - if you agree then post this iteration we should look at chunked transfer to better support both http/1.1 and eventually http/2 by enhancing when needs arises - wdyt?

PR #1514, opened 2025-04-03
Task Runner E2E: Simplify the components' `start` and `stop` processes

noopurintel:
## Summary
Simplify the `start` and `stop` processes of aggregator and collaborators.

**IMP:** Earlier we were using `shell=True` while running these components in the background. That was resulting in 2 processes running for every component. This has caused problem in resiliency once in a while when the process ids are to be fetched and killed before restarting a component. This PR fixes the same.

## Type of Change (Mandatory)
- E2E resiliency logic improvement

## Description (Mandatory)
We have experienced in the past that if an aggregator process is running despite of being stopped/killed due to any reason and another instance starts, collaborators are not able to connect to either of them. And there won't be any error log in aggregator as well. Also, due to 2 processes being associated with every component, it was vulnerable.

1. Modified the `start()` functions to ensure single process is running once the component starts.
2. Added `self.start_process` to aggregator and collaborators models and assigning the process when these components start in 1st step.
3. Added functions `kill_process()` to aggregator and collaborators.
4. Instead of using `ps -ef`, `grep` to fetch pids (which has caused problems off late), using `psutil` now; also added the logic to newly added function `get_pids_for_active_command()` for better modularity and reusability.
5. Added function `kill_processes()` which calls `get_pids_for_active_command()` to fetch pids and kills them in loop. 
6. Modified `run_command_background()` to include env variable. Helpful for component start functionality where we explicitly pass `LOG_FILE=<log_file>` in the environment.

## Testing
Resiliency test is present in PR pipeline itself.
PQ pipeline with my branch - https://github.com/noopurintel/openfl/actions/runs/14261005918/job/39974882078

PR pipeline with pytest 8.3.5 - https://github.com/securefederatedai/openfl/actions/runs/14260650904
Dockerized workspace E2E - https://github.com/noopurintel/openfl/actions/runs/14261814821


payalcha (approved):

teoparvanov:
@noopurintel , please provide a bit more context in the description. Is this a fix for the intermittent PQ test failures?

noopurintel:
@teoparvanov  - The intermittent PQ failure - due to Resiliency test failure shows no error logs in aggregator and collaborators keep trying to connect to it. So we can't pinpoint the exact reason right now.
However, during our local testing, we figured out that if aggregator process is already running and a new one starts (due to any reason)... there will be 2 aggregator processes and collaborators will start waiting for the connection with no error on aggregator screen.
This PR is a proactive measure to fix such situation.

teoparvanov (approved):

noopurintel:
With pytest 8.3.5 (PR is already available from dependabot), resiliency isn't working properly in my local machine. Moving the PR to draft stage to debug that. So that once pytest version is bumped up, we don't have to worry about it.

tests/end_to_end/utils/federation_helper.py
```diff
@@ -356,20 +357,13 @@ def _verify_completion_for_participant(
...
-        # Find the process ID
-        pids = []
-        for line in os.popen(f"ps ax | grep '{get_process_id}' | grep -v grep"):
-            fields = line.split()
-            pids.append(fields[0])
-
-        if not pids:
+        # Process poll None means process is still running
+        # If it is 0, it means process is completed
+        if participant.start_process.poll() is None:
```
payalcha: Can you please reverse this if/else
generally we give true condition
 if participant.start_process.poll():
else

tests/end_to_end/utils/federation_helper.py
```diff
@@ -17,6 +17,7 @@
 import tests.end_to_end.utils.db_helper as db_helper
 import tests.end_to_end.utils.docker_helper as dh
 import tests.end_to_end.utils.exceptions as ex
+import tests.end_to_end.utils.interruption_helper as int_helper
```
payalcha: Instead of int_helper intr_helper will make it more readable.

tests/end_to_end/utils/federation_helper.py
noopurintel: Addressed.

tests/end_to_end/utils/federation_helper.py
noopurintel: Done.

PR #1517, opened 2025-04-04
Change to add time taken for each round in metrics

payalcha:
## Summary
Adding time taken by each round in metrics log

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This change is to document time taken by each round in second. This information will only be gather if we keep write_metrics True in plan.yaml. 
It can help to analyse if number of rounds changing the duration for each round. Also in aggregator side it can help to understand how increase in collaborators is impacting each round duration 

## Testing
Simple PR pass.

### DIFF
--- a/openfl/callbacks/metric_writer.py
+++ b/openfl/callbacks/metric_writer.py
@@ -51,16 +51,19 @@ class MetricWriter(Callback):
             logs: A key-value pair of scalar metrics.
         """
         logs = logs or {}
-        logs["round_number"] = round_num
-        # time taken for the round to complete in seconds
-        logs["time_taken"] = time.monotonic() - self._round_start_time
-        logger.info(f"Round {round_num}: Metrics: {logs}")
+        elapsed_seconds = time.monotonic() - self._round_start_time
+        metrics = {
+            "round_number": round_num,
+            "elapsed_seconds": elapsed_seconds,
+            **logs,
+        }
+        logger.info(f"Round {round_num}: Metrics: {metrics}")
 
-        self._log_file_handle.write(json.dumps(logs) + "\n")
+        self._log_file_handle.write(json.dumps(metrics) + "\n")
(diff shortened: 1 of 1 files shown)


openfl/callbacks/metric_writer.py
```diff
@@ -49,6 +51,9 @@ def on_round_end(self, round_num: int, logs=None):
             logs: A key-value pair of scalar metrics.
         """
         logs = logs or {}
+        logs["round_number"] = round_num
+        # time taken for the round to complete in seconds
```
MasterSkepticista: Comment not really required; self-explanatory...

openfl/callbacks/metric_writer.py
```diff
@@ -49,6 +51,9 @@ def on_round_end(self, round_num: int, logs=None):
...
+        logs["time_taken"] = time.monotonic() - self._round_start_time
```
MasterSkepticista: `elapsed_seconds` is more descriptive?

openfl/callbacks/metric_writer.py
```diff
@@ -49,6 +51,9 @@ def on_round_end(self, round_num: int, logs=None):
...
         logger.info(f"Round {round_num}: Metrics: {logs}")
 
         self._log_file_handle.write(json.dumps(logs) + "\n")
```
MasterSkepticista: Suggested:

```python
metrics = {"round_number": round_number, "elapsed_seconds": dt, **logs}
self._log_file_handle.write(json.dumps(metrics) + "\n")
```
This ensure `logs` is not modified. Dicts are mutable in python, and since logs are provided by the _caller_, we do not want to modify it (at least not until we see good use of it outside)

kminhta (approved):

theakshaypant (approved):

PR #1518, opened 2025-04-04
Support generalized dtype `np.ndarray` transformers

MasterSkepticista:
### Status quo
Today, only `np.float32` arrays are transported over gRPC. There is a need for supporting wider range of (still numpy array) dtypes e.g. key shares in secure aggregation are strings.

There are two other issues:
* Use of other numerics (`np.intX`, `np.floatX`) leads to silent typecasts to `np.float32`. User is not aware of this and also may lead to loss of precision.
* Other dtypes fail with undefined behavior and/or garbage output.

### Proposal
This PR proposes a generalized lossless `NumpyArrayToBytes` transformer which supports all dtypes that `numpy` supports. A new field in the tensor metadata is inserted: `string dtype`.

### Impact
Secure aggregation currently creates a dependency on the component to serialize keys as json-encoded strings. It is not possible to separate `TensorCodec` from the `Aggregator` or `Collaborator` component without an informed way of sending arbitrary dtype arrays over communication channel.

Changes tested, ready for review (modulo any tests that failed due to tip moving ahead).


openfl/pipelines/tensor_codec.py
MasterSkepticista, line 1: Note for reviewer(s): Changes in this file are unrelated to the PR. Variables are renamed for readability and python GC action

payalcha:
All tests are failing.  Attached participant logs for reference.
[collaborator1.log](https://github.com/user-attachments/files/19604410/collaborator1.log)
[aggregator.log](https://github.com/user-attachments/files/19604420/aggregator.log)
[collaborator2.log](https://github.com/user-attachments/files/19604423/collaborator2.log)

theakshaypant (commented):
One of the reasons for the CI failing seems to be the absence of "dtype" in metadata. I believe we also need to modify places where metadata dictionary is generated from proto as it does not take `proto.dtype` into account.
Specifically [these](https://github.com/securefederatedai/openfl/blob/develop/openfl/protocols/utils.py#L32-L39) lines need to be changed such that
```python
        metadata_dict[tensor_proto.name] = [
            {
                "int_to_float": proto.int_to_float,
                "int_list": proto.int_list,
                "bool_list": proto.bool_list,
                "dtype": proto.dtype,
            }
            for proto in tensor_proto.transformer_metadata
        ]
```
Another such instance can be found [here](https://github.com/securefederatedai/openfl/blob/develop/openfl/component/aggregator/aggregator.py#L866-L873).

PR #1527, opened 2025-04-11
No-op workspace for TaskRunner API

teoparvanov:
## Summary
This "no-op" workspace template serves as a tool for testing the PKI setup and connectivity within a federation.

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This workspace template can be used in conjunction with the [recently introduced](https://github.com/securefederatedai/openfl/pull/1516) `fx collaborator ping` command.

It can be particularly useful in the initial phases of establishing a federation, as it allows for the separation of the infrastructure setup from the actual FL plan and task definitions.

Moreover, a "no-op" workspace offers the benefit of requiring minimal code review, making it easy for collaborators to agree to run it locally. Additionally:
* it does not access any data;
* it minimizes compute resource utilization for initial tests;
* only the `openfl` package is required (no 3rd party ML frameworks need to be installed).

## Testing
1. Configured a local federation with two collaborators (`bob` and `charlie`).
    * **NB!** There's no need to specify a data path when initializing the collaborators - simply go `fx collaborator create -n bob` and `fx collaborator create -n charlie`
3. Started the aggregator process.
4. Ran `fx collaborator ping -n bob & fx collaborator ping -n charlie` to check connectivity and TLS certificate validity.
5. Ran `fx collaborator start -n bob & fx collaborator start -n charlie` to run the federation E2E (0 rounds, with no tasks executed).
6. PKI and connectivity successfully tested without executing any ML tasks!


openfl/protocols/utils.py
```diff
@@ -44,6 +44,9 @@ def model_proto_to_bytes_and_metadata(model_proto):
                 f"Round numbers in model are inconsistent: {round_number} "
                 f"and {tensor_proto.round_number}"
             )
+
+    if round_number is None:
+        round_number = 0
```
teoparvanov: Can this code change be avoided somehow?

Alternatively, we could accept it as a corner case fix for the "no-op" scenario, given that there doesn't seem to be any regression related to it.

sarthakpati:
This is an awesome idea! Would be really useful to have this to test network connectivity between collaborators. 😄 👍🏽

kminhta (commented):
Overall LGTM @teoparvanov ! Added my two cents to your question

openfl/protocols/utils.py
kminhta: The only way that comes to mind that doesn't require any framework-level changes would be to pass a tensor dict with  `np.float32` (or `np.integer`, [see](https://github.com/securefederatedai/openfl/blob/develop/openfl/utilities/split.py#L63)) when `get_tensor_dict()` is called from the taskrunner

```python
def get_tensor_dict(self, with_opt_vars):
        return {'dummy_tensor': np.float32(1)}
```

You'd have to import numpy, but that's probably fine since numpy is installed with openfl anyway

Deconstructing of the model proto pulls the round number from the [tensor proto directly](https://github.com/securefederatedai/openfl/blob/develop/openfl/protocols/utils.py#L41), so when you pass a completely empty tensor, the round number never gets stored.


Otherwise, the solution you provided is likely the quickest and simplest, though I think a better long term solution would be to turn off the automatic creation and loading of `init.pbuf` since there are workloads that don't use it (i.e. interoperability, federated evaluation, federated analytics, etc.)

teoparvanov: Thanks, @kta-intel, returning a dummy tensor eliminated the need for changing `utils.py`. This indeed required importing `numpy`, but as you said, it is required by the core openfl framework anyway.

Regarding the longer-term changes you suggested, we can also consider those, but in the meantime offer the "no-op" workspace as a practical workaround.

kminhta (approved):

payalcha (approved):

noopurintel (approved):

PR #1290, opened 2025-01-21
Extend plan initialize to optionally load OpenFL trained model protobuf file

ishaileshpant:
- extend plan initialize to have additional optional argument to take init model path (pbuf format)
- added new function to utils.py
- rebased 21.Jan.1
- reduce cyclo-complexity of initialize function
Signed-off-by: Shailesh Pant <shailesh.pant@intel.com>

PR #1528, opened 2025-04-11
Load native model with fx plan initialize

kminhta:
## Summary
This PR enables pretrained models to be loaded in their native file format via:
`fx plan initialize -i <model_path>`

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
The ability to load a pretrained model as a protobuf was enabled in https://github.com/securefederatedai/openfl/pull/1290 for the federated evaluation efforts.

This PR extends that functionality to also accept models in their native file format (i.e. `.pth`) by calling the already existing `taskrunner.load_native()` method using the same `fx plan initialize` flag

When the `--init_model_path` flag is set, `fx plan initialize` will check if the file is a `pbuf` otherwise, it will try to load using `load_native()` before raising an error

_Additional changes:_
- updated both the `runner_gandalf.py` and `runner_pt.py` to be able to load the model state dictionary directly (users will not always save the `model.state_dict()` and the `optimizer.state_dict()` in the same `.pth` file, especially for inferencing)
- added `load_native()` method to `runner_xgb.py`

## Testing
Tested locally by loading in pretrained model and verifying weights

## Additional Information

_Security considerations:_
We purposely did not use an allow list for file extensions in order to stay framework agnostic. Since `load_native()` is part of the task runner and can be modified in the `taskrunner.py` subclass, a user can potentially modify it to load anything during initialization. This is important for flexibility for more complex models, but it is also a risk that participants must consider and it could be easily but thoroughly checked when reviewing the workload if `load_native()` is modified

### DIFF
--- a/openfl/federated/task/runner_gandlf.py
+++ b/openfl/federated/task/runner_gandlf.py
@@ -419,9 +419,13 @@ class GaNDLFTaskRunner(TaskRunner):
                 dict in picked file. Defaults to 'optimizer_state_dict'.
             **kwargs: Additional keyword arguments.
         """
-        pickle_dict = pt.load(filepath)  # nosec B614
-        self.model.load_state_dict(pickle_dict[model_state_dict_key])
-        self.optimizer.load_state_dict(pickle_dict[optimizer_state_dict_key])
+        pickle_dict = pt.load(filepath, weights_only=True)  # nosec B614
+
+        if model_state_dict_key in pickle_dict and optimizer_state_dict_key in pickle_dict:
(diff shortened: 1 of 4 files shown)


teoparvanov (approved):
LGTM, thanks @kta-intel !!

ishaileshpant:
LGTM @kta-intel - Thanks for taking this up 🎉

ishaileshpant (approved):

rahulga1 (approved):

Issue #1321, opened 2025-01-30
we shouldn't be saving any new models if running in evaluation mode.


ishaileshpant:
Current behavior:
In evaluation run of the federation the aggregator saves the model protobuf states (best and last pbuf files) which is not relevant for evaluation run as no training is performed.

Affected Version:
1.7

Expected behavior:

We shouldn't save any model proto buf files only an init.pbuf should be sufficient to run a evaluation run and make evaluation run idempotent from model state perspective

rahulga1:
Fixed with https://github.com/securefederatedai/openfl/pull/1538

PR #1538, opened 2025-04-16
return from save_model immediately if federation evaluation mode is enabled

rahulga1:
## Summary
currently save_model function is storing best.pbuf and last.pbuf was saved unnecessarily in federated evaluation. As evaluation is not training the model, this step is not needed, hence skip the step.

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
currently save_model function is storing best.pbuf and last.pbuf was saved unnecessarily in federated evaluation. As evaluation is not training the model, this step is not needed, hence skip the step. Fixes Issue https://github.com/securefederatedai/openfl/issues/1321

## Testing
Locally tested

### DIFF
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -342,7 +342,7 @@ class Aggregator:
         Returns:
             None
         """
- 	# Skip saving model if running in evaluation mode
+        # Skip saving model if running in evaluation mode
         if self.assigner.is_task_group_evaluation():
             logger.info(
                 "Skipping model save for round %s in evaluation mode.",
--- a/openfl/component/aggregator/aggregator.py
+++ b/openfl/component/aggregator/aggregator.py
@@ -342,14 +342,14 @@ class Aggregator:
         Returns:
             None
         """
-        # Skip saving model if running in evaluation mode
+        # Skip saving model files during evaluation runs
         if self.assigner.is_task_group_evaluation():
             logger.info(
                 "Skipping model save for round %s in evaluation mode.",
                 round_number,
             )
             return
-        
+            
         # Extract the model from TensorDB and set it to the new model
         og_tensor_dict, _ = utils.deconstruct_model_proto(
             self.model, compression_pipeline=self.compression_pipeline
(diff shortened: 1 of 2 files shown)


ishaileshpant:
@rahulga1 it would be good to add a unit test for this fix/case as well - wdyt?

ishaileshpant (approved):
LGTM once the unit test case is added

rahulga1:
> @rahulga1 it would be good to add a unit test for this fix/case as well - wdyt?

isnt it too basic?

openfl/component/aggregator/aggregator.py
```diff
@@ -1104,11 +1107,18 @@ def _compute_validation_related_task_metrics(self, task_name) -> dict:
...
-                        self._save_model(round_number, self.best_state_path)
+                        if not self.assigner.is_task_group_evaluation():
+                            logger.info(
+                                f"Round {round_number}: saved the best model with score "
+                                "{agg_results:f}"
+                            )
+                            self._save_model(round_number, self.best_state_path)
+                        else:
+                            logger.info(
+                                f"Round {round_number}: updated best score to {agg_results:f} "
```
ishaileshpant: since its evaluation with only single round run  only and no training is involved - potentially there is no "update" to best score - better to convey/log that something like "... best score observed ... "
wdyt?

ishaileshpant (changes requested):

openfl/component/aggregator/aggregator.py
rahulga1: resolved in latest commit.

ishaileshpant (approved):
Thanks for this change @rahulga1

payalcha:
Post this change in case of Federated Evaluation, key_value_store table of tensor.db will be empty.
And to get evaluated details one need to refer to the aggregator logs.

noopurintel (approved):

teoparvanov (approved):
Looks great, thanks @rahulga1 !

openfl/component/aggregator/aggregator.py
```diff
@@ -1104,11 +1107,18 @@ def _compute_validation_related_task_metrics(self, task_name) -> dict:
...
                     if self.best_model_score is None or self.best_model_score < agg_results:
-                        logger.info(
-                            f"Round {round_number}: saved the best model with score {agg_results:f}"
-                        )
                         self.best_model_score = agg_results
-                        self._save_model(round_number, self.best_state_path)
+                        if not self.assigner.is_task_group_evaluation():
+                            logger.info(
+                                f"Round {round_number}: saved the best model with score "
+                                "{agg_results:f}"
```
MasterSkepticista, line 1114: Bug: this prints as `Round 9: saved the best model with score {agg_results:f}` due to unformatted f-string.

I'll fix this in a subsequent PR.

PR #1542, opened 2025-04-16
Skipping the check to validate collaborator data path as aggregator might not have access to it

rahulga1:
## Summary
Skiping the check to validate collaborator data path in `fx plan initialize`

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
When we initialize the plan, aggregator tries to verify the path of dataset which are on collaborator, in case of remote experiment, collaborator wont be on the same machine. hence it doesnt make sense to validate the path in that case.
Now, we have added input_shape, so MockDataLoader can be intialized with this shape, at the same time we have kept the provision to fall back to validating collaborator path if for some workspace its not possible to get that information. For example: gadlf.

## Testing
Locally tested.


openfl/utilities/dataloading.py
```diff
@@ -73,14 +73,19 @@ def get_dataloader(
...
-    if "seed_data" in plan.config["data_loader"]["settings"] and not os.path.isdir(
-        collaborator_data_path
-    ):
-        os.makedirs(collaborator_data_path)
-        sample_data_zip_file = plan.config["data_loader"]["settings"]["seed_data"]
-        with zipfile.ZipFile(sample_data_zip_file, "r") as zip_ref:
-            zip_ref.extractall(collaborator_data_path)
+    # Skip data path check when prefer_minimal=True and we're likely running on the aggregator
+    # This handles the plan initialization case where aggregator shouldn't check client data paths
+    is_aggregator_initialization = prefer_minimal and input_shape is not None
```
teoparvanov: I have to admit I don't understand very well this "equality":
```python
is_aggregator_initialization = prefer_minimal and input_shape is not None
```

Could you explain a bit more why this is the case? AFAIK, `--prefer_minimal` would rather be used on the collaborator side to initialize the task runner.

kminhta (changes requested):
Thanks for looking into this @rahulga1 ! 

I have some comments:

Isn't this function only called during `fx plan initialize` and `fx model save` ? 

I think using the `collaborator_x` naming convention here is a potential source of confusion, and I also think there is opportunity to clean up the logic.

The intended logic (from what I understand) is that a model owner should be able to initialize the model either by:
1. providing the input shape via `--input_shape`
2. providing the input shape via `data_loader.settings.input_shape` in `plan.yaml`
3. providing seed data 

Option 3 is where things get tricky because it expects `data.yaml` to be populated with a name and data path already. The expectation is either:
1. the model owner provides seed data as a zip file and the specified data path doesn't exist so the code will create the data path and extract the zip file into it and then load this
2. the model owner does not provide a seed path and instead the data path already exists and is populated with the data, so the plan initialize step will load this directly

I see a few issues here:
1. there is an unaccounted for condition where the model owner provides a zip file but the data path already exists. The code should still unzip that file into that path, but it doesn't seem like it will
2. user experience: the data path for this should not be through `data.yaml`, IMO `data/yaml` should be specific to the collaborators or else it will cause this type of confusion. It should be fully contained either in the aggregator configurations or the dataloader configurations. In other words, the concept of the seed data and all the configurations regarding where to find it, extract/save it to, load it from, etc. should be found in the same place.
3. readability: as you called out, this is not intended for the model owner node to pull data from the collaborator node, so having it read `collaborator_x` is confusing (in simulation, we are able to use a collaborator dataloader to initialize the model since it downloads the data on the fly, but this feels more like an exception than an expectation)

Let me know what you think

rahulga1:
> 2\. the data path for this should not be through `data.yaml`, IMO `data/yaml` should be specific to the collaborators or else it will cause this type of confusion. It should be fully contained either in the aggregator configurations or the dataloader configurations. In other words, the concept of the seed data and all the configurations regarding where to find it, extract/save it to, load it from, etc. should be found in the same place.

Thanks @kta-intel for your input. I agree on most of your inputs. So, I have modified the code to have more clarity on what each data-loader is for. Also took care of seed data being used primarily. 

Although I am not sure about data.yaml segregation, may be we can bring it up incrementally, as we go through it.  Let me know what do you think of latest changes.

P.S. I am yet to test my changes, but wanted to give a glimpse what I am thinking of :)

openfl/utilities/dataloading.py
rahulga1: Hey Teo, going through the code, My assumption is when collaborator uses this method they set prefer_minimal as False, while when it is getting called from `fx plan initialize`, it is being set to True, hence I used this existing assumption instead of defining a new variable.

openfl/utilities/dataloading.py
```diff
@@ -41,47 +38,91 @@ def get_dataloader(
...
 
+            # Always extract seed data when it's provided
+            # This ensures fresh data even if the directory already exists
+            with zipfile.ZipFile(seed_data_zip, "r") as zip_ref:
+                zip_ref.extractall(data_path)
+        else:
+            # Warn if seed data was specified but file doesn't exist
+            import logging
+
+            logging.getLogger(__name__).warning(
```
MasterSkepticista: Could you please make these declarations on top?

openfl/utilities/dataloading.py
```diff
@@ -41,47 +38,91 @@ def get_dataloader(
...
-        with zipfile.ZipFile(sample_data_zip_file, "r") as zip_ref:
-            zip_ref.extractall(collaborator_data_path)
+    # Handle seed data if provided in the plan
+    if "seed_data" in plan.config["data_loader"]["settings"]:
+        seed_data_zip = plan.config["data_loader"]["settings"]["seed_data"]
 
-    data_loader = plan.get_data_loader(collaborator_name)
+        # Extract seed data if the zip file exists
+        if os.path.isfile(seed_data_zip):
+            # Create the data directory if it doesn't exist
```
MasterSkepticista: General/overall: we can remove self-explanatory comments

MasterSkepticista (changes requested):

openfl/utilities/dataloading.py
rahulga1: Done

openfl/utilities/dataloading.py
rahulga1: Reduced self-explanatory comments

openfl/utilities/dataloading.py
```diff
@@ -41,47 +40,86 @@ def get_dataloader(
...
-        sample_data_zip_file = plan.config["data_loader"]["settings"]["seed_data"]
-        with zipfile.ZipFile(sample_data_zip_file, "r") as zip_ref:
-            zip_ref.extractall(collaborator_data_path)
+    # Handle seed data if provided in the plan
+    if "seed_data" in plan.config["data_loader"]["settings"]:
+        seed_data_zip = plan.config["data_loader"]["settings"]["seed_data"]
 
-    data_loader = plan.get_data_loader(collaborator_name)
+        if os.path.isfile(seed_data_zip):
+            os.makedirs(data_path, exist_ok=True)
```
ishaileshpant, line 114: It would be good to check the error flow - permissions issues etc while creating the dir

openfl/utilities/dataloading.py
```diff
@@ -41,47 +40,86 @@ def get_dataloader(
...
+
+            # Always extract seed data when it's provided
+            # This ensures fresh data even if the directory already exists
+            with zipfile.ZipFile(seed_data_zip, "r") as zip_ref:
```
ishaileshpant, line 118: similar as above we should consider adding a check for valid zip file before attempting extraction

ishaileshpant:
```suggestion
            # Check for corruption
            corrupt_file = zip_ref.testzip()
            if corrupt_file:
                logger.error(f"Corrupt file found in zip: {corrupt_file}")
                return False
           ... follow through the extraction flow
```

rahulga1: But with current way of implementation, it will throw the exception if file is corrupted and `fx plan initialize` command will fail. As this is first step of the process, we are failing fast. IMO, that should be ok, what do you think?

openfl/utilities/dataloading.py
rahulga1: This part is happening on collaborator node where the assumption is, we have single user with all the required permission, all the relevant port open.

rahulga1:
@teoparvanov @kta-intel @MasterSkepticista will appreciate your further inputs on this PR, if any.

psfoley (commented):
@rahulga1 thanks for the contribution and the refactor! Changes generally look good at this point for what this PR is trying to solve (I do agree with @kta-intel that there's opportunity for larger cleanup, but we can take that piecewise). I would just request some unit tests to verify the functionality in `_get_minimal_dataloader` and `_get_collaborator_dataloader` and after that I think this is good to merge.

payalcha:
In my understanding these changes are not needed. Instead 
1. Removing entries in data.yaml.
2. updating plan.yaml "dataloader" "settings" "input_shape" to some default value.
This change will not impact any downstream like openfl-security as all of it is done in workspace.
Later when we make sure no downstream is getting impacted or all downstream workspace is align, we can throw error in case input_shape is not provided.

rahulga1:
> In my understanding these changes are not needed. Instead
> 
> 1. Removing entries in data.yaml.
> 2. updating plan.yaml "dataloader" "settings" "input_shape" to some default value.
>    This change will not impact any downstream like openfl-security as all of it is done in workspace.
>    Later when we make sure no downstream is getting impacted or all downstream workspace is align, we can throw error in case input_shape is not provided.

@psfoley can we remove the backward compatibility and make input-shape mandatory to be passed while running `fx plan initialize`? That will cleanly seperate aggregator dataloader initialization from collaborator.

psfoley:
> > In my understanding these changes are not needed. Instead
> > 
> > 1. Removing entries in data.yaml.
> > 2. updating plan.yaml "dataloader" "settings" "input_shape" to some default value.
> >    This change will not impact any downstream like openfl-security as all of it is done in workspace.
> >    Later when we make sure no downstream is getting impacted or all downstream workspace is align, we can throw error in case input_shape is not provided.
> 
> @psfoley can we remove the backward compatibility and make input-shape mandatory to be passed while running `fx plan initialize`? That will cleanly seperate aggregator dataloader initialization from collaborator.

`input_shape` is not always needed - so it should not be a mandatory command line parameter to pass with `fx plan initialize`. But I like @payalcha's idea of having a placeholder `input_shape` default plan value that can be pulled in for every workspace. This would result in always being able to rely on the `MockDataLoader` when creating the initial model weights. I agree this would lead to a cleaner separation. Let's talk through this further offline.

openfl-workspace/torch/mnist/plan/plan.yaml
theakshaypant, line 1: With these changes, you reckon we need to modify the [README](https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/torch/mnist/README.md#how-to-run-this-tutorial-local-simulation) for this experiment?

openfl-workspace/flower-app-pytorch/plan/plan.yaml
```diff
@@ -33,6 +33,7 @@ data_loader :
   template : src.loader.FlowerDataLoader
   settings :
     collaborator_count : 2
+        input_shape: [3, 224, 224]
```
noopurintel: Indentation issue.. should be under `settings` directly

rahulga1: removed this plan.yaml

rahulga1:
> > > In my understanding these changes are not needed. Instead
> > > 
> > > 1. Removing entries in data.yaml.
> > > 2. updating plan.yaml "dataloader" "settings" "input_shape" to some default value.
> > >    This change will not impact any downstream like openfl-security as all of it is done in workspace.
> > >    Later when we make sure no downstream is getting impacted or all downstream workspace is align, we can throw error in case input_shape is not provided.
> > 
> > 
> > @psfoley can we remove the backward compatibility and make input-shape mandatory to be passed while running `fx plan initialize`? That will cleanly seperate aggregator dataloader initialization from collaborator.
> 
> `input_shape` is not always needed - so it should not be a mandatory command line parameter to pass with `fx plan initialize`. But I like @payalcha's idea of having a placeholder `input_shape` default plan value that can be pulled in for every workspace. This would result in always being able to rely on the `MockDataLoader` when creating the initial model weights. I agree this would lead to a cleaner separation. Let's talk through this further offline.

@psfoley @payalcha I now defined input_shape in some of the workspace plan so dataloader can leverage this to create the MockDataLoader for initialization purpose. But at the same time, it cant be done for the all the workspaces without bringing additional changes for each workspace, for example: Gandlf, as it expects train_csv and test_csv. 
This way, maintaining these specific values wrt to each workspace in MockDataLoader will become cumbersome. 
Let me know your opinion.

MasterSkepticista:
I'll try to summarize details from offline discussions and this thread.

There are two entities:
* Aggregator: This only needs `input_shape` during fresh initialization, not when pretrained models are a starting point - we agree on this.
* Collaborator: This needs `input_shape` all the time - because input shape determines the construction of a data augmentation pipeline. In simple words, collaborator uses `input_shape` to transform raw samples from disk to a particular tensor of given shape that gets fed to the model(s).

That said, `input_shape` is an attribute of the `Dataset` class (or equivalently `DataLoader` class in OpenFL). We need a static location where `input_shape` can be defined, I do not think the `plan` is the right place for it for the same reason - it is a dataset property.

**Suggested approach:**
Dataset or Dataloader class which is **part** of the workspace package, must declare attributes like `input_shape`, `num_classes`, `sequence_length` etc. as part of its own specification. In code, this may look roughly like:

```python
class HistologyDataset:
  ...
  def get_input_shape(self):
    return [None, 3, 150, 150]

  def get_num_classes(self):
    return 8
```

**Benefits of doing this**
This is standard practice in TFDS. Here is an [example (check the `_info` method)](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image_classification/cifar.py). However, let's consider practical/immediate benefits:
* Dataset remains standalone. When instantiating datasets for debugging, one need not parse the plan.
* **`MockDataloader` is no longer necessary** even for aggregator who does not have access to the data.
* Data scientists can migrate their dataloaders to OpenFL without touching the plan.

**What needs to change to support this?**
* Dataloaders in OpenFL use an initialization constructor. This means when instantiated on the aggregator, dataset must be available or downloaded. We need a separate `ds.download_and_prepare(...)` method which does that on explicit call. This allows aggregator to instantiate a "template" dataset without any real data.
* All attribute methods are already part of dataloader, but they are never overriden. We migrate the responsibility of defining the attributes explicitly within each workspace. This has the added benefit of being cognitively easy on the user - they know what happens where.

rahulga1:
will create a separate PR.

PR #1555, opened 2025-04-23
Disable `tensorboard` logging by default

MasterSkepticista:
Metric writing via TensorBoard is not supported within enclaves due to lack of full support for pythonic multiprocessing within Gramine.

This PR disables TensorBoard logging. By default, metrics will only be synchronously written to a text file. Outside enclave environments, users can enable tensorboard logging via `TENSORBOARD=1` environment variable.


payalcha (approved):

payalcha:
Please update documentation as well.

noopurintel (approved):

PR #1561, opened 2025-04-24
remote attestation for openfl participants

anshubit5516:
## Summary
As part of approved changes for [design discussion](https://github.com/securefederatedai/openfl/discussions/1457 ), added the changes for providing attestation reports for openFl participants. 

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This PR adds basic sgx attestation feature to openFl participants. Feature is controlled via plan flag : enable_remote_attestation. If enable_remote_attestation is True, then before any participant starts, it generates it enclave identity which includes mrenclave/ecdsa key pair and certs/ and attested token of it sgx quote via ITA.

if enable_remote_attestation is True, then there are few additional parameters are required which must be passed an env vars.

- ITA_API_KEY : Its a api key which user must pass to interact with ITA.
- AVS_URL :  Its ITA base url
- ATTESTATION_REPORT_PATH : Its the path which attestation reports will be added like public key/ ITA token report. 

Example on how to pass these params :

CERT_HOST_PATH="<host_path>r"
CERT_CONTAINER_PATH="/workspace/attestation/"

docker run --rm \
        --name "aggregator_container" \
        --network host \
        --security-opt seccomp=unconfined \
        --device=/dev/sgx_enclave \
        -v /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket \
        --mount type=bind,source=./cert_agg.tar,target=/certs.tar \
         **--mount type=bind,source=$CERT_HOST_PATH,target=$CERT_CONTAINER_PATH** \
        **--env ITA_API_KEY=$ITA_API_KEY \
        --env AVS_URL=$AVS_URL \
        --env ATTESTATION_REPORT_PATH=$CERT_CONTAINER_PATH**  \
        ${workspace_name} bash -c "mkdir -p $CERT_CONTAINER_PATH && tar -xf /certs.tar && gramine-sgx fx aggregator start" > /dev/null  2>&1 &

As summarized in [design discussion](https://github.com/securefederatedai/openfl/discussions/1457 ), currently user is expected to verify the authenticity of other participants manually.

## Testing
[Describe the testing done for this PR. If applicable include screenshots.]


teoparvanov:
@anshubit5516, a couple of general comments regarding the PR quality:
- please provide a more thorough description, including a detailed `Testing` section
- could you sign your commit(s) via `git commit -s -m "..."`, so that the DCO check passes and the CI can start?
- make sure the code formatting check passes too (I currently see some non-compliant spacing/indentation in certain places across the PR).

openfl-docker/gramine_app/fx.manifest.template
```diff
@@ -71,4 +71,6 @@ sgx.allowed_files = [
   "file:{{ workspace_root }}/plan/cols.yaml",
   "file:{{ workspace_root }}/plan/data.yaml",
   "file:{{ workspace_root }}/plan/plan.yaml",
+  "file:{{ workspace_root }}/attestation",
+  "file:{{ workspace_root }}/attestation/*",
```
teoparvanov: Wouldn't `"file:{{ workspace_root }}/attestation"` suffice?

openfl-workspace/workspace/plan/defaults/network.yaml
```diff
@@ -7,4 +7,5 @@ settings:
     client_reconnect_interval  : 5
     require_client_auth        : True
     cert_folder                : cert
-    enable_atomic_connections  : False
\ No newline at end of file
+    enable_atomic_connections  : False
+    enable_remote_attestation  : False
```
teoparvanov: (minor) I suggest placing this flag right under `use_tls`, to highlight the future potential for RA-TLS

openfl/interface/aggregator.py
```diff
@@ -34,6 +34,7 @@
 from openfl.utilities import click_types
 from openfl.utilities.path_check import is_directory_traversal
 from openfl.utilities.utils import getfqdn_env
+from openfl.utilities.attestation import attestation_utils as attestation_utils
```
teoparvanov: I think it would be cleaner to simply do `from openfl.utilities.attestation.attestation_utils import AttestationManager`

openfl/interface/collaborator.py
```diff
@@ -25,7 +25,7 @@
 from openfl.interface.cli_helper import CERT_DIR
 from openfl.utilities.path_check import is_directory_traversal
 from openfl.utilities.utils import rmtree
-
+from openfl.utilities.attestation import attestation_utils as attestation_utils
```
teoparvanov: same here?

openfl/interface/aggregator.py
```diff
@@ -90,6 +91,21 @@ def start_(plan, authorized_cols, task_group):
             parsed_plan.config["assigner"]["settings"] = {}
         parsed_plan.config["assigner"]["settings"]["selected_task_group"] = task_group
         logger.info(f"Setting aggregator to assign: {task_group} task_group")
+    
+    # Check if remote attestation is enabled in the plan configuration
+    if parsed_plan.config["network"]["settings"].get("enable_remote_attestation", False):
+        # Fetch remote attestation environment variables
+        env_vars = attestation_utils.fetch_attestation_env_vars()
```
teoparvanov: I suggest renaming `env_vars` to `attestation_env`

openfl/interface/collaborator.py
```diff
@@ -78,6 +78,21 @@ def start_(plan, collaborator_name, data_config):
 
     # TODO: Need to restructure data loader config file loader
 
+    # Check if remote attestation is enabled in the plan configuration
+    if plan.config["network"]["settings"].get("enable_remote_attestation", False):
+        # Fetch remote attestation environment variables
+        env_vars = attestation_utils.fetch_attestation_env_vars()
```
teoparvanov: same here

openfl/interface/collaborator.py
```diff
@@ -78,6 +78,21 @@ def start_(plan, collaborator_name, data_config):
...
+        logger.info(f"Remote attestation environment variables fetched from env: {env_vars}")
+        if env_vars is not None:
+            attestation_mr = attestation_utils.AttestationManager(collaborator_name, env_vars["ATTESTATION_REPORT_PATH"], env_vars["ITA_API_KEY"], env_vars["AVS_URL"])
+            # Generate and store the attestation report
+            attestation_report = attestation_mr.get_attestation_report()
```
teoparvanov: same here - just do `attestation_mr.get_attestation_report()`.

openfl/interface/aggregator.py
```diff
@@ -90,6 +91,21 @@ def start_(plan, authorized_cols, task_group):
...
+        if env_vars is not None:
+            attestation_mr = attestation_utils.AttestationManager("aggregator", env_vars["ATTESTATION_REPORT_PATH"], env_vars["ITA_API_KEY"], env_vars["AVS_URL"])
+            # Generate and store the attestation report
+            attestaion_report = attestation_mr.get_attestation_report()
```
teoparvanov: The `attestation_report` object would only be used in OpenFL-Security, so here I suggest simply doing `attestation_mr.get_attestation_report()`.

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+import hashlib
+import logging
+
+from openfl.utilities.attestation.signer_utils import ECDSASigner
+
+
+logger = logging.getLogger(__name__)
+
+class AttestationManager:
+    """Class to manage attestation for enclaves.
```
teoparvanov: nit - participant enclaves

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def __init__(self, participant_name, attestation_report_path, ita_api_key, avs_url):
+        """Initializes the AttestationManager with the provided parameters.
+
+        Args:
+            participant_name (str): Name of the enclave.
+            attestation_report_path (str): Path to the store attestation reports.
+            ita_api_key (str): API key for ITA.
+            avs_url (str): URL for AVS.
+            enable_remote_attestation (bool): Flag to enable remote attestation.
```
teoparvanov: you don't have a `enable_remote_attestation` parameter here

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+            self.participant_name,
+            self.attestation_report_path,
+            cert,
+            self.avs_url,
+            self.ita_api_key
+        )
+        logger.info(f"AVS report: {avs_report}")
+        return {'mrenclave': mrenclave,
+                'name': self.participant_name,
+                'token': avs_report["token"]}
```
teoparvanov: As mentioned above, you should also return the `cert` and `private_key` here, which will be used in OpenFL-Security

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+            avs_url (str): URL for AVS.
+            enable_remote_attestation (bool): Flag to enable remote attestation.
+        """
+        self.participant_name = participant_name
+        self.attestation_report_path = attestation_report_path
+        self.ita_api_key = ita_api_key
+        self.avs_url = avs_url
+
+
+    def get_attestation_report(
```
teoparvanov: as this also constructs the enclave's identity (ECDSA key pair), I suggest renaming this method to `get_attested_identity`

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+        The function raises a FileNotFoundError if the attestation report path
+        does not exist, and a ValueError if the ITA API key or AVS URL is not set.
+        Args:
+            None
+        Raises:
+            FileNotFoundError: If the attestation report path does not exist.
+            ValueError: If the ITA API key or AVS URL is not set.
+        Returns:
+            dict: A dictionary containing the MRENCLAVE value, participant name,
+                  and AVS report token.
```
teoparvanov, line 236: PS: instead of a dict, I suggest returning an `AttestedIdentity` object here, which contains the following elements:
- name
- private key
- public key (cert)
- MRENCLAVE
- ITA report (token)

The `AttestedIdentity` object could also have a `get_passport()` method that would simply return everything but the private key (useful for saving the "attestation bundle" - effectively the enclave's passport).

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+        # This is the enclave specific ECDSA-P 384 keypair for setting up mTLS
+        # with collaborator enclave
+        priv_key_path = os.path.join(self.attestation_report_path, f"{self.participant_name}_privkey.pem")
+        ecdsa_p_384_signer = ECDSASigner.get_instance()
+        cert = ecdsa_p_384_signer.cert("localhost",mrenclave).encode('utf-8')
+        key = ecdsa_p_384_signer.serialize_private_key()
+        pub_key = ecdsa_p_384_signer.get_pubkey()
+        pubkey_path = os.path.join(self.attestation_report_path, f"{self.participant_name}_pubkey.pem")
+        with open(pubkey_path, "wb") as fh:
+            fh.write(pub_key.encode('utf-8'))
```
teoparvanov: I suggest keeping doing this outside of the method, based on the `AttestedIdentity` object.

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+        quote_avs["quote"] = base64.b64encode(quote).decode('utf-8')
+
+        if orig_data is not None:
+            quote_avs['runtime_data'] = base64.b64encode(orig_data).decode('utf-8')
+        else:
+            quote_avs['runtime_data'] = ''
+
+        # Save the quote in the specified location
+        with open(quote_dump_path, 'w') as fh:
+            json.dump(quote_avs, fh)
```
teoparvanov: Just wondering, why do we need to persist the quote in a file? AFAIK, it's only needed when asking ITA for remote attestation.

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,259 @@
...
+
+        return quote_avs
+
+    def fetch_mrenclave_from_quote(self,quote):
```
teoparvanov: let's explicitly mention that `quote=None` by default

teoparvanov (commented):
LGTM overall, left a few comments from my first review pass.

openfl-workspace/workspace/plan/defaults/network.yaml
anshubit5516: done

openfl/interface/aggregator.py
anshubit5516: there are some generic util methods which are part of attestation_utils and not AttestationManager like fetch_attestation_env_vars. Hence importing such

openfl/interface/collaborator.py
anshubit5516: there are some generic util methods which are part of attestation_utils and not AttestationManager like fetch_attestation_env_vars. Hence importing such

openfl/interface/aggregator.py
anshubit5516: done

openfl/interface/collaborator.py
anshubit5516: done

openfl/interface/collaborator.py
anshubit5516: done

openfl/interface/aggregator.py
anshubit5516: done

openfl/utilities/attestation/attestation_utils.py
anshubit5516: removed

openfl/utilities/attestation/attestation_utils.py
anshubit5516: done

openfl/utilities/attestation/attestation_utils.py
anshubit5516: renamed

openfl/utilities/attestation/attestation_utils.py
anshubit5516: added a new class AttestedIdentity which can be used by openFl security as well

openfl/utilities/attestation/attestation_utils.py
anshubit5516: removed. Added a method in AttestedIdentity to save.

openfl/utilities/attestation/attestation_utils.py
anshubit5516: Its implemented as such in openfl-security. Didn't wanna change the flow, hence kept the same. BTW, ts added in /tmp.

openfl/utilities/attestation/attestation_utils.py
anshubit5516: done

openfl-docker/gramine_app/fx.manifest.template
anshubit5516: removed

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,383 @@
...
+from openfl.utilities.attestation.signer_utils import ECDSASigner
+
+
+logger = logging.getLogger(__name__)
+
+
+class AttestedIdentity:
+    """Class to represent an attested identity for a participant enclave.
+    This class holds the MRENCLAVE value, participant name, AVS report token,
+    certificate, and private key for mTLS communication with the collaborator enclave.
```
teoparvanov: I think this sentence is redundant with the code (and may even become outdated over time). I suggest simply removing it.

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,383 @@
...
+        pubkey_path = os.path.join(self.attestation_report_path, f"{self.participant_name}_pubkey.pem")
+        avs_report = self.get_avs_report_ita(
+            self.participant_name,
+            self.attestation_report_path,
+            cert,
+            self.avs_url,
+            self.ita_api_key
+        )
+        logger.info(f"AVS report: {avs_report}")
+        attestation_identity = AttestedIdentity(
```
teoparvanov: nit - `attested_identity`

openfl/utilities/attestation/attestation_utils.py
teoparvanov: As part of the OpenFL-Security refactoring, I suggest trying to remove this and see if it affects the overall flow.

openfl/utilities/attestation/attestation_utils.py
anshubit5516: done

openfl/utilities/attestation/attestation_utils.py
anshubit5516: removed

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,372 @@
+pass
```
theakshaypant: Unnecessary `pass`.

theakshaypant (commented):
1. Would suggest running the formatting script to fix the code formatting pipeline failure. `sh scripts/format.sh`
2. Tests failing
  ```
  FAILED tests/openfl/interface/test_aggregator_api.py::test_aggregator_start - KeyError: 'network'
  FAILED tests/openfl/interface/test_collaborator.py::test_collaborator_start - TypeError: 'Mock' object is not subscriptable
  ```
3. Would suggest adding the copyright header in the newly added files.

theakshaypant (commented):
I would also suggest adding a page in the documentation to educate the user on how to use this feature.

openfl/interface/aggregator.py
```diff
@@ -91,6 +92,25 @@ def start_(plan, authorized_cols, task_group):
         parsed_plan.config["assigner"]["settings"]["selected_task_group"] = task_group
         logger.info(f"Setting aggregator to assign: {task_group} task_group")
 
+    # Check if remote attestation is enabled in the plan configuration
+    if parsed_plan.config["network"]["settings"].get("enable_remote_attestation", False):
+        # Fetch remote attestation environment variables
+        attestation_env = attestation_utils.fetch_attestation_env_vars()
+        if attestation_env is not None:
+            attestation_mr = attestation_utils.AttestationManager(
```
teoparvanov: The `mr` suffix may be misinterpreted as "measurement", so I suggest using the full `attestation_manager` name.

openfl/interface/collaborator.py
```diff
@@ -78,6 +79,25 @@ def start_(plan, collaborator_name, data_config):
 
     # TODO: Need to restructure data loader config file loader
 
+    # Check if remote attestation is enabled in the plan configuration
+    if plan.config["network"]["settings"].get("enable_remote_attestation", False):
+        # Fetch remote attestation environment variables
+        attestation_env = attestation_utils.fetch_attestation_env_vars()
+        if attestation_env is not None:
+            attestation_mr = attestation_utils.AttestationManager(
```
teoparvanov: same here => `attestation_manager`

openfl/interface/collaborator.py
```diff
@@ -23,6 +23,7 @@
 from openfl.cryptography.participant import generate_csr
 from openfl.federated import Plan
 from openfl.interface.cli_helper import CERT_DIR
+from openfl.utilities.attestation import attestation_utils as attestation_utils
```
psfoley, line 27: Why not this?

```suggestion
from openfl.utilities.attestation import attestation_utils
```

anshubit5516:
> 1. Would suggest running the formatting script to fix the code formatting pipeline failure. `sh scripts/format.sh`
> 
>     2. Tests failing
> 
> 
> ```
> FAILED tests/openfl/interface/test_aggregator_api.py::test_aggregator_start - KeyError: 'network'
> FAILED tests/openfl/interface/test_collaborator.py::test_collaborator_start - TypeError: 'Mock' object is not subscriptable
> ```
> 
>     3. Would suggest adding the copyright header in the newly added files.

openfl/interface/aggregator.py
anshubit5516: done

openfl/interface/collaborator.py
anshubit5516: done

openfl/utilities/attestation/attestation_utils.py
anshubit5516: removed

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,389 @@
...
+            participant_name (str): Name of the enclave.
+            attestation_report_path (str): Path to the store attestation reports.
+            ita_api_key (str): API key for ITA.
+            avs_url (str): URL for AVS.
+        """
+        self.participant_name = participant_name
+        self.attestation_report_path = attestation_report_path
+        self.ita_api_key = ita_api_key
+        self.avs_url = avs_url
+        self.ecdsa_p_384_signer = ECDSASigner.get_instance()
```
teoparvanov: Now that I realize some nuances between OpenFL and OpenFL-Secuiry, I think `AttestationManager` should have two modes. The mode can be f.e. selected via a boolean flag `generate_identity` in combination with optional `private_key` and `cert` parameters, which are file paths:
- Mode 1 (OpenFL): receive an existing private key and cert pair (`generate_identity=False`)
- Mode 2 (OpenFL-Security): generate a public/private key pair (`generate_identity=True`)

In OpenFL you'd get the key/cert paths to those from the plan via `parsed_plan.config["network"]["settings"].get("private_key", None)` and `parsed_plan.config["network"]["settings"].get("certificate", None)` and pass them when instantiating `AttestationManager`, along with `generate_identity=False`.

In OpenFL-Security, you would instantiate `AttestationManager` by passing `generate_identity=True`, which generate a private key and a cert, saving them and building them into the `ECDSASigner` object that is used in the rest of the OpenFL-Security flow:
`ECDSASigner.get_instance(private_key_path)`.

I hope this is clear. If not, we can also discuss additionally offline.

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def get_attested_identity(self, cert_host="localhost"):
+        """Generates the attestation report for the enclave.
```
teoparvanov: I think the following would be more accurate:
```python
"""Generates an attested identity, where a public/private key pair is bound to the workload measurement (MRENCLAVE) via a remote attestation report.
```

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,389 @@
...
+    Raises:
+        None
+
+    Returns:
+        dict: A dictionary containing the attestation environment variables.
+    """
+    env_vars = {
+        "ITA_API_KEY": os.getenv("ITA_API_KEY"),
+        "AVS_URL": os.getenv("AVS_URL"),
+        "ATTESTATION_REPORT_PATH": os.getenv("ATTESTATION_REPORT_PATH", None),
```
teoparvanov, line 447: You should make sure those environment variables are available in the OpenFL-Security workspace container too.

openfl/utilities/attestation/signer_utils.py
teoparvanov, line 1: As `signer_utils` isn't (only) related to attestation, I'd suggest moving it to `openfl/cryptography`, and renaming the module to simply `signing`.

teoparvanov (commented):
It's looking good @anshubit5516, thanks a ton ! I've added a couple of comments from the perspective of how the `AttestationManager` would be used downstream (f.e. in OpenFL-Security).

openfl/utilities/attestation/attestation_utils.py
anshubit5516: done

openfl/utilities/attestation/attestation_utils.py
anshubit5516: As discussed, since certs generated via Attestation manager has mrenclave embedded in certificate. So it's not advisable to use it two modes suggested above.

openfl/utilities/attestation/signer_utils.py
anshubit5516: moved and renamed

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,385 @@
...
+        """
+        self.participant_name = participant_name
+        self.attestation_report_path = attestation_report_path
+        self.ita_api_key = ita_api_key
+        self.avs_url = avs_url
+        self.privkey_path = (
+            privkey_path is not None
+            and privkey_path
+            or os.path.join(attestation_report_path, f"{participant_name}_privkey.pem")
+        )
```
teoparvanov, line 161: nice 😊 !

openfl/interface/aggregator.py
```diff
@@ -91,6 +92,28 @@ def start_(plan, authorized_cols, task_group):
         parsed_plan.config["assigner"]["settings"]["selected_task_group"] = task_group
         logger.info(f"Setting aggregator to assign: {task_group} task_group")
 
+        # Check if remote attestation is enabled in the plan configuration
+    if "network" in parsed_plan.config and "settings" in parsed_plan.config["network"]:
+        if parsed_plan.config["network"]["settings"].get("enable_remote_attestation", False):
```
teoparvanov: this would be equivalent to simply writing:
```python
if parsed_plan.config.get("network", {}).get("settings", {}).get("enable_remote_attestation", False):
```

openfl/interface/collaborator.py
```diff
@@ -77,7 +78,27 @@ def start_(plan, collaborator_name, data_config):
 
     # TODO: Need to restructure data loader config file loader
     logger.info(f"Data paths: {plan_obj.cols_data_paths}")
-    echo(f"Data = {plan_obj.cols_data_paths}")
+    # this check is added to avoid mock objects failing
+    if isinstance(plan_obj, Plan):
+        if "network" in plan_obj.config and "settings" in plan_obj.config["network"]:
+            if plan_obj.config["network"]["settings"].get("enable_remote_attestation", False):
```
teoparvanov: same here

openfl/interface/aggregator.py
```diff
@@ -91,6 +92,28 @@ def start_(plan, authorized_cols, task_group):
...
+            if attestation_env is not None:
+                attestation_manager = attestation_utils.AttestationManager(
+                    "aggregator",
+                    attestation_env["ATTESTATION_REPORT_PATH"],
+                    attestation_env["ITA_API_KEY"],
+                    attestation_env["AVS_URL"],
+                )
+                # Generate and store the attestation report
+                attestation_manager.get_attested_identity()
+                logger.info("Remote attestation report fetched successfully.")
```
teoparvanov: Why not mention the attestation report path too?

openfl/interface/collaborator.py
```diff
@@ -77,7 +78,27 @@ def start_(plan, collaborator_name, data_config):
...
+                if attestation_env is not None:
+                    attestation_manager = attestation_utils.AttestationManager(
+                        collaborator_name,
+                        attestation_env["ATTESTATION_REPORT_PATH"],
+                        attestation_env["ITA_API_KEY"],
+                        attestation_env["AVS_URL"],
+                    )
+                    # Generate and store the attestation report
+                    attestation_manager.get_attested_identity()
+                    logger.info("Remote attestation report stored successfully.")
```
teoparvanov: same

openfl/interface/aggregator.py
```diff
@@ -91,6 +92,28 @@ def start_(plan, authorized_cols, task_group):
...
+            else:
+                logger.error("Failed to fetch remote attestation environment variables.")
+        else:
+            logger.info(
+                "Remote attestation is not enabled. Skipping attestation report generation."
+            )
```
teoparvanov: As this is practically the same code between the collaborator and aggregator, why not extract it as `get_remote_attestation()`, f.e. in `attestation_utils.py` ?

openfl/utilities/secagg/signer.py
teoparvanov, line 1: I don't think `signer.py` has anything to do with `secagg`, so please consider moving it to `openfl.cryptography` or `openfl.utilities`, as discussed previously.

teoparvanov (approved):
LGTM, with a final set of comments:

openfl/utilities/secagg/signer.py
psfoley: +1 to this. It belongs in `openfl.cryptography`

openfl-workspace/workspace/plan/defaults/network.yaml
```diff
@@ -4,6 +4,7 @@ settings:
     agg_port                   : auto
     hash_salt                  : auto
     use_tls                    : True
+    enable_remote_attestation  : False
```
psfoley: I would suggest changing this and moving the `enable_remote_attestation` parameter to the `aggregator` and `collaborator` sections so that it can be configured independently; I expect it will be common that only the aggregator can run inside SGX

openfl/interface/aggregator.py
psfoley: Yes - this code should be pulled out to `openfl.utils` to maintain readability of the aggregator / collaborator `start` functions

psfoley (changes requested):
Thanks for the PR, @anshubit5516. I see a couple gaps here:

- You have logic to create TLS credentials specific to the enclave, but it doesn't look like it gets used after the ITA report is generated (i.e. in the gRPC server). Am I missing something?
- I have some questions about why a private key path should be passed into the `AttestationManager`. I don't think there's a security hole because I think what's intended is the user runs the docker run command with `...gramine-sgx fx aggregator start` as the entrypoint, but generally I think it would be safer to hardcode the path to ensure it's specific to the enclave's local FS
- Enabling remote attestation for aggregator / collaborators independently. Secure aggregator running inside SGX will be a very common ask and use case.
- Documentation: What remote attestation is, what problems it solves, and all of the steps to perform it with this PR.
- Tests: I understand it will not be easy to implement for E2E CI because of SGX dependency, but some tests would be useful here given the scope of the changes

teoparvanov:
Hey @psfoley , thank you for the detailed review!
Let me try to address your comments:
> * You have logic to create TLS credentials specific to the enclave, but it doesn't look like it gets used after the ITA report is generated (i.e. in the gRPC server). Am I missing something?

We envisioned this approach as an initial iteration for 1.9, where the attestation "bundle" (cert+mrenclave+ITA token) is generated, but not wired in OpenFL, and only used E2E in OpenFL-Security (ref. [design discussion summary](https://github.com/securefederatedai/openfl/discussions/1457#discussioncomment-12767350)). I understand this is incomplete from an OpenFL perspective, but it was intended as an intermediate step. Now that it becomes clear that the OpenFL-Security changes won't be able to make it into GA, we could consider the following options (among others potentially):
1. Deliver what we currently have for OpenFL 1.9, knowing that the ITA token and cert from the bundle can be verified together as a sort of "offline passport", but without using this identity in the actual TLS communication. This is probably of limited value, but it's a start (esp. for any upcoming live federations built on OpenFL 1.9)...
2. Attempt to course-correct by additionally wiring the cert generated within the enclave to the gRPC server, knowing that this might not make it into 1.9. The difficulty here is related to distributing the certificates, which is a manual process in OpenFL versus automated in OpenFL-Security. Short of implementing a Governor-like entity, one option would be if only the aggregator exposed an "attested" certificate. So, instead of generating a CSR and signing the Aggregator certificate via `fx collaborator certify`, in this approach the Aggregator's cert would be generated on enclave start up, and distributed to collaborators manually (as part of the attestation bundle).

> * I have some questions about why a private key path should be passed into the `AttestationManager`. I don't think there's a security hole because I think what's intended is the user runs the docker run command with `...gramine-sgx fx aggregator start` as the entrypoint, but generally I think it would be safer to hardcode the path to ensure it's specific to the enclave's local FS

I see, it makes sense. We'll discuss with @anshubit5516 how this can be achieved to align with the [usage in OpenFL-Security](https://github.com/intel-innersource/frameworks.ai.openfl.openfl-security/blob/develop/client/openfl_security/interface/collaborator_fp.py#L202).

> * Enabling remote attestation for aggregator / collaborators independently. Secure aggregator running inside SGX will be a very common ask and use case.

Good point, we'll provide separate properties for those. Maybe we could even start with just the aggregator for this first iteration (`enable_aggregator_attestation`), because the workflow around semi-manual attestation of the collaborators isn't very clear at this point.

> * Documentation: What remote attestation is, what problems it solves, and all of the steps to perform it with this PR.

This will be done either in this PR, or in a follow-up one, but it is indeed in scope for the task.
> * Tests: I understand it will not be easy to implement for E2E CI because of SGX dependency, but some tests would be useful here given the scope of the changes

We will discuss with @anshubit5516 what can be done in a non-SGX context (maybe by using mock SGX device files and a mock ITA service?). It may be challenging though, so we will likely primarily rely on the E2E test by QE (manual, to begin with).

openfl/utilities/attestation/attestation_utils.py
```diff
@@ -0,0 +1,385 @@
...
+            avs_url (str): URL for AVS.
+        """
+        self.participant_name = participant_name
+        self.attestation_report_path = attestation_report_path
+        self.ita_api_key = ita_api_key
+        self.avs_url = avs_url
+        self.privkey_path = (
+            privkey_path is not None
+            and privkey_path
+            or os.path.join(attestation_report_path, f"{participant_name}_privkey.pem")
```
teoparvanov: Now that I re-look at this, it is likely not safe to save the private key into the attestation report path. Let's discuss additionally how to deal with this (hard-coded path vs defaults vs other options?).

teoparvanov (changes requested):

noopurintel:
Please ensure to add documentation changes as well to include this functionality or modify the `README.md` file.

openfl/interface/aggregator.py
anshubit5516: This check is added as unit tests are sending Mock Object which doesn't have config or is an instance of PLAN. Hence added the check.

openfl/interface/collaborator.py
anshubit5516: This check is added as unit tests are sending Mock Object which doesn't have config or is an instance of PLAN. Hence added the check.

openfl/interface/aggregator.py
```diff
@@ -91,6 +92,13 @@ def start_(plan, authorized_cols, task_group):
         parsed_plan.config["assigner"]["settings"]["selected_task_group"] = task_group
         logger.info(f"Setting aggregator to assign: {task_group} task_group")
 
+    # check if remote attestation is enabled
+    if parsed_plan.config["aggregator"]["settings"].get("enable_remote_attestation", False):
+        # check if the aggregator is running in a remote attestation environment
+        attestation_utils.get_remote_attestation(parsed_plan, "aggregator")
```
teoparvanov: As pointed out by @psfoley , we will likely need to introduce a `fx collaborator certify --import-self-signed` command for the aggregator's attested certificate to be added to each collaborators' trust store.

openfl/utilities/secagg/__init__.py
```diff
@@ -1,13 +1,2 @@
 # Copyright 2020-2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-
-
-from openfl.utilities.secagg.crypto import (
```
theakshaypant, line 5: Removing these imports might break the secure aggregation flow.  I would suggest keeping them unless it is causing any unexpected issues.

openfl/interface/collaborator.py
```diff
@@ -78,7 +79,14 @@ def start_(plan, collaborator_name, data_config):
...
     logger.info(f"Data paths: {plan_obj.cols_data_paths}")
-    echo(f"Data = {plan_obj.cols_data_paths}")
+    # this check is added to avoid mock objects failing
+
+    # check if remote attestation is enabled
+    if plan_obj.config["collaborator"]["settings"].get("enable_remote_attestation", False):
+        attestation_utils.get_remote_attestation(plan_obj, collaborator_name)
+    else:
+        logger.info("Remote attestation is not enabled.")
+
```
teoparvanov: As discussed offline, let's remove the collaborator attestation code from this PR.

openfl/interface/aggregator.py
anshubit5516: done

anshubit5516:
In the last commit I have updated the code so that if enable_remote_attestation is True for Aggregator, then it will generate its own certificate and with the help of env var (ROOT_CERT_PATH) can export its certificate to host. Now if this cert is passed to collaborators use this as root cert it can allow connection to its grpc connection. But catch is we have to **disable require_client_auth**. It will only work if collaborators certs are available to aggregator before gRPC server starts.

openfl/interface/collaborator.py
anshubit5516: done

PR #1567, opened 2025-04-28
Enhancement in pypi publish

payalcha:
## Summary
Enhancement and improvement in pypi publish,

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
1. Adding commit hash in the source so that one can get exact commit-id for which Pypi hash is created. image
2. Introduces a new input (**test_pypi**). It will help someone to understand implication in downstream as well. By this, one can push package in testpypi option from their PR branch. And test with that package downstream applications like openfl-security.

3. **Improvement** - Testing the package post push in respective repository.
4. As discussed below in comments - 
There is requirement to add multiple push on same date in case of any hotfix. To achieve the same, changed the versioning format from 
1.9.0.dev<date> ---> 1.9.0.dev<date>0 for openfl-nightly 
Reason of above format - Due to PEP440 restrictions "." or "-" is not supported after 1.9.0.dev<date>.
To decide the new_version
1st I verify old_version with suggested new_version if same I add 1 to last digit of old_version.
**"+"** Pypi is not allowing in its version.

As I result for same date version will be like 
image

## Testing
Tested against Test Pypi - 
https://github.com/payalcha/openfl/actions/runs/14713844986

### DIFF
--- a/.github/workflows/publish_nightly_package.yml
+++ b/.github/workflows/publish_nightly_package.yml
@@ -74,6 +74,7 @@ jobs:
         echo "OpenFL version is $(pip list | grep openfl)" 
         python3 -m pip uninstall -y openfl openfl-nightly
         python3 -m pip install ${{ env.OPENFL_NIGHTLY_WHL_FILE }}
+        python3 -m pip install -r test-requirements.txt
         python -m pytest -s tests/end_to_end/test_suites/task_runner_tests.py \
         -m task_runner_basic --model_name "keras/mnist" \
         --num_rounds 1 --num_collaborators 1
--- a/.github/workflows/publish_nightly_package.yml
+++ b/.github/workflows/publish_nightly_package.yml
@@ -52,12 +52,13 @@ jobs:
         echo "Repository URL: $full_url"
         version=$(grep -oP "(?<=version=')[^']+" setup.py)
         date_suffix=$(date +%Y%m%d)
-        new_version="${version}${date_suffix}0"
+        new_version="${version}${date_suffix}-0"
         sed -i 's/name=.*/name="openfl-nightly",/' setup.py
         sed -i "s/version=.*/version='$new_version',/" setup.py
         sed -i 's/Development Status :: 5 - Production\/Stable/Development Status :: 4 - Beta/' setup.py
         sed -i "s|'Source Code': '.*'|'Source Code': '${full_url}'|g" setup.py
         echo "NEW_VERSION=${new_version}" >> $GITHUB_ENV
+        echo ${{env.TEST_PYPI}} ${{inputs.test_pypi}}
 
     - name: Build package
       id: build_package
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_nightly_package.yml
```diff
@@ -42,12 +47,17 @@ jobs:
     - name: Modify setup.py for nightly build
       run: |
         echo "Creating package..."
+        base_url="https://github.com/securefederatedai/openfl/tree/"
+        full_url="${base_url}${{ env.COMMIT_ID }}"
+        echo "Repository URL: $full_url"
         version=$(grep -oP "(?<=version=')[^']+" setup.py)
         date_suffix=$(date +%Y%m%d)
-        new_version="${version}${date_suffix}"
+        new_version="${version}${date_suffix}0"
```
teoparvanov: What's the `0` suffix for?

teoparvanov (approved):
LGTM!

bsenthilr:
hi @payalcha thanks for taking care of this. 
As we discussed we would require a way to push hotfix. This requires a workflow support to trigger a hotfix which will promote wheels with PEP standard semantic versioning to reflect it.

bsenthilr (approved):

payalcha:
@bsenthilr 
Simplest suggestion can be adding '0' post date. 
For this we need to do below changes - 
1. Check if the pypi exists for current date 
2. If exists increment last digit (limitation max 10 push allowed in a day)

@teoparvanov  @rahulga1  @psfoley 
Shall we change the version from 
`1.9.0.dev<date>` ---> `1.9.0.dev<date>0` for openfl-nightly

For e.g - 
image

rahulga1 (approved):

rahulga1:
LGTM

.github/workflows/publish_nightly_package.yml
```diff
@@ -58,30 +57,48 @@ jobs:
...
         -m task_runner_basic --model_name "keras/mnist" \
-        --num_rounds 1 --num_collaborators 2
+        --num_rounds 1 --num_collaborators 1
         echo "Verified the package installation successfully!"
 
     - name: Upload package to PyPI
       run: |
-        python3 -m twine upload dist/openfl_nightly* --verbose -u __token__ -p ${{ secrets.PYPI_API_TOKEN }}
-        echo "Package published successfully!"
+        if [ "${{ env.TEST_PYPI }}" ]; then
```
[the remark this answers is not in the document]
payalcha: Boolean can't be compare like this.

.github/workflows/publish_nightly_package.yml
payalcha: Added in description needed details.

bsenthilr:
@payalcha wondering if https://peps.python.org/pep-0440/#local-version-identifiers is cleaner.
Keep the default format. For hotfixes
1.9.0.dev<date>+<hotfix#>

Eg., 1.9.0.dev20250428+2

WDYT?

theakshaypant:
Perhaps we can consider `1.9.0.dev<date>+<commit_hash>`.

psfoley:
> Perhaps we can consider `1.9.0.dev<date>+<commit_hash>`.

@theakshaypant I suspect that the last suffix should be ascending to be recognized as the latest.

@bsenthilr I was thinking the same. Keep the counter but separate with either a `+` or `.` so that date is read cleanly. @payalcha after that change I think this is ready to merge.

theakshaypant:
> > Perhaps we can consider `1.9.0.dev<date>+<commit_hash>`.
> 
> @theakshaypant I suspect that the last suffix should be ascending to be recognized as the latest.

@psfoley You are right! 
> Comparison and ordering of local versions considers each segment of the local version (divided by a `.`) separately. If a segment consists entirely of ASCII digits then that section should be considered an integer for comparison purposes and if a segment contains any ASCII letters then that segment is compared lexicographically with case insensitivity. When comparing a numeric and lexicographic segment, the numeric section always compares as greater than the lexicographic segment. Additionally a local version with a great number of segments will always compare as greater than a local version with fewer segments, as long as the shorter local version’s segments match the beginning of the longer local version’s segments exactly.

[Ref](https://peps.python.org/pep-0440/#local-version-identifiers).

payalcha:
> > Perhaps we can consider `1.9.0.dev<date>+<commit_hash>`.
> 
> @theakshaypant I suspect that the last suffix should be ascending to be recognized as the latest.
> 
> @bsenthilr I was thinking the same. Keep the counter but separate with either a `+` or `.` so that date is read cleanly. @payalcha after that change I think this is ready to merge.


@psfoley 
As I mentioned in comment we can't keep  '." after 1.9.0.dev<date> as it is against [PEP440](https://peps.python.org/pep-0440/) mandate. And "+" pypi is not accepting.

@theakshaypant As sequence is issue, I kept commit-hash in source-code url, so that you can get commit-hash.

payalcha:
> https://peps.python.org/pep-0440/#local-version-identifiers

PyPi is not allowing '+' in version after 1.9.0.dev<date>.


image

noopurintel (approved):

bsenthilr:
> PyPi is not allowing '+' in version after 1.9.0.dev.
thats a bummer so we need to keep looking for pep way. Looks like pypi is more strict towards upstream versioning.

PR #1570, opened 2025-04-29
Verify that one collaborator cannot get tensors from another one.

yuliasherman:
## Summary
Verify that one collaborator cannot get tensors from another one.

## Type of Change (Mandatory)
Bug fix  - security hole

## Description (Mandatory)
The PR prohibits getting tensors of another collaborator specified through the 'tags' fielsd.

## Testing
Added unit tests.


openfl/component/aggregator/aggregator.py
```diff
@@ -545,13 +545,28 @@ def _straggler_cutoff_time_elapsed(self) -> None:
             # Check if minimum collaborators reported results
             self._end_of_round_with_stragglers_check()
 
+    def _check_tags(self, tags: tuple[str, ...], allowed_col: str) -> bool:
+        """
+        Check if all tags are either the allowed column or unauthorized.
```
teoparvanov: "col" should refer to collaborator, rather than column, right?

openfl/component/aggregator/aggregator.py
```diff
@@ -545,13 +545,28 @@ def _straggler_cutoff_time_elapsed(self) -> None:
...
+        Args:
+            tags (tuple[str, ...]): The set of tags to check.
+            allowed_col (str): The only authorized column allowed in the tags.
+
+        Returns:
+            bool: True if all tags are valid, False if an unauthorized column
+            (other than allowed_col) is found.
+        """
+        for tag in tags:
+            if tag in self.authorized_cols and tag != allowed_col:
```
teoparvanov, line 564: `authorized_cols` contains the names of collaborators authorized for the experiment (effectively, the entries in `cols.yaml`)

teoparvanov (commented):
LGTM, with some minor comments:

openfl/component/aggregator/aggregator.py
yuliasherman: Fixed

openfl/component/aggregator/aggregator.py
yuliasherman: Correct

teoparvanov (approved):
LGTM, thanks @yuliasherman !

openfl/component/aggregator/aggregator.py
```diff
@@ -545,13 +545,28 @@ def _straggler_cutoff_time_elapsed(self) -> None:
...
+        return True
+
     def get_aggregated_tensor(
-        self,
-        tensor_name,
-        round_number,
-        report,
-        tags,
-        require_lossless,
+        self, tensor_name, round_number, report, tags, require_lossless, requested_by
```
rahulga1, line 569: where is this field `requested_by` is being set?

psfoley: @rahulga1 here: https://github.com/securefederatedai/openfl/pull/1570/files#diff-5ccc7efabe87b6f422c5c1911011b5665e3d06fd0959b82f80706c880b4876d5R251

psfoley (approved):
Thanks for implementing this @yuliasherman. Approved

openfl/component/aggregator/aggregator.py
rahulga1: Got it. Thanks @psfoley

rahulga1 (approved):

openfl/component/aggregator/aggregator.py
```diff
@@ -577,9 +593,14 @@ def get_aggregated_tensor(
         else:
             compress_lossless = False
 
+        if not self._check_tags(tags, requested_by):
+            logger.error(
+                "Tag check failed: unauthorized tags detected. Only '%s' is allowed.", requested_by
+            )
+            return NamedTensor()
```
theakshaypant, line 600: Should we not raise an error if someone tries to access an unauthorised tensor?

tests/openfl/component/aggregator/test_aggregator.py
```diff
@@ -135,16 +136,41 @@ def test_get_tasks(agg, col_name, tasks, time_to_quit,
     assert (tasks, sleep_time, time_to_quit) == (exp_tasks, exp_sleep_time, exp_time_to_quit)
 
 
-def test_get_aggregated_tensor(agg):
-    """Test that test_get_tasks is failed without a correspond data."""
-    tensor_name = 'test_tensor_name'
-    require_lossless = False
+def test_get_aggregated_tensor_behavior(agg):
```
theakshaypant, line 139: Nit: In my opinion, this can be split into two separate tests.

openfl/component/aggregator/aggregator.py
teoparvanov: We wouldn't like the aggregator to crash if a malicious collaborator requested another participant's tensors, would we? At the same time, we have a trace of this in the logs (via the error log above). I believe this is sufficient for now.

tests/openfl/component/aggregator/test_aggregator.py
yuliasherman: The test verifies a correct behaviour of the get_aggregated_tensor API, it does not worth splitting.

openfl/component/aggregator/aggregator.py
theakshaypant: Right, makes sense.

yuliasherman: We agreed on this behaviour in an offline discussion.

theakshaypant (approved):

PR #1575, opened 2025-04-30
Batched fetching of tensors to reduce RPC calls [1/2]

MasterSkepticista:
### Motivation

Collaborators today fetch each model tensor from the aggregator via a dedicated RPC call. Aggregator has limited resources to serve requests, and it is not uncommon to have hundreds (if not thousands) of model tensors waiting to be served to each collaborator. A federation itself may have hundreds (if not thousands) of participants making these requests.

### Example
Consider `torch/histology` experiment. 

* Model size is ~64MB with about 20 tensors. 
* In a federation with 8 collaborators, a total of 160 RPC calls are made **each** round to the aggregator just to fetch latest model tensors. 
* In contrast, only `(get_tasks + 3 x send_local_task_results) x 8 collaborators` 32 RPC calls are made for other purposes in each round. 
* **>80%** of all requests are consumed in serving tensors. This also leads to thread contention. Aggregator cannot generally afford more than 8-32 threads assuming modest CPU/memory availability.
* Had the tensors been batched per request, each collaborator could be served with a single RPC thread and no aggregator-side contention.

Problem gets worse when models have hundreds of tensors and experiments span more collaborators.

### Goal of _this_ PR

This is **Part 1 (of 2)** PRs that adds support for batched fetching of tensors over gRPC. This PR makes the following major changes:
1. Remove `send_model_deltas` and `use_delta_updates` support.
2. Simplify `nparray_to_named_tensor` (henceforth called `serialize_tensor`) and `named_tensor_to_nparray` (henceforth called `deserialize_tensor`) without any brittle conditional checks.
3. Batch gRPC requests for tensors on the collaborator and aggregator side.

In **Part 2** of this PR, delta update support will be added back. The reason for removal is to straighten the logic for actions that concern a communication pipeline (de/compression, de/serialization) and the aggregator/collaborator component.

### Implementation

A new RPC call replaces the `get_aggregated_tensor` with `get_aggregated_tensors` (plural 's'). Collaborators request a batch of tensors in a new `TensorSpec` format.

```proto
message TensorSpec {
  string tensor_name = 1;
  int32 round_number = 2;
  bool report = 3;
  repeated string tags = 4;
  bool require_lossless = 5;
}
```

You may observe the resemblance with `TensorKey`, except that origin field is missing. 
The RPC request and response formats are shown below:

```proto
message GetAggregatedTensorsRequest {
  MessageHeader header = 1;
  repeated TensorSpec tensor_specs = 2;
}

message GetAggregatedTensorsResponse {
  MessageHeader header = 1;
  repeated NamedTensor tensors = 2;
}
```

On collaborator side, tensors are fetched via `self.fetch_tensors_from_aggregator(tensor_keys)` where `tensor_keys` are of type `List[TensorKey]`.

### Cost/Benefit Analysis
One downside of this PR is that potential bandwidth savings achieved due to delta updates is lost.
The upside of this PR is a significantly simplified mental model of how tensors are processed on both ends, higher correctness confidence and higher maintainability.

Indeed, the second part will close the drawback by bringing in delta updates, keeping the mental model simple.

**Reviewer note(s)**: This PR has a lot of cleanup. It is best understood by looking at the proposed changes directly and not diffs.


openfl/component/aggregator/aggregator.py
```diff
@@ -992,82 +823,13 @@ def _prepare_trained(self, tensor_name, origin, round_number, report, agg_result
...
-        if base_model_nparray is not None and self.use_delta_updates:
-            delta_tk, delta_nparray = self.tensor_codec.generate_delta(
-                agg_tag_tk, agg_results, base_model_nparray
-            )
-        else:
-            # This condition is possible for base model
-            # optimizer states (i.e. Adam/iter:0, SGD, etc.)
-            # These values couldn't be present for the base
-            # model because no training occurs on the aggregator
-            delta_tk, delta_nparray = agg_tag_tk, agg_results
```
teoparvanov, line 1022: AFAIK, transmitting deltas instead of the actual updated weights only makes the transmitted tensor "sparse", without reducing its dimensions. In fact, if my understanding is correct, using deltas only reduces the payload size if combined with some form of compression that benefits from the sparsity of the transmitted tensors.

For small-ish models, dropping the "delta" capability shouldn't make a difference, however LLM support may be impacted for TaskRunner API. WDYT?

openfl/protocols/aggregator.proto
```diff
@@ -11,8 +11,8 @@ import "openfl/protocols/base.proto";
 service Aggregator {
   rpc Ping(PingRequest) returns (PingResponse) {}
   rpc GetTasks(GetTasksRequest) returns (GetTasksResponse) {}
-  rpc GetAggregatedTensor(GetAggregatedTensorRequest) returns (GetAggregatedTensorResponse) {}
-  rpc SendLocalTaskResults(stream DataStream) returns (SendLocalTaskResultsResponse) {}
+  rpc GetAggregatedTensors(GetAggregatedTensorsRequest) returns (GetAggregatedTensorsResponse) {}
+  rpc SendLocalTaskResults(TaskResults) returns (SendLocalTaskResultsResponse) {}
```
teoparvanov: The [InfoSec overview](https://openfl.readthedocs.io/en/latest/infosec-orverview.html#overview-of-contents-of-network-messages) will need to be slightly adjusted following these API changes.

openfl/transport/grpc/aggregator_client.py
```diff
@@ -416,8 +416,7 @@ def send_local_task_results(
             tensors=named_tensors,
         )
 
-        # convert (potentially) long list of tensors into stream
-        response = self.stub.SendLocalTaskResults(utils.proto_to_datastream(request))
+        response = self.stub.SendLocalTaskResults(request)
```
teoparvanov: What's the reasoning behind removing the streaming functionality?

openfl/component/collaborator/collaborator.py
```diff
@@ -266,142 +262,35 @@ def do_task(self, task, round_number) -> dict:
...
         )
+        if len(tensor_keys) > 0:
+            logger.info("Fetching %d tensors from the aggregator", len(tensor_keys))
+            named_tensors = self.client.get_aggregated_tensors(tensor_keys, require_lossless=True)
 
-        # this translates to a numpy array and includes decompression, as
-        # necessary
-        nparray = self.named_tensor_to_nparray(tensor)
+            # Deserialize tensors and mark them as coming from the aggregator.
+            for tensor_key, named_tensor in zip(tensor_keys, named_tensors):
```
teoparvanov, line 299: `zip(tensor_keys, named_tensors)` assumes that `get_aggregated_tensors` returns the tensors in the same order as the keys. Is this guaranteed? If so, maybe you should document this in the method's docstring.

teoparvanov (commented):
I have a couple of comments from my first read-through:

openfl/component/aggregator/aggregator.py
MasterSkepticista: Yes, it is impactful for larger models, and this feature will be restored immediately after this PR. We do not have any concrete numbers of achievable compression ratios as of now.

openfl/protocols/aggregator.proto
MasterSkepticista: Yes, thanks for pointing this out!

openfl/transport/grpc/aggregator_client.py
MasterSkepticista: Streaming packets is linked to memory growth rate from local experiments. Currently streaming is disabled here to be consistent with other RPCs.

That said, streaming is required for large models, and it ought to be a generic implementation - one that can be toggled without consideration for the return type of RPCs (like a wrapper)

openfl/component/collaborator/collaborator.py
MasterSkepticista: Order will be the same as both components iterate in loops. It will be documented, thanks!

openfl/transport/grpc/aggregator_client.py
teoparvanov: As discussed during our meeting, I'd suggest doing this in a separate PR.

kminhta (approved):
Thanks for taking this up @MasterSkepticista - this is a great fundamental adjustment to how we transport the model across the network.  I don't see any major concerns, I'm going to go ahead and approve this after my review and our offline discussion with the understanding that:

1. delta updates get reintroduced in the follow up PR
2. streaming gets reintroduced in a more generic manner (as you called out in https://github.com/securefederatedai/openfl/pull/1575#discussion_r2093260966) in some subsequent PR as well - or otherwise some manner of handling large models from both the aggregator and collaborator

I know you called out point 1 in the PR, but do you have a mental model for how point 2 can be achieved or is the general implementation still an open? I ask mainly because it would be good to close the gap for first class LLM support (thinking in terms of pretraining), but it is also true that LLM fine-tuning has a large library of parameter efficient methods that may still allow for expanding OpenFL-LLM capabilities in the interim

kminhta:
FYI, a couple of the tests are failing - secure aggregation in particular https://github.com/securefederatedai/openfl/actions/runs/15063524782/job/42343165534?pr=1575

kminhta:
Also, going to tag @ishaileshpant to keep him in the loop as there are modifications to RPC calls that may affect https://github.com/securefederatedai/openfl/pull/1500

openfl/component/aggregator/aggregator.py
```diff
@@ -607,74 +606,28 @@ def get_aggregated_tensor(
...
-
-        start_retrieving_time = time.time()
-        while nparray is None:
-            logger.debug("Waiting for tensor_key %s", agg_tensor_key)
-            time.sleep(5)
-            nparray = self.tensor_db.get_tensor_from_cache(agg_tensor_key)
-            if (time.time() - start_retrieving_time) > 60:
-                break
-
+        nparray = self.tensor_db.get_tensor_from_cache(tensor_key)
```
porteratzo, line 573: Is this polling loop never used or why is it being removed?

porteratzo (approved):
LGTM

openfl/component/aggregator/aggregator.py
MasterSkepticista: Yes, I have never seen it being used in historical logs or any recent experiments. Given the locations where `get_aggregated_tensors` are called, I made the assumption that unavailability of those tensors has to be a logical error -> raise exception instead

MasterSkepticista:
> Thanks for taking this up @MasterSkepticista - this is a great fundamental adjustment to how we transport the model across the network. I don't see any major concerns, I'm going to go ahead and approve this after my review and our offline discussion with the understanding that:
> 
> 1. delta updates get reintroduced in the follow up PR
> 2. streaming gets reintroduced in a more generic manner (as you called out in [Batched fetching of tensors to reduce RPC calls [1/2] #1575 (comment)](https://github.com/securefederatedai/openfl/pull/1575#discussion_r2093260966)) in some subsequent PR as well - or otherwise some manner of handling large models from both the aggregator and collaborator
> 
> I know you called out point 1 in the PR, but do you have a mental model for how point 2 can be achieved or is the general implementation still an open? I ask mainly because it would be good to close the gap for first class LLM support (thinking in terms of pretraining), but it is also true that LLM fine-tuning has a large library of parameter efficient methods that may still allow for expanding OpenFL-LLM capabilities in the interim

Yes, delta updates will be reintroduced shortly. Streaming changes are reverted in this PR. I don't have concrete idea on how to generalize (2). But since we discussed its orthogonality with the current set of changes, we will attack it separately

openfl/protocols/aggregator.proto
MasterSkepticista: Done https://github.com/securefederatedai/openfl/pull/1575/commits/fcbfff42306686e086e93b28ee5ec0b844c69480

openfl/component/collaborator/collaborator.py
MasterSkepticista: Documented in function call

openfl/component/aggregator/aggregator.py
```diff
@@ -830,45 +750,47 @@ def process_task_results(
...
+            # Record
+            result_tensor_dict[tensor_key] = nparray
+            task_results.append(tensor_key)
 
             if "metric" in tensor_key.tags:
-                # Caution: This schema must be followed. It is also used in
-                # gRPC message streams for director/envoy.
+                assert nparray.ndim == 0, (
+                    f"Expected metric to be a scalar, got shape {nparray.shape}"
+                )
```
teoparvanov, line 726: Is this constraint really required? Granted it's rare, but we could imagine certain more sophisticated metrics to be multi-dimensional.

teoparvanov (approved):
LGTM, thanks @MasterSkepticista !

openfl/component/aggregator/aggregator.py
MasterSkepticista: Metrics used to train/validate a model always reduce over batch dimension. I haven't seen multidimensional metrics... do you have any examples? If so I will lift the assert. 

P.S.: `assert` is only added for readability, it will fail regardless when `float(x)` encounters a non-scalar input

teoparvanov: OK, so we actually don't support non-scalar metrics because of the `float(x)` transformation. It's then indeed better to have a more explicit assertion that captures this limitation.

theakshaypant (approved):
Nice!
I do think we wait for 1.9 cut-off before merging this given the temporary removal of use_delta_updates.

PR #1576, opened 2025-04-30
Edar/datasets in experiment

Efrat1:
## Summary
Support using verifiable datasets in experiment, includes
* Multiple data sources
* S3 data sources
* Verifiability

Add histology S3 workspace example.

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
* Allow multiple data sources, including S3
* S3 workspace example
* CLI command for calculating root hash of a given dataset
* Add 'name' to data sources classes
* Keep backward compatibility

## Testing
Pass verifiable existing unittests
Run full experiments using JSON definition of the data sources (new method)
and  full experiments using data_path (the tradtional method)

**Run instructions:**
1. Run MINIO or S3 server.
2. Set: 
`export YOUR_ACCESS_KEY=<your_access_key>`
`export YOUR_SECRET_KEY=<your_secret_key>`
example:
`export MINIO_ROOT_PASSWORD=minioadmin`
`export MINIO_ROOT_USER=minioadmin`

3. Setup workspace and aggregator:
`fx workspace create --prefix ./hist_s3 --template torch/histology_s3`
`cd hist_s3/`
`pip install -r requirements.txt > /dev/null`
`fx workspace certify`
`fx aggregator generate-cert-request --fqdn localhost`
`fx aggregator certify --fqdn localhost --silent`

4. Create client: `--data_path` should point a directory that contains a file `datasources.json`, foe example:
`fx collaborator create --data_path plan/collaborator1  -n bob --silent`

5. Optional: Calculate the root hash of the collaborator's dataset.
The hash will be saved at `<data_path>/hash.txt`. In dataloader, if this file is found, dataset integrity verification will be performed.
`fx collaborator calchash --data_path plan/collaborator1`

6. Complete client certification:
`fx collaborator generate-cert-request -n bob --silent`
`fx collaborator certify --request-pkg col_bob_to_agg_cert_request.zip --silent`
`fx collaborator certify --import agg_to_col_bob_signed_cert.zip`

7. Repeat steps 4,5,6 for each client

8. Initialize plan:
`fx plan initialize -a localhost`

9. Start experiment (for example, collaborators bob and charlie):
`fx aggregator start & fx collaborator start -n bob & fx collaborator start -n charlie`

### DIFF
--- a/openfl-workspace/torch/histology_s3/src/dataloader.py
+++ b/openfl-workspace/torch/histology_s3/src/dataloader.py
@@ -1,0 +1,213 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""You may copy this file as the starting point of your own model."""
+
+from collections.abc import Iterable
+from logging import getLogger
+import os
+import sys
+
+
+from openfl.federated import PyTorchDataLoader
+import numpy as np
+from openfl.federated.data.sources.torch.verifiable_map_style_image_folder import VerifiableImageFolder
+from openfl.federated.data.sources.data_sources_json_parser import DataSourcesJsonParser
+from openfl.utilities.path_check import is_directory_traversal
+import torch
+from torch.utils.data import random_split
+from torchvision.transforms import ToTensor
+
+
+logger = getLogger(__name__)
+
+
+class PyTorchHistologyVerifiableDataLoader(PyTorchDataLoader):
+    """PyTorch data loader for Histology dataset."""
+
+    def __init__(self, data_path, batch_size, **kwargs):
+        """Instantiate the data object.
+
+        Args:
+            data_path: The file path to the data
+            batch_size: The batch size of the data loader
+            **kwargs: Additional arguments, passed to super init
+             and load_mnist_shard
+        """
+        super().__init__(batch_size, random_seed=0, **kwargs)
+
+        # Set Histology-specific default attributes
+        self.feature_shape = [3, 150, 150]
+        self.num_classes = 8
+
+        # If data_path is None, this is being used for model initialization only
+        if data_path is None:
+            logger.info("Initializing dataloader for model creation only (no data loading)")
+            return
+
+        verifible_dataset_info = self.get_verifiable_dataset_info(data_path)
+
+        hash_file_path = os.path.join(data_path, "hash.txt")
+        verify_dataset_items = False
+        if os.path.isfile(hash_file_path):
+            logger.info(f"Found hash file at: {hash_file_path}. Verifying dataset...")
+            verify_dataset_items = True
+            with open(hash_file_path, "r", encoding="utf-8") as file:
+                hash_value = file.read()
+            dataset_valid = verifible_dataset_info.verify_dataset(root_hash=hash_value)
+            if not dataset_valid:
+                logger.error("The dataset is not valid.")
+                sys.exit(1)
+            else:
+                logger.info("The dataset is valid.")
+
+        _, num_classes, X_train, y_train, X_valid, y_valid = load_histology_shard(
+            verifible_dataset_info=verifible_dataset_info, verify_dataset_items=verify_dataset_items, **kwargs)
+
+        self.X_train = X_train
+        self.y_train = y_train
+        self.X_valid = X_valid
+        self.y_valid = y_valid
+
+        self.num_classes = num_classes
+
+
+    def get_feature_shape(self):
+        """Returns the shape of an example feature array.
+
+        Returns:
+            list: The shape of an example feature array [3, 150, 150] for Histology images.
+        """
+        return self.feature_shape
+
+    def get_num_classes(self):
+        """Returns the number of classes for classification tasks.
+
+        Returns:
+            int: The number of classes (8 for Histology dataset).
+        """
+        return self.num_classes
+
+    def get_verifiable_dataset_info(self, data_path):
+        """
+        Parse dataset information from `datasources.json` in the given directory.
+
+        Args:
+            data_path (str): Path to the directory containing `datasources.json`.
+
+        Returns:
+            VerifiableDatasetInfo: Parsed dataset information.
+
+        Raises:
+            SystemExit: If `data_path` is invalid or missing `datasources.json`.
+        """
+        """Return the verifiable dataset info object for the given data sources."""
+        if data_path and is_directory_traversal(data_path):
+            logger.error("Data path is out of the openfl workspace scope.")
+        if not os.path.isdir(data_path):
+            logger.error("The data path must be a directory.")
+            sys.exit(1)
+
+        datasources_json_path = os.path.join(data_path, "datasources.json")
+        if not os.path.isfile(datasources_json_path):
+            logger.error("The directory must contain a file named 'datasources.json' at the first level.")
+            sys.exit(1)
+        with open(datasources_json_path, "r", encoding="utf-8") as file:
+            data = file.read()
+        return DataSourcesJsonParser.parse(data)
+
+
+class HistologyDataset(VerifiableImageFolder):
+    """Colorectal Histology Dataset."""
+
+    def __init__(self, **kwargs) -> None:
+        """Initialize."""
+        super().__init__(**kwargs)
+
+    def __getitem__(self, index):
+        """Allow getting items by slice index."""
+        if isinstance(index, Iterable):
+            return [super().__getitem__(i) for i in index]
+        else:
+            return super().__getitem__(index)
+
+
+def _load_raw_data(verifiable_dataset_info, verify_dataset_items=False, train_split_ratio=0.8, **kwargs):
+    """
+    Load the raw data by shard.
+
+    Returns tuples of the dataset shard divided into training and validation.
+
+    Args:
+        data_path (str): The path to the dataset.
+
+    Returns:
+        2 tuples: (image, label) of the training, validation dataset
+    """
+    dataset = HistologyDataset(
+        vds=verifiable_dataset_info,
+        transform=ToTensor(),
+        verify_dataset_items=verify_dataset_items
+    )
+    n_train = int(train_split_ratio * len(dataset))
+    n_valid = len(dataset) - n_train
+    ds_train, ds_val = random_split(
+        dataset, lengths=[n_train, n_valid], generator=torch.manual_seed(0))
+
+    # create the shards
+    X_train, y_train = list(zip(*ds_train))
+    X_train, y_train = np.stack(X_train), np.array(y_train)
+
+    X_valid, y_valid = list(zip(*ds_val))
+    X_valid, y_valid = np.stack(X_valid), np.array(y_valid)
+
+    return (X_train, y_train), (X_valid, y_valid)
+
+
+
+def load_histology_shard(verifible_dataset_info, verify_dataset_items,
+                         categorical=False, channels_last=False, **kwargs):
+    """
+    Load the Histology dataset.
+
+    Args:
+        data_path (str): path to data directory
+        categorical (bool): True = convert the labels to one-hot encoded
+         vectors (Default = True)
+        channels_last (bool): True = The input images have the channels
+         last (Default = True)
+        **kwargs: Additional parameters to pass to the function
+
+    Returns:
+        list: The input shape
+        int: The number of classes
+        numpy.ndarray: The training data
+        numpy.ndarray: The training labels
+        numpy.ndarray: The validation data
+        numpy.ndarray: The validation labels
+    """
+    img_rows, img_cols = 150, 150
+    num_classes = 8
+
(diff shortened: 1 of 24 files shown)


openfl-workspace/torch/histology_s3/plan/datasources.yaml
```diff
@@ -0,0 +1,6 @@
+# Copyright (C) 2025 Intel Corporation
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+bob, plan/datasources_conf.json
+charlie,plan/datasources_2_conf.json
```
teoparvanov: Instead of creating a separate yaml file for data sources, I'd suggest enhancing the functionality around `data.yaml` (incl. the `fx collaborator create -d` command) - to accommodate for (optional) datasources defined as `json` configurations. This could be done for example by adding a `-ds`/`--data_sources` CLI parameter (see [documentation reference](https://openfl.readthedocs.io/en/latest/reference/fx.collaborator.html#fx-collaborator-create)).

Note that the names of the collaborators should not be present in the `yaml` files of the workspace template, as they are defined when configuring the actual workspace (see the [quickstart guide](https://openfl.readthedocs.io/en/latest/tutorials/taskrunner.html) for reference).

openfl/utilities/dataloading.py
```diff
@@ -35,7 +35,7 @@ def get_dataloader(
             Defaults to `None`.
         collaborator_index (int ?):
             which collaborator should be used for initializing dataloader
-            among collaborators specified in plan/data.yaml.
```
teoparvanov: I'm not sure I understand the need for these changes...

teoparvanov (changes requested):
Thanks @Efrat1 , this is a good starting point for showcasing the new dataset/datasource class hierarchy in practice. However, I have a suggestion to change the approach to configuring the datasources themselves, to align with the existing process of setting up `data.yaml`:

openfl-workspace/torch/histology_s3/plan/datasources.yaml
```diff
@@ -0,0 +1,6 @@
+# Copyright (C) 2025 Intel Corporation
+# Licensed subject to the terms of the separately executed evaluation license agreement between Intel Corporation and you.
+
+bob, plan/datasources_conf.json
```
MasterSkepticista: "I know a file, who knows a file" :p

openfl-workspace/torch/histology_s3/plan/datasources.yaml
MasterSkepticista, line 1: Since this is a `yaml`, it may be possible to cross-reference information from the same file. Would it be easier to define collaborator/datapaths on top and then define the remainder dataset configuration (i.e. contents of JSON) in this same file?

openfl-workspace/torch/histology_s3/src/dataloader.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(self, **kwargs) -> None:
+        """Initialize."""
+        super().__init__(**kwargs)
+
+    def __getitem__(self, index):
+        """Allow getting items by slice index."""
+        if isinstance(index, Iterable):
+            return [super(HistologyDataset, self).__getitem__(i) for i in index]
+        else:
+            return super(HistologyDataset, self).__getitem__(index)
```
MasterSkepticista: Nit: Python 3 `super()` behaves same as `super(BaseClass, self)`

openfl-workspace/torch/histology_s3/src/dataloader.py
```diff
@@ -0,0 +1,161 @@
...
+            
+    # To verify the integrity of the dataset, follow these steps:
+    # 1. Calculate the hash of the dataset using the appropriate command:
+    #    Example: `fx collaborator calchash -n <collaborator_name> -j <path/to/datasources_file.json>`
+    # 2. Retrieve the verifiable dataset information from the arguments:
+    #    `vds = kwargs.get('verifiable_dataset_info')`
+    # 3. Verify the dataset hash using the `vds.verify_dataset(calculated_hash)` method.
+    #    This can be called during the initialization of the dataloader.
+    # Note: To enable `verify_dataset_items=True`, you must first call `vds.verify_dataset(calculated_hash)`.
+    #       This ensures that the hash for each individual item in the dataset is calculated and verified internally.
```
MasterSkepticista: Suggest moving this to function docstring. Helps during editor syntax highlighting or autogenerated documentation.
On a different note, I'm not sure who is the target audience of this comment block. Is this a guideline to the user to verify the integrity? If yes, is this the right contextual location?

openfl/federated/data/sources/data_source.py
```diff
@@ -23,16 +23,19 @@ class DataSource(ABC):
 
     Attributes:
         type (str): The storage type of the data source
+        name (str): The name of the data source
```
MasterSkepticista: Generally a good practice to mention accepted (or expected) values for `type`.

openfl/federated/plan/plan.py
```diff
@@ -450,13 +483,101 @@ def get_data_loader(self, collaborator_name):
...
-        defaults[SETTINGS]["data_path"] = self.cols_data_paths[collaborator_name]
+        datasources = self._get_local_data_sources(collaborator_name, defaults)
+        datasources.extend(self._get_data_sources(collaborator_name))
+
+        if len(datasources) == 0:
+            raise ValueError(f"No data sources found for collaborator {collaborator_name}.")
+
+        vds = VerifiableDatasetInfo(
+            data_sources=datasources,
+            label="",
```
MasterSkepticista: What does `label` refer to? Couldn't infer from the class definition (also worth documenting)
It may cause confusion with `label` in the context of ML

openfl/federated/plan/plan.py
```diff
@@ -450,13 +483,101 @@ def get_data_loader(self, collaborator_name):
...
+
+    def _load_json_file(self, file_path):
+        """Load and parse a JSON file."""
+        try:
+            with open(file_path, "r", encoding="utf-8") as file:
+                return json.load(file)
+        except json.JSONDecodeError as e:
+            raise Exception(f"Invalid JSON in file {file_path}: {e}")
+        except Exception as e:
+            raise Exception(f"An unexpected error occurred while loading {file_path}: {e}")
```
MasterSkepticista: I don't think this much error checking is required. `json.load` handles most standard conditions and exceptions are raised in python unless explicitly handled differently.

MasterSkepticista (changes requested):
Overall, I'm not fully convinced of adding an extra config JSON for datasets (and the downstream impact on relevant commands with a new flag)

I was a bit iffy on the role of `data.yaml` earllier because it only pointed to a path that each collaborator knew separately. There was then a discussion around eliminating `data.yaml` and instead request the data path information during `fx collaborator start -n name -d /path/to/data`, since that was the only contribution made by that file.

However, this PR raises an interesting scenario where each collaborator doesn't immediately start at a data path. Instead it reads a JSON which then fetches the dataset. I see two options to keep the number of config jumps minimal (and also make it easy to developers):

1. Expand the scope of `data.yaml` to also include JSON contents as a yaml within the same file, OR
2. Point collaborator to the same data path. Instead the onus is on dataset class to search for standard identifier file(s) like `sources.json`. If found, those contents are read and datasets are fetched from relevant location(s). If not, the folder itself is treated as a dataset as it is being done today.

openfl-workspace/torch/histology_s3/plan/datasources.yaml
vrancurel: @teoparvanov let's talk when you are back.

openfl/utilities/dataloading.py
Efrat1: Reverted

setup.py
```diff
@@ -94,6 +94,7 @@ def run(self):
         'tensorboardX',
         'protobuf>=4.21,<6.0.0',
         'grpcio>=1.56.2,<1.66.0',
+        'boto3>=1.37.19',
```
teoparvanov: We are very conservative when adding dependencies to the core OpenFL package.

Why is `boto3` needed at the OpenFL level? Could it instead be made as a workspace-level dependency by including it in `histology_s3/requirements.txt`?

openfl-workspace/torch/histology_s3/plan/collaborator1/datasources.json
teoparvanov, line 1: The data source descriptors don't belong to the plan, which is shared with all participants. In practice they will only be available locally at each collaborator.

So, to better emulate the real topology in this example, I suggest creating a `data` folder right under `histology_s3`, and moving the `collaborator1` and `collaborator2` folders there. This would effectively yield the following directory structure:
```
histology_s3/
├── data
        └── collaborator1
        └── collaborator2
```

This way, you would more naturally write:
```shell
fx collaborator create --data_path data/collaborator1  -n bob --silent
fx collaborator calchash --data_path data/collaborator1
...
fx collaborator create --data_path data/collaborator2  -n bob --silent
fx collaborator calchash --data_path data/collaborator2
```

WDYT?

openfl-workspace/torch/histology_s3/src/dataloader.py
```diff
@@ -0,0 +1,187 @@
...
+from openfl.utilities.path_check import is_directory_traversal
+import torch
+from torch.utils.data import random_split
+from torchvision.transforms import ToTensor
+
+
+logger = getLogger(__name__)
+
+
+class PyTorchHistologyInMemory(PyTorchDataLoader):
```
teoparvanov: nit - this is no longer "in memory". We could rename it to `PyTorchHistologyVerifiableDataLoader`

teoparvanov (approved):
This looks great, thank you @Efrat1 ! Also, thanks for describing the user workflow - this both helps with the review, and serves as a foundation for a documentation page 😉 

A couple of final comments from me:

openfl/federated/data/sources/data_source.py
Efrat1: I fixed the type to expect `DataSourceType`

openfl-workspace/torch/histology_s3/src/dataloader.py
Efrat1: Thanks, done

openfl-workspace/torch/histology_s3/src/dataloader.py
Efrat1: Right, fixed

openfl-workspace/torch/histology_s3/plan/collaborator1/datasources.json
Efrat1: Good idea, done

setup.py
Efrat1: Not sure. The thing is that collaborator.py (interface file) imports `DataSourcesJsonParser` (for `calchash`) which in the end imports `boto3`, so any workflow uses the collaborator CLI commands imports `boto3` in the end. See the current failure in CI for example.
Do you have an idea how can we avoid that?

theakshaypant: Hey @Efrat1 ! I faced a similar issue with `pycrytodome` dependency for secure aggregation. What helped there was moving the import statements inside the classes/methods where appropriate. Linking the PR for your reference. 
https://github.com/securefederatedai/openfl/pull/1420/

teoparvanov: Hi @Efrat1 and @theakshaypant, I see... For now I would suggest extracting this functionality into a separate `datahash.py` script at the base of the `histology_s3` workspace. Similarly to the `fx` command, it would receive a single parameter for the path to the data folder. This way, you could also "contain" the `boto3` dependency at the workspace level.

Later on, we'll try to figure out what the general governance flow for OpenFL should look like (including the data commitment step).

Efrat1: I think that what @theakshaypant suggests works nicely and simply (thank you!)

teoparvanov: Ok, but let's help the user by providing a hint about the required package, as it was done for pycryptodome:
```python
from importlib import util

# Check if pycryptodome is installed.
if util.find_spec("Crypto") is None:
    raise Exception(
        "'pycryptodome' not installed."
        "This package is necessary when secure aggregation is enabled."
    )
```

(adapt for boto3 using an appropriate known class that could replace "Crypto")

openfl/federated/data/sources/s3_data_source.py
```diff
@@ -16,14 +14,17 @@ class S3DataSource(DataSource):
...
+        name,
         uri: str,
         endpoint=None,
         access_key_env_name=None,
         secret_key_env_name=None,
         secret_name=None,
         hash_func=None,
     ):
-        super().__init__(DataSourceType.S3)
+        import boto3
```
theakshaypant, line 30: As Teo suggested in another comment, let's raise an exception telling user the package they should install.

setup.py
Efrat1: Nice, added before the import of boto3, and same for azure.

teoparvanov (approved):

PR #1580, opened 2025-04-30
[Task Runner API] [Flower Interoperability] Enable option to run `ClientApp` as a separate process in SGX

kminhta:
## Summary
This is an optimization change. When running in an SGX enclave, each process forks a new enclave via checkpoint-and-restore. This is a very expensive operation that introduces significant overhead. The default behavior in a Flower experiment is for the SuperNode to spawn a new ClientApp subprocess for each task/round. Flower has an option to spawn the ClientApp manually via `--isolation process` flag. By doing this, a single ClientApp process will run for the duration of the experiment, significantly reducing overall runtime. This PR brings in that functionality when running in SGX.

## Type of Change (Mandatory)
- Feature enhancement  
- Refactor (naming clean-up)

## Description (Mandatory)
This PR:
- adds `--isolation process` flag to the SuperNode command in order to run the ClientApp manually when in SGX
- adds additional logic create clientapp process command and manage (start/stop) it
- changes `patch` -> `SGX_enabled` in `plan.yaml` and whenever called. This is to better describe the functionality of this setting/flag
- change `local_grpc_server` -> `interop_server`. This is to more accurate reflect the component name
- remove `try... except KeyboardInterrupt` from `runner.py` since the interop_server object already has handling for signal interruption

## Testing
Manually tested with and without SGX enclave

Unofficial profiling of container/experiment runtime when in default (`--isolation subprocess`) vs as a manual process (`isolation process`) relative to a baseline (no SGX) for 10 rounds. Already, there is great performance boost. The expectation is for the gap to increase as the rounds increases since the default mode will still be spawning child processes each round whereas the manual process mode will not.

image


openfl-workspace/flower-app-pytorch/plan/plan.yaml
```diff
@@ -22,7 +22,7 @@ connector :
     flwr_run_params :
       flwr_app_name : "app-pytorch"
       federation_name : "local-poc"
-      patch: True
+      SGX_enabled: False
```
teoparvanov: Nice! It's much more self-explanatory now than the somewhat crypting `patch` flag.

openfl-workspace/flower-app-pytorch/plan/plan.yaml
```diff
@@ -39,7 +39,7 @@ task_runner :
   template : src.runner.FlowerTaskRunner
   settings :
     flwr_app_name: app-pytorch
-    patch: True
+    SGX_enabled: False
```
teoparvanov: nit - `sgx_enabled` maybe, for visual consistency?

teoparvanov (approved):
LGTM!

PR #1586, opened 2025-05-02
Retry download of `openfl-nightly` package after its publish

noopurintel:
## Summary
Retry download of `openfl-nightly` package after its publish.

## Type of Change (Mandatory)
- Feature enhancement

## Description (Mandatory)
As part of Product Quality Pipeline run, after the `openfl-nightly` package is pushed to pypi, we are downloading the same and running a basic E2E test.

Somehow it takes time for the package to be available for download as we get error like

`ERROR: No matching distribution found for openfl-nightly==1.9.0.dev202505012`

Although, if we try the same after a few minutes, it works fine.

In this PR, I have added the retry mechanism with 5 attempts and 20s sleep time to help give sometime between publish and download of the package.

## Testing
Triggered workflow from my branch - https://github.com/noopurintel/openfl/actions/runs/14791388068/job/41529117596#step:8:45

*IMP - if it still doesn't work, will need to further increase the attempt/sleep time.*

### DIFF
--- a/.github/workflows/publish_nightly_package.yml
+++ b/.github/workflows/publish_nightly_package.yml
@@ -64,16 +64,16 @@ jobs:
         echo "OPENFL_NIGHTLY_WHL_FILE=$(ls dist/*.whl)" >> $GITHUB_ENV
         echo "Package built successfully!"
 
-    # - name: Verify package installation before pushing to PyPI
-    #   run: |
-    #     echo "OpenFL version is $(pip list | grep openfl)" 
-    #     python3 -m pip uninstall -y openfl openfl-nightly
-    #     python3 -m pip install ${{ env.OPENFL_NIGHTLY_WHL_FILE }}
-    #     python3 -m pip install -r test-requirements.txt
-    #     python3 -m pytest -s tests/end_to_end/test_suites/task_runner_tests.py \
-    #     -m task_runner_basic --model_name "keras/mnist" \
-    #     --num_rounds 1 --num_collaborators 1
-    #     echo "Verified the package installation successfully!"
+    - name: Verify package installation before pushing to PyPI
+      run: |
+        echo "OpenFL version is $(pip list | grep openfl)" 
+        python3 -m pip uninstall -y openfl openfl-nightly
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_nightly_package.yml
```diff
@@ -89,10 +89,25 @@ jobs:
...
-          python3 -m pip install -i https://test.pypi.org/simple/ openfl-nightly==${{ env.NEW_VERSION }}
-        else
-          python3 -m pip install openfl-nightly==${{ env.NEW_VERSION }}
+        # Install the package with retry logic
+        for i in {1..5}; do
+          if [ "${{ env.TEST_PYPI }}" == "true" ]; then
+            PIP_CMD="python3 -m pip install -i https://test.pypi.org/simple/ openfl-nightly==${{ env.NEW_VERSION }}"
+          else
+            PIP_CMD="python3 -m pip install openfl-nightly==${{ env.NEW_VERSION }}"
+          fi
```
teoparvanov: `PIP_CMD` doesn't need to be re-calculated on every retry iteration, so I suggest moving this logic out of the `for` loop.

.github/workflows/publish_nightly_package.yml
```diff
@@ -89,10 +89,25 @@ jobs:
...
+          if $PIP_CMD; then
+            echo "Successfully installed openfl-nightly using command '$PIP_CMD'"
+            break
+          else
+            echo "Attempt $i failed. Retrying in 20 seconds..."
```
teoparvanov: `Attempt $i/5 ...` ?

teoparvanov (approved):
LGTM, thanks @noopurintel ! I've just noticed a couple of minor things:

.github/workflows/publish_nightly_package.yml
noopurintel: Done.

.github/workflows/publish_nightly_package.yml
noopurintel: Done; also kept variables for max attempts and wait time instead of hard coded values.

PR #1590, opened 2025-05-06
Refactor data loading components for improved initialization and flexibility

rahulga1:
## Summary
Restrucure data_loader functionality so it can be decoupled from actual data loading from the model initialization purpose

## Type of Change (Mandatory)
- Restructuring Fix for https://github.com/securefederatedai/openfl/issues/73

## Description (Mandatory)
Cleanup the plan.py and dataloader.py
add relevant methods in workspace-> dataloader class, so initialization can happen properly

## Testing
Local testing


openfl/utilities/dataloading.py
```diff
@@ -1,13 +1,15 @@
...
 # SPDX-License-Identifier: Apache-2.0
+import importlib
+import logging
 import os
 import zipfile
 from typing import Union
 
 from openfl.federated import Plan
 from openfl.federated.data.loader import DataLoader
-from openfl.utilities.mocks import MockDataLoader
```
MasterSkepticista, line 9: I assume this PR will also remove `MockDataLoader` class?

openfl/utilities/dataloading.py
```diff
@@ -25,13 +27,11 @@ def get_dataloader(
         plan (Plan):
             plan object linked with the dataloader
         prefer_minimal (bool ?):
-            prefer to use MockDataLoader which can be used to more easily
-            instantiate task_runner without any initial data.
+            prefer to initialize dataloader without loading actual data.
+            Used primarily for model initialization.
```
MasterSkepticista: I think an aggregator can and should always build shape without actual data. Even if samples in a dataset are variable-shape, it is padded/cropped to a fixed shape for any ML model. Therefore, it need not be `prefer_minimal`. It will _always_ be minimal, i.e., constructed based on static shapes defined in the custom method overrides of the dataloader.

This will greatly simplify the design.

Cc: @psfoley, @kta-intel Please confirm?

openfl/utilities/dataloading.py
```diff
@@ -42,31 +42,43 @@ def get_dataloader(
...
-    collatorators_count = len(collaborator_names)
+    collaborators_count = len(collaborator_names)
 
-    if collaborator_index >= collatorators_count:
-        raise Exception(
+    if collaborator_index >= collaborators_count:
+        raise ValueError(
             f"Unable to construct full dataloader from collab_index={collaborator_index} "
-            f"when the plan has {collatorators_count} as total collaborator count. "
+            f"when the plan has {collaborators_count} as total collaborator count. "
```
MasterSkepticista, line 86: Collaborator count/idx will become irrelevant if aggregator loads shape statically always (see prev. comment). Can be removed if that is the approach taken.

openfl/federated/data/loader_gandlf.py
```diff
@@ -52,8 +66,24 @@ def get_feature_shape(self):
...
+        # If we have a train dataloader with a dataset that has a gandlf_params attribute with patch_size
+        if (self.train_dataloader is not None and 
+            hasattr(self.train_dataloader, 'dataset') and
+            hasattr(self.train_dataloader.dataset, 'gandlf_params') and
+            'patch_size' in self.train_dataloader.dataset.gandlf_params):
+            # Return the patch size from GANDLF config
+            return self.train_dataloader.dataset.gandlf_params['patch_size']
+            
+        # Default fallback value
+        return [32, 32, 32]
```
MasterSkepticista: This is very arbitrary and not the best place to hardcode. Why not have this in gandlf workspace dataloader, just like keras/mnist?

openfl-workspace/keras/mnist/src/dataloader.py
```diff
@@ -43,5 +51,40 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        return 0 if self.X_train is None else len(self.X_train)
+
+    def get_valid_data_size(self):
+        """
+        Return the size of the validation dataset.
+
+        Returns:
+            int: Size of the validation dataset or 0 if not loaded
+        """
+        return 0 if self.X_valid is None else len(self.X_valid)
```
MasterSkepticista: Aggregator does not require these methods during initialization. Only collaborator uses them. So these need not be overridden

openfl-workspace/keras/mnist/src/dataloader.py
```diff
@@ -4,31 +4,39 @@
...
-        # TODO: There needs to be a method to ask how many collaborators and
-        #  what index/rank is this collaborator.
-        # Then we have a way to automatically shard based on rank and size of
-        # collaborator list.
+        # Set default values for model initialization
+        self.X_train = None
+        self.y_train = None
+        self.X_valid = None
+        self.y_valid = None
+        self.num_classes = 10  # MNIST has 10 classes
```
MasterSkepticista: Duplicate declaration?

openfl/utilities/dataloading.py
rahulga1: yes

openfl/federated/data/loader_gandlf.py
rahulga1: This is not arbitrary, for gandlf workload, dataloader is part of federated folder instead of workspace, hence that is the same path which is defined in plan.yaml, which the generic code is referring to fetch this information. Hence I added the method in the existing dataloader.

openfl/utilities/dataloading.py
rahulga1: removed this code, as it is not being used from minimal dataloader.

openfl/utilities/dataloading.py
rahulga1: I agree on this, refactored and remove this variable.

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: Removed.

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: Taken care.

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: these were required to handle the logger statement which was printing these values, remove those those logging statement and remove these functions.

openfl/utilities/dataloading.py
MasterSkepticista: OK, so if I understand, we have two initializers: standard dataloader (for collaborators) and minimal dataloader (for aggregator)?

openfl/utilities/dataloading.py
```diff
@@ -1,72 +1,87 @@
...
+        plan (Plan): Plan object linked with the dataloader
 
     Returns:
-        data_loader (DataLoader): DataLoader instance
+        DataLoader: A minimal dataloader instance with no data loaded
+
+    Raises:
+        ValueError: If required configuration is missing or dataloader class cannot be found
     """
+    try:
```
MasterSkepticista: Can we rewrite this block without nested (and ideally without) `try...except` blocks?

MasterSkepticista (commented):
This approach is shaping up well. Are we planning to update all example workspaces accordingly in this PR?

openfl/utilities/dataloading.py
kminhta: +1

openfl-workspace/keras/mnist/src/dataloader.py
```diff
@@ -4,31 +4,39 @@
...
+
+        # If data_path is None, this is being used for model initialization only
+        if data_path is None:
```
kminhta, line 29: Is it possible for this logic to be included in the base class? This way, all subclasses will inherit it by default. We expect users to subclass and modify the workspace dataloader so reducing any "openfl-specific" steps they'd need to take would improve user experience in the long run.

If there's no clean way to do this, it's likely  it's fine to keep here as long as we document its purpose and what attributes should be initialized before/after

openfl/utilities/dataloading.py
rahulga1: Correct.

openfl/utilities/dataloading.py
rahulga1: Done.

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: Thanks Kevin for suggestion, but I see moving it to base dataloader class will make that more messy, as each data might have different fields, for example Gandlf have fields like train_csv, val_csv, train and val dataloader and fields like train_x, train_y which is defined in mnist is not relevant to Gandlf. Hence I think it may not be cleaner way.

rahulga1:
> This approach is shaping up well. Are we planning to update all example workspaces accordingly in this PR?

Yes Karan, I will be working on it next, just wanted quick confirmation on the approach with two workspace before I extend this for other workspaces.

openfl/utilities/dataloading.py
psfoley: +1

openfl-workspace/keras/jax/mnist/src/dataloader.py
```diff
@@ -40,3 +50,19 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        """
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model.
+        Returns:
+            list: The input shape for the model [28, 28, 1]
+        """
+        return [28, 28, 1]
```
tanwarsh: shape should not be hardcoded and we should return actual shape of the data.

openfl-workspace/keras/mnist/src/dataloader.py
```diff
@@ -43,5 +50,22 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model.
+
+        Returns:
+            list: The input shape for the model [28, 28, 1]
+        """
+        return [28, 28, 1]
```
tanwarsh: same here.

openfl-workspace/keras/torch/mnist/src/dataloader.py
```diff
@@ -40,3 +51,19 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        """
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model.
+        Returns:
+            list: The input shape for the model [28, 28, 1]
+        """
+        return [28, 28, 1]
```
tanwarsh: same here.

openfl-workspace/torch/mnist/src/dataloader.py
```diff
@@ -48,6 +62,22 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        """
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model.
+        Returns:
+            list: The input shape for the model [28, 28, 1]
+        """
+        return [28, 28, 1]
```
tanwarsh: same here.

openfl-workspace/torch/mnist_eden_compression/src/ptmnist_inmemory.py
```diff
@@ -48,3 +60,20 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model. This works even when no data is loaded.
+
+        Returns:
+            list: The input shape for the model [1, 28, 28]
+        """
+        return [1, 28, 28]
```
tanwarsh: same here.

openfl-workspace/torch/mnist_fed_eval/src/ptmnist_inmemory.py
```diff
@@ -48,3 +61,20 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model. This works even when no data is loaded.
+
+        Returns:
+            list: The input shape for the model [1, 28, 28]
+        """
+        return [1, 28, 28]
```
tanwarsh: same here.

openfl-workspace/torch/mnist_straggler_check/src/ptmnist_inmemory.py
```diff
@@ -47,3 +60,20 @@ def __init__(self, data_path, batch_size, **kwargs):
...
+        return 10
+
+    def get_feature_shape(self):
+        """
+        Return the input shape for the model. This works even when no data is loaded.
+
+        Returns:
+            list: The input shape for the model [1, 28, 28]
+        """
+        return [1, 28, 28]
```
tanwarsh: same here.

tanwarsh (commented):
Thanks for this PR @rahulga1. Just few minor comments.

can we move get_feature_shape to base data loader classes?

rahulga1:
> Thanks for this PR @rahulga1. Just few minor comments.
> 
> can we move get_feature_shape to base data loader classes?

@tanwarsh Please refer to the reply I provided for the same suggestion given by Kevin. In short, it will make it very messy as each dataloader have different variable, so we cant have base dataloader with all such fields. 

Also on the hardcoding comment, MNIST is very particular dataset, which has fixed size only, there cant be any other sizes, hence input shape will always be [1,28,28], similarly num of classes will always be 10 only. Hence the hardcoding is there. If user uses any other dataset, they can modify this value. 
Also, it was discussed that we want to pass the ownership of these values to dataowner, so when they are defined the workspace they can provide the relevant dataset info through these methods which can be used for initilization purpose.

openfl-workspace/keras/mnist/src/dataloader.py
kminhta: Sorry, I meant moving it to the framework specific dataloader then (i.e. `KerasDataLoader`, `PyTorchDataLoader`, etc.). This is what the workspace `dataloader.py` would ultimately subclass from. Your reasoning for not including it in the base `DataLoader` makes sense and I agree

If this is also too messy, I'd recommend to load the data if the data path is available rather than cutting the function short if it is not. IMO, this would make more intuitive sense. A developer wouldn't have to consider all the information they need to have initialized before/after the `if data_path is None... return` statement. They'd only have to think "I am going to supply all the information I need to construct the model and transform the data, if the dataset is present then load and transform it, if not then don't" WDYT? Maybe I am nitpicking

openfl-workspace/keras/jax/mnist/src/dataloader.py
```diff
@@ -10,20 +10,30 @@
...
 
         Args:
-            data_path: File path for the dataset
+            data_path: File path for the dataset. If None, initialize for model creation only.
             batch_size (int): The batch size for the data loader
             **kwargs: Additional arguments, passed to super init and load_mnist_shard
         """
         super().__init__(batch_size, **kwargs)
 
+        # Set default values for model initialization
```
kminhta: do we need to reinitialize all this information if `KerasDataLoader` already sets `self.X_train=None`, etc? This question applies for all the dataloaders that are doing this

openfl-workspace/keras/mnist/src/dataloader.py
```diff
@@ -43,5 +50,22 @@ def __init__(self, data_path, batch_size, **kwargs):
...
         self.num_classes = num_classes
+
+    def get_num_classes(self):
+        """
+        Return the number of classes for the dataset.
+
+        Returns:
+            int: Number of classes for the dataset
+        """
+        return 10
```
kminhta: The `get_num_classes()` and `get_feature_shape()` methods seem a bit disconnected here. I think we should define `self.num_classes` and `self.feature_shape` and these methods should be returning those attributes.

Also, those attributes need to be passed into the `load_dataset()` function, or otherwise used in that function. These attributes need to be used to construct the model and then transform the data to be ingestible

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: Made the changes.

openfl-workspace/keras/jax/mnist/src/dataloader.py
rahulga1: removed from all the workspaces redundant initialization.

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: I think this should be dependent on the dataset and not generic enough. For example, in this workspace, we dont need that information, even we dont use load_dataset method.

kminhta: Sorry, my comment was ambigious, by `load_data()`, I meant _any_ function or method (whether part of the dataloader or imported from a utils file) to load in the data. 

For the keras mnist, there is a function to load the data `mnist_utils.load_mnist_shard()` which hardcoded the number of classes and feature shape (https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/keras/mnist/src/mnist_utils.py#L87)

This should, instead, be using number of classes and feature shape that is assigned in `KerasMNISTInMemory`

What I was recommending was something like this (pseudo-code):
```python
...
class KerasMNISTInMemory(KerasDataLoader):
...
  self.num_classes = 10
  self.feature_shape = [1,28,28]
  _ = load_mnist_shard(self.num_classes, self.feature_shape, shard_num=int(data_path), **kwargs)
...
  def get_num_classes(self):
     return self.num_classes
  ...
  def get_feature_shape(self):
     return self.feature_shape 
...
```

or a user can access the `num_classes` and `feature_shape` directly if you want to get rid of the `get_` methods (i.e. `dataloader.num_classes`). I'm a fan of `get_` methods for better encapsulation, but the data loader may be simple enough that direct access is likely safe.

openfl-workspace/keras/mnist/src/dataloader.py
kminhta: Thanks! not sure I see the changes we were discussion, though. I may be missing something

openfl/federated/data/loader_keras.py
```diff
@@ -40,10 +40,33 @@ def __init__(self, batch_size, **kwargs):
...
+
+        Returns:
+            list: The shape of an example feature array.
+        """
+        if self.X_train is not None and len(self.X_train) > 0:
+            # Return actual shape from loaded data if available
+            return list(self.X_train[0].shape)
+        # For minimal initialization, return a default MNIST-like shape
+        # Child classes should override this method if they work with different data shapes
+        return [28, 28, 1]  # Default Keras shape: [height, width, channels]
```
psfoley: These hardcoded values for feature shape / num classes in the base class (i.e. `[28, 28, 1]` in this example) will cause unexpected behavior for models where feature shape is intended to be set dynamically. It might be better to set this to `None` to force the user to overwrite this in their examples / existing OpenFL workspace. Otherwise this could lead to the model being initialized, then seeing a mismatch during the first round of an experiment that becomes more difficult to debug.

psfoley (commented):
Thanks for taking this on, @rahulga1. There's a couple problematic areas I see:

1. Dataloaders have default values for feature shape. The base class should return None to enforce that users have defined the feature shape in their workspace. Otherwise there's a risk of shape mismatches at runtime, which will confuse users and require redistribution of the workspace.
2. Regarding the OpenFL workspace changes, feature shape should be defined for each of these. In its current form, the default value would be taken (i.e. `[28,28,1]` for keras) and would likely lead to the risk called out in point 1.

openfl-workspace/keras/mnist/src/dataloader.py
rahulga1: Aah, you are right, I didnt do it. But if its ok, can we look into this in next iteration, if its really needed?
For now, I have defined the flow in a way, user should define the required value for feature_shape and num_classes and then cut the function short.

rahulga1:
@kta-intel @psfoley Thanks for valuable feedback. Based on your inputs, here is what I have done:
1. Defined feature_shape and num_classes in all workspace dataloader
2. Removed returning mnist default values from base dataloader's get feature_shape/num_classes method, instead it returns the value with which the dataloader was initialized or read it from data if data was provided. If nothing was provided, it will raise the error.
3. Modified load_X(data/shard) functions to take feature_shape and num_classes explicitly, so it is not being defined separately in util class and there is only one place in the whole workspace where these values are defined to avoid confusion. 

 Please let me know, what do you think of these changes, also if you have any other feedback.

openfl/federated/data/loader_pt.py
```diff
@@ -46,10 +46,46 @@ def __init__(self, batch_size, random_seed=None, **kwargs):
...
         Returns:
-            tuple: The shape of an example feature array.
+            int: The number of classes.
+
+        Raises:
+            NotImplementedError: If self.num_classes is not defined.
         """
-        return self.X_train[0].shape
+        if hasattr(self, "num_classes"):
+            return self.num_classes
```
MasterSkepticista: @rahulga1 One clarification: Base classes aren't required to guess the value from attributes. They must always return `NotImplementedError`. The functions must be overridden by the data scientist in their dataloader implementation. The shapes/num classes defined by them are the ground truth and used by the dataloader to parse datasets accordingly. 

In other words: the shapes written in the subclassed dataloader determine how the dataset is read. The dataset does not determine the input/output shapes

openfl-workspace/keras/mnist/src/dataloader.py
MasterSkepticista: @psfoley @kta-intel 
I thought `get_input_shape` and `get_output_shape` is how we generalize across all kinds of models, no?

Classification is a special case where `get_output_shape` is `(num_classes,)`

openfl/federated/data/loader_keras.py
MasterSkepticista: +1
@rahulga1 
Related: https://github.com/securefederatedai/openfl/pull/1590#discussion_r2081435260

openfl/federated/data/loader_gandlf.py
MasterSkepticista: It is arbitrary, Rahul. The current implementation might be hardcoding it. There is nothing unique about `[32, 32, 32]` except for the GaNDLF workspace itself (which is where the shape needs to be defined like MNIST example). GaNDLF is a wrapper on pytorch to train classification/segmentation/detection models.

So, this would raise `NotImplementedError` in the base class

payalcha:
@rahulga1  - We noticed in this PR runs, collaborator is getting killed silently (intermittently). This issue we haven't seen in any of our previous runs. 
Since there is no error in any logs we can't be sure of the actual issue, but please look into this once.

openfl/federated/data/loader_pt.py
rahulga1: @MasterSkepticista this is conflicting suggestion what @kta-intel earlier suggested to move these common functions to base loader class, @kta-intel please confirm if my understanding is correct here?

openfl/federated/data/loader_gandlf.py
rahulga1: so, is the suggestion is to create one more subclass which inherit this class and define shape and class? isnt it overkill, given we dont have any other workspace leveraging this functionality. 
also GanDLF may be a wrapper of pytorch, but GaNDLFDataLoaderWrapper class has nothing to do with pytorch and in my opinion should have never been part of the core code in the first place.

rahulga1:
> @rahulga1 - We noticed in this PR runs, collaborator is getting killed silently (intermittently). This issue we haven't seen in any of our previous runs. Since there is no error in any logs we can't be sure of the actual issue, but please look into this once.

@payalcha found out that issue with mnist_straggler_check, there was a mismatch between number of return values by util function, hopefully, it shouldnt happen again, will monitor if it happens again.

openfl/federated/data/loader_pt.py
kminhta: > @MasterSkepticista this is conflicting suggestion what @kta-intel earlier suggested to move these common functions to base loader class, @kta-intel please confirm if my understanding is correct here?

Are you referring to https://github.com/securefederatedai/openfl/pull/1590#discussion_r2077611214?  I was actually specifically talking about the `if data_path is None: ... return` logic, but after some discussion it seemed like it would be too messy to move to the base class so I suggested the alternative approach of checking `if data_path exists then load in data` instead. I agree that the base class should not hard code any values

openfl-workspace/keras/mnist/src/dataloader.py
kminhta: @MasterSkepticista actually, we just use `get_feature_shape()`: https://github.com/securefederatedai/openfl/blob/develop/openfl/federated/data/loader.py#L26, which really just the `input_shape` and we have no generalized method / attribute for `output_shape`. 

It's good you bring it up though, because I do think `get_input_shape` and `get_output_shape` is the most intuitive generalization and this PR could be a good place to standardize that

openfl-workspace/keras/mnist/src/dataloader.py
kminhta: Sure, I think it's fine to move this ask to a separate PR since it doesn't really affect usability

openfl/federated/data/loader_gandlf.py
rahulga1: @psfoley @MasterSkepticista @kta-intel given it doesnt have any other dependency, I would suggest moving it to workspace then keeping in the base data loader class. WDYT?

@MasterSkepticista post that I will make it in the similar format as other dataloader.

openfl/federated/data/loader_pt.py
rahulga1: @MasterSkepticista can you have a look now? I kept the base dataloader clean.

openfl-workspace/keras/hippmapp3r/src/dataloader.py
```diff
@@ -34,6 +41,22 @@ def __init__(self, data_path, batch_size, **kwargs):
         self.y_train = np.asarray(y_train)
         self.y_valid = np.asarray(y_valid)
 
+    def get_feature_shape(self):
+        """Returns the shape of an example feature array.
+
+        Returns:
+            list: The shape of an example feature array [64, 64, 128] for Hippmapp3r.
+        """
+        return self.feature_shape
```
MasterSkepticista: Can we put the `[64, 64, 128]` value here to maintain consistency? If other class attributes need this, they can call `self.get_feature_shape()`.

openfl-workspace/keras/jax/mnist/src/taskrunner.py
```diff
@@ -3,12 +3,14 @@
...
 
+import logging
+
 import keras
 
-from .model import CNNModel
 from openfl.federated import KerasTaskRunner
 
-import logging
+from .model import CNNModel
```
MasterSkepticista: Please replace relative imports with absolute at the workspace level: `from src.model import CNNModel`

openfl-workspace/torch/histology/src/dataloader.py
```diff
@@ -26,35 +26,61 @@
...
+            data_path: The file path to the data. If None, initialize for model creation only.
             batch_size: The batch size of the data loader
             **kwargs: Additional arguments, passed to super init
              and load_mnist_shard
         """
         super().__init__(batch_size, random_seed=0, **kwargs)
 
+        # Set Histology-specific default attributes
+        self.feature_shape = [3, 150, 150]  # Histology image shape for PyTorch (channels first)
+        self.num_classes = 8  # Histology has 8 classes
```
MasterSkepticista: Inline comments not necessary here (applies to all similar instances)

MasterSkepticista (approved):
Thanks @rahulga1. I am satisfied with this first iteration towards generalizing dataloader initialization. Approving this PR (modulo minor changes commented below).

tests/end_to_end/utils/federation_helper.py
```diff
@@ -228,24 +228,21 @@ def run_federation(fed_obj):
     Returns:
         bool: True if successful, else False
     """
-    executor = concurrent.futures.ThreadPoolExecutor()
+    concurrent.futures.ThreadPoolExecutor()
```
noopurintel: This can be removed.

tests/end_to_end/utils/federation_helper.py
```diff
@@ -228,24 +228,21 @@ def run_federation(fed_obj):
...
         _ = set_keras_backend(fed_obj.model_name)
 
-    # As the collaborators will wait for aggregator to start, we need to start them in parallel.
-    futures = [
-        executor.submit(
-            participant.start
-        )
-        for participant in [fed_obj.aggregator] + fed_obj.collaborators
-    ]
+    # As the collaborators will wait for aggregator to start, we need to start them in sequence.
```
noopurintel: This comment can be removed.

noopurintel (approved):
Approving with two minor comments.

tests/end_to_end/utils/federation_helper.py
rahulga1: Modified

tests/end_to_end/utils/federation_helper.py
rahulga1: Done

payalcha:
> @rahulga1 - We noticed in this PR runs, collaborator is getting killed silently (intermittently). This issue we haven't seen in any of our previous runs. Since there is no error in any logs we can't be sure of the actual issue, but please look into this once.

Post these changes, data is getting downloaded during collaborator start. Since in our test we were starting collaborator parallelly it was giving issue. Now we made it one after another and the issue is solved.

tanwarsh (approved):
LGTM

openfl/federated/data/loader_gandlf.py
theakshaypant: Maybe we can raise an error is a shape is not determined by data loader or no shape is provided instead of returning `[32, 32, 32]`?

rahulga1: @theakshaypant @MasterSkepticista I think thats what you were suggesting.

theakshaypant (approved):

PR #1592, opened 2025-05-06
[Task Runner API] Collaborators should only send optimizer information after training when `self.opt_treatment== "CONTINUE_GLOBAL"`

kminhta:
## Summary
The expected behavior is to send only model updates between the agg and col unless `self.opt_treatment== "CONTINUE_GLOBAL"` (which can configured in the `plan.yaml`). The Task Runners were sending optimizer states keys by default after training. Not only is this extra information that doesn't need to be sent, if a user wants to run model integrity check by hashing the values sent and received, they would have to filter out the opt values. Furthermore, keras workspaces have an additional `update_tensorkeys_for_functions()` method that will automatically update the tensorkeys to include optimizer state keys, which seems to be a less conditioned version of: 
```
if self.opt_treatment == "CONTINUE_GLOBAL":
            self.initialize_tensorkeys_for_functions(with_opt_vars=True)
```

## Type of Change (Mandatory)
- Bug fix  

## Description (Mandatory)
This PR:
- Sets `self.get_tensor_dict(with_opt_vars=(self.opt_treatment == "CONTINUE_GLOBAL"))` for all `runner_*.py` during the train task so that optimizer key states are only send if explicitly set to
-  Removes `update_tensorkeys_for_functions()` method from `runner_keras.py` as this method is redundant when compared against `self.initialize_tensorkeys_for_functions(with_opt_vars=True)` and is also adding optimizer key states to the list of tensor keys without a check.

## Testing
Manually

## Additional Information
-

### DIFF
--- a/openfl/federated/task/runner_gandlf.py
+++ b/openfl/federated/task/runner_gandlf.py
@@ -223,7 +223,7 @@ class GaNDLFTaskRunner(TaskRunner):
             )
 
         # output model tensors (Doesn't include TensorKey)
-        tensor_dict = self.get_tensor_dict(with_opt_vars=True)
+        tensor_dict = self.get_tensor_dict(with_opt_vars=(self.opt_treatment == "CONTINUE_GLOBAL"))
 
         metric_dict = {"loss": epoch_train_loss}
         for k, v in epoch_train_metric.items():
--- a/openfl/federated/task/runner_keras.py
+++ b/openfl/federated/task/runner_keras.py
@@ -120,7 +120,9 @@ class KerasTaskRunner(TaskRunner):
         }
 
         # output model tensors (Doesn't include TensorKey)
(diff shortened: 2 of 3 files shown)


MasterSkepticista (approved):
Are these changes sufficient, or do we need to modify how optimizer states are sent during `_get_weights_dict`?

If not, LGTM

kminhta:
> Are these changes sufficient, or do we need to modify how optimizer states are sent during `_get_weights_dict`?
> 
> If not, LGTM

As of right now, these changes are sufficient to prevent optimizer information from being automatically sent after training. I manually checked the case when we do want optimizers to be sent (`opt_vars=True`) and `_get_weights_dict` looks to be parsing it correctly. Possible we'll need to revisit it if keras (or torch) change their structure, but it looks good for now.

image

PR #1593, opened 2025-05-06
Scaling issue with tensor_db

porteratzo:
**Pull Request Title:** Optimize TensorDB Query Performance in OpenFL Aggregator

**Description:**

This PR addresses a performance issue in OpenFL that causes a slowdown as training progresses, particularly when scaling with more collaborators. The slowdown was identified through profiling, which revealed that the `get_aggregated_tensor` function in the aggregator was taking increasingly more time due to the `get_tensor_from_cache` operation. This operation queries the `tensor_db` pandas dataframe, which grows larger over time as it records multiple metrics from each collaborator. This growth results in progressively slower queries.

**Figures:**

1. **Aggregator Times Comparison:**
image
   - *Caption:* The figure above shows the comparison of aggregator step times, with the new implementation above and the old below. We observe a mean reduction of 16.4 seconds in round time, primarily due to the reduction in `get_aggregated_tensor` time (15.6s).

2. **Total Time Between Rounds:**
image
   - *Caption:* The figure above illustrates the reduction in total time between rounds across the entire experiment. Despite the improvement, a linear increment in time persists in the new experiment, warranting further investigation.

3. **Collaborator TensorDB Benchmark:**
image
   - *Caption:* The figure above displays the benchmark results for collaborators, showing an overall improvement in speed attributed to faster `get_aggregated_tensor` calls, particularly in the `aggregated_model_validation` where model tensors are first obtained (~19s).

4. **Get Cached Tensor Collaborator:**
image

   - *Caption:* The figure above compares the old and new implementations of `get_cached_tensor` on the collaborator, showing a minor per-round speedup (~0.9s) due to the dropping of metrics.

5. **Get Cached Tensor Aggregator:**
image

   - *Caption:* The figure above highlights the significant speedup (82%) in the aggregator's `get_cached_tensor` function, achieved by drastically reducing the query size by ignoring collaborator and metric keys.

6. **Cleanup Operation Impact:**
image
image

   - *Caption:* The figure above demonstrates the negligible impact of the cleanup operation on both the aggregator and collaborator, as it is only run once per round.

7. **Cache Tensor Regression:**
image
image

   - *Caption:* The figure above illustrates the non-negligible regression in `cache_tensor` due to the overhead of filtering out metrics and collaborator keys for `get_cache_tensor`. Despite the 4.5s slowdown, the speedup in `get_cached_tensor` (15.9s) compensates for this regression.

**Future Work:**

Further optimization is needed for `get_cached_tensor` and `cache_tensor`, as they are called over 4000 times per round. Consolidating these calls could lead to additional performance improvements. Also there is still linear increments in round time even with the new fixes which need further investigation.


openfl/databases/tensor_db.py
```diff
@@ -93,6 +94,13 @@ def clean_up(self, remove_older_than: int = 1) -> None:
             (self.tensor_db["round"].astype(int) > current_round - remove_older_than)
             | self.tensor_db["report"]
         ].reset_index(drop=True)
+        self.secondary_db = self.tensor_db[
+            ~self.tensor_db["tags"].apply(
+                lambda x: any(
+                    keyword in item for item in x for keyword in ["collaborator", "metric"]
+                )
+            )
+        ].reset_index(drop=True)
```
MasterSkepticista, line 103: As I understand, this snippet removes collaborator metrics on the aggregator side (which are merely a couple of floats for each collaborator). I am not sure if this is related to scaling issues, or memory usage in general. Do you have any pre/post analysis in support of this proposal?

porteratzo: Well as evidence I have profiled the get_cached_tensor with and without filtering these rows and got an ~80% speed up image

As an example I have this calculation
image

So the collaborator tensor rows grows a lot with more collaborators, this number of rows remains constant through the experiment and though the metrics increase relatively little per round, given many reported metrics and many collaborators and rounds these can grow to be a significant part of the tensor_db.

Most calls to get_tensor_from_cache only really care about the aggregator tensor so removing all the collaborator model rows and metric rows does make a large improvement to query time

MasterSkepticista: Thanks, two experiment suggestions:
1. Could you enable memory profiling and check the memory trends with/without proposed changes?
2. What is the time taken for 10 rounds of histology (assuming 2 collaborators) with/without proposed changes?

If either of these two results are significant improvements, we can prioritize merging this PR

porteratzo: The changes are most impactful with more collaborators and also depend on model size so the following experiment was run on a custom modified workspace from MNIST with 30 collaborators using resnet50 for 10 rounds.

Without changes
real    39m51.381s

With changes
real    34m12.750s

so around 13% reduction in time to completion.

porteratzo:
https://github.com/securefederatedai/openfl/pull/1575 made major changes to function calls optimized here, possibly conflicting with these changes, once that is merged more profiling will be done to see the effect

PR #1595, opened 2025-05-07
Add `on_task_*` callbacks

MasterSkepticista:
This PR generalizes existing support for callbacks at the beginning and end of a task. Task callbacks now take `task_name` as a positional argument, allowing downstream implementations to trigger functions based on the task.


bsenthilr (approved):
LGTM

kminhta (approved):

PR #1596, opened 2025-05-07
[Task Runner API] enable saving native model via callback at the aggregator

kminhta:
## Summary
Currently, models are saved as `.pbuf` file by default. A developer will inevitably need to convert this back to native format. `fx model save` accomplishes this, but it requires an extra step as well as direct access to the Task Runner. Saving the model in native format during training will remove this extra step and requirement.

This PR adds this functionality and can be set in the `plan.yaml` file under the aggregator settings

```
aggregator :
  settings :
    init_state_path : save/init.pbuf
    rounds_to_train : 2
    save_native_model : True
```

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This PR:
- introduces new default callback for aggregator called `ModelSaver()` which will save the native model at the end of the experiment. It uses similar logic as `fx model save`
- updates aggregator to append the proper params (`last_state_path` and `best_state_path`) when `save_native_model` is set to `True`
- updates the `save_native()` methods in the respective Task Runners in order to automatically add file extension if none is provided (which is the case when saving at the aggregator since the aggregator is framework agnostic) and to return the filename for logging

## Testing
Manual testing:
image

## Additional Information
- We need more elegant handling of optimizer states. Currently, the default optimizer treatment (`opt_treatment`) is to `RESET` and/or `CONTINUE_LOCAL` (i.e. collaborators do not send or retrieve optimizer results). This leaves a minor gap at the aggregator side where they receive a final aggregated model but not the aggregated optimizer state, so it is technically an incomplete checkpoint (though not the worst case since they can still use this model for inference without loss of performance or for further training with the understanding that the optimizer gets re-initialized).
   - optimizer treatment is actually a collaborator setting and not a task runner setting, so the task runner that gets initialized at the aggregator to load and save the model also doesn't have this attribute (the collaborator sets it). currently it is safer to just save out the optimizer even if it is technically the initial state
   - blindly sending the optimizer states to the aggregator affects model integrity checks for downstream applications unless it also includes some sort of tensor key check, which is why the collaborator only sends the optimizer states when `opt_treatment` is set to `CONTINUE_GLOBAL`


openfl/callbacks/model_saver.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2020-2024 Intel Corporation
```
teoparvanov: nit - 2025, as this is a newly added file

openfl/component/aggregator/aggregator.py
```diff
@@ -87,6 +87,7 @@ def __init__(
         initial_tensor_dict=None,
         log_memory_usage=False,
         write_logs=False,
+        save_native_model=False,
```
teoparvanov, line 91: Maybe worth documenting in the Aggregator section of the [FL plan docs](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html#configurable-settings)?

teoparvanov (approved):
Looks great, thanks @kta-intel !

openfl/callbacks/model_saver.py
```diff
@@ -0,0 +1,85 @@
...
+from openfl.callbacks.callback import Callback
+from openfl.federated import Plan
+from openfl.protocols import utils
+from openfl.utilities.dataloading import get_dataloader
+from openfl.utilities.workspace import set_directory
+
+logger = logging.getLogger(__name__)
+
+
+class ModelSaver(Callback):
```
MasterSkepticista: Suggest renaming this to `ModelCheckpoint`

openfl/federated/task/runner_pt.py
```diff
@@ -460,7 +461,11 @@ def save_native(
             model_state_dict_key: self.state_dict(),
             optimizer_state_dict_key: self.optimizer.state_dict(),
         }
+        if "." not in str(filepath).split("/")[-1]:
```
MasterSkepticista: Use `os.sep` instead of `/` to ensure Windows doesn't break...

openfl/federated/task/runner_keras.py
```diff
@@ -271,8 +271,13 @@ def save_native(self, filepath):
 
         Args:
             filepath (str): The file path to save the model.
+                By default, model will be saved as `*.keras`
         """
+        if "." not in str(filepath).split("/")[-1]:
```
MasterSkepticista: `os.sep`

openfl/component/aggregator/aggregator.py
```diff
@@ -194,14 +195,21 @@ def __init__(
                 on_round_end=lambda round_num, logs=None: self.save_analytics_result()
             )
         )
+        # Prepare the parameters dictionary
+        callback_params = {"origin": "aggregator"}
+
+        if save_native_model:
+            callback_params["last_state_path"] = self.last_state_path
+            callback_params["best_state_path"] = self.best_state_path
+
```
MasterSkepticista: I recommend creating a `ModelCheckpoint` instance here, in code, and pass the checkpoint save paths as arguments to the constructor.

```python
ckpt_callback = ModelCheckpoint(root_dir="/path/to/save", taskrunner=taskrunner_object, ...)
callbacks.append(ckpt_callback)
...
self.callbacks = CallbackList(
  callbacks, 
  add_memory_profiler=log_memory_usage,
  add_metric_writer=write_logs,
  ...
)
```

The `add_*` arguments are for backward compatibility with plan flags, and not how we intend to add future callbacks

openfl/component/aggregator/aggregator.py
MasterSkepticista, line 91: Is there any harm _not_ keeping it user-configurable through the plan? Two options:

1. It can always be true. Models are always saved in native format.
2. It can be "inferred" based on `init_path` or the likes provided in the arguments.

MasterSkepticista (changes requested):
Comments mostly around where best to place considering code maintainability...

openfl/callbacks/model_saver.py
MasterSkepticista: Additional context: Initially, I planned to implement `ModelCheckpoint` callback which would eliminate `self._save_model` from aggregator code entirely. It was deprioritized for implementation challenges. It is required however...

Since this PR is a step in the direction to save model only at the end of experiment, we will eventually use the same class to save "buffer" checkpoints that `self._save_model` does in protobuf format.

Overall, this class may just be saving native model at the end of experiment today. In the future, this will also handle checkpoint management, for all complex kinds of FL experiments.

openfl/component/aggregator/aggregator.py
kminhta: Agreed, added

openfl/callbacks/model_saver.py
kminhta: Makes sense, done. Kept the docstring currently focused on save native model, but we can edit that as we adapt it

openfl/component/aggregator/aggregator.py
kminhta: Oh, thanks for the clarity. This is a cleaner way to do this. Done

openfl/component/aggregator/aggregator.py
kminhta: Let me get back to you on this. It was originally requested to be configurable in this iteration, but I don't see any reason why it shouldn't just be defaulted to one of those options instead of being use-configurable.

openfl/callbacks/callback_list.py
```diff
@@ -52,6 +52,7 @@ def _add_default_callbacks(self, add_memory_profiler, add_metric_writer):
         """Add default callbacks to callbacks list if not already present."""
         self._memory_profiler = None
         self._metric_writer = None
+        self._model_saver = None
```
MasterSkepticista: Since we are now appending callbacks during init, this may no longer be required

docs/about/features_index/taskrunner.rst
```diff
@@ -47,6 +47,7 @@ Configurable Settings
  - :code:`write_logs`: (boolean) Metric logging callback feature. By default, logging is done through `tensorboard <https://www.tensorflow.org/tensorboard/get_started>`_ but users can also use custom metric logging function for each task.     
  - :code:`persist_checkpoint`: (boolean) Specifies whether to enable the storage of a persistent checkpoint in non-volatile storage for recovery purposes. When enabled, the aggregator will restore its state to what it was prior to the restart, ensuring continuity after a restart. 
  - :code:`persistent_db_path`: (str:path) Defines the persisted database path. 
+ - :code:`save_native_model`: (boolean): Specifies whether to save the model in its native format. This is useful for post-experiment analysis and deployment. The model will be saved in the base path specified by the :code:`last_state_path` and :code:`best_state_path` setting.
```
MasterSkepticista: "Only applicable for tasks involving federated training, this is ignored for federated evaluation. If set, the model is saved in the framework native format at the end of the experiment. Saving methods are defined for each framework under `save_native` function of the respective [supported](https://github.com/securefederatedai/openfl/tree/develop/openfl/federated/task) framework plugins."

P.S.: Please check for no side-effects of this callback in evaluation mode.

MasterSkepticista (changes requested):

docs/about/features_index/taskrunner.rst
kminhta: > Please check for no side-effects of this callback in evaluation mode.

Oh, this is a good foresight, let me check that

edit: added conditional to skip adding this callback if in evaluation mode

openfl/callbacks/callback_list.py
kminhta: missed it - good catch :)

openfl/component/aggregator/aggregator.py
```diff
@@ -194,6 +196,20 @@ def __init__(
...
+                ckpt_callback = callbacks_module.ModelCheckpoint(
+                    self.last_state_path,
+                    self.best_state_path,
+                )
+                callbacks.append(ckpt_callback)
+            else:
+                logger.info(
+                    "Saving native model is disabled in evaluation mode. "
+                    "Skipping ModelCheckpoint callback."
+                )
```
MasterSkepticista: Nit: Save lines!

```python
if save_native_model and not self.assigner.is_task_group_evaluation():
    ckpt_callback = callbacks_module.ModelCheckpoint(...)
    callbacks.append(ckpt_callback)

# contd
```

Log message not required as a user would know that evaluation doesn't generate any artifacts.

MasterSkepticista (approved):

openfl/component/aggregator/aggregator.py
kminhta: that's how I had it before https://github.com/securefederatedai/openfl/pull/1596/commits/3a79d4c3daf04553285db454209707d8072c9c1c. Always okay with saving lines :)

PR #1585, opened 2025-05-01
[Task Runner API] [Flower Interoperability] Integrate changes for `flower-app-pytorch` workspace's custom `Aggregator` into core component

kminhta:
## Summary
`flower-app-pytorch` workspace was originally written to be as self-contained as possible, this included its own custom aggregator that subclassed the core `Aggregator`. This PR cleans up and pulls these features back into the core `Aggregator`

This is part of a larger effort to pull custom workspace code back to the core OpenFL for better maintainability and for more native support of Flower workloads. This is being done piecewise as they are not interdependent and it'll be easier to track and control individual component changes.

## Type of Change (Mandatory)
Specify the type of change being made. 
- Feature enhancement  
- Refactor

## Description (Mandatory)
This PR:
- removes `openfl-workspace/flower-app-pytorch/src/aggregator.py`
- adds `connector input and `self.connector` attribute to the core `Aggregator`
- adds conditionals related to `self.connector` in order to:
   - defining `self.model` as an empty dict since we will allow Flower apps handle the "inner" experiment at this stage
   - skip processing task results (same reason as above)
   - skip computing validation metrics (same reason as above)
   - skip saving model (same reason as above)
 - clean up connector related logic at the aggregator server to reduce lines of code (the original custom aggregator has additional connector methods (i.e. 'is_connector_available()` that were deemed superfluous)

## Testing
Manual testing

PR #1598, opened 2025-05-08
[Task Runner API] [Flower Interoperability] Integrate changes for flower-app-pytorch workspace's custom Collaborator into core component 

kminhta:
## Summary
Part 2/x of bringing flower-app-pytorch custom components back into core components

- [x]  https://github.com/securefederatedai/openfl/pull/1585
- [ ] `openfl.component.Collaborator`
...
...

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
This PR:
- removes `src.collaborator.CollaboratorFlower` and pulls changes back into `openfl.component.Collaborator`
- modifies the task name from `start_client_adapter` to `interop` for better generalizability
- adds `prepare_interop_server()` callback to `Collaborator` in order to minimize changes needed in `do_task()`. Originally, there was a separate conditional that started a separate set of tasks specific to interopability, which made it harder to read. Instead, we now pass the module to import (`src.grpc.connector.flower.interop_server`) via `plan.task` setting. This module is imported and initialized and sent to the Task Runner in `kwargs` at the beginning of the experiment
- `flower.interop_server` arg is modified to accept a callable function (`receive_message_from_interop()`, which is defined in `prepare_interop_server()`. The reason for this change is that originally, we send the entire client to the `interop_server()` and the `interop_server()` called `self.client.send_message_to_server()` directly. It better encapsulates the client and the collaborator for the `interop_server()` to pass just the response to a callable function of the `Collaborator`

## Testing
Manual testing in non-SGX and SGX


openfl-workspace/flower-app-pytorch/plan/plan.yaml
```diff
@@ -52,12 +52,12 @@ assigner :
       - name  : Connector_Flower
         percentage : 1.0
         tasks : 
-          - start_client_adapter
+          - interop
```
teoparvanov: I'd suggest the more explicit `prepare_for_interop` (inspired by the underlying python method name)

openfl-workspace/flower-app-pytorch/src/runner.py
```diff
@@ -52,20 +52,30 @@ def __init__(self, **kwargs):
...
         Args:
-            interop_server: The local gRPC server instance.
-            **kwargs: Additional parameters, including 'local_server_port'.
+            col_name (str, optional): The collaborator name. Defaults to None.
+            round_num (int, optional): The current round number. Defaults to None.
+            input_tensor_dict (dict, optional): The input tensor dictionary. Defaults to None.
+            **kwargs: Additional parameters for configuration.
+                includes:
+                    interop_server (object): The FlowerInteropServer instance.
+                    local_server_port (int): The port for the local server.
```
teoparvanov: For the sake of symmetry, shouldn't this be called `interop_server_port`?

openfl-workspace/flower-app-pytorch/src/runner.py
```diff
@@ -52,20 +52,30 @@ def __init__(self, **kwargs):
...
         """
         local_server_port = kwargs.get('local_server_port')
```
teoparvanov: I suggest moving this a couple of lines below, so it reads:
```python
interop_server = kwargs.get('interop_server')
local_server_port = kwargs.get('local_server_port')
```
or, if you like my previous comment:
```python
interop_server = kwargs.get('interop_server')
interop_server_port = kwargs.get('interop_server_port')
```

openfl-workspace/flower-app-pytorch/src/runner.py
```diff
@@ -117,6 +127,8 @@ def message_callback():
                 interop_server.stop_server()
             time.sleep(0.1)
 
+        return {}, {}
```
teoparvanov: Is this method expected to return something?

openfl/component/collaborator/collaborator.py
```diff
@@ -246,6 +247,9 @@ def do_task(self, task, round_number) -> dict:
         func = getattr(self.task_runner, func_name)
         logger.debug("Using TaskRunner subclassing API")
 
+        if task_name == "interop":
+            kwargs = self.prepare_interop_server(kwargs)
```
teoparvanov: This looks a bit out of place here... Also, I guess the idea is that it's done only once, at the start of the experiment?
If so, then could the same be achieved using a callback, similar to how we initialize secure aggregation "on_experiment_begin"?
https://github.com/securefederatedai/openfl/blob/develop/openfl/callbacks/secure_aggregation.py#L21

teoparvanov (commented):
A couple of thoughts from my first read-through:

openfl/component/collaborator/collaborator.py
kminhta: Oh, this is a good idea! let me think about it then update the PR

openfl-workspace/flower-app-pytorch/src/runner.py
kminhta: So it needed to return empty dict so that I didn't have to change the collaborator logic: https://github.com/securefederatedai/openfl/blob/develop/openfl/component/collaborator/collaborator.py#L249

Bu  I do see that is looks a bit jarring so let me think of a cleaner way to do this

openfl-workspace/flower-app-pytorch/src/runner.py
kminhta: yes, agreed!

openfl/component/collaborator/collaborator.py
```diff
@@ -143,6 +145,15 @@ def __init__(
             else:
                 callbacks = [secure_aggregation_callback]
 
+        # Interoperability mode
+        self._interop_mode_enabled = interop_mode
+        if self._interop_mode_enabled:
+            callbacks.append(
+                callbacks_module.LambdaCallback(
+                    on_round_begin=lambda round_num, logs=None: self.prepare_interop_server()
```
teoparvanov: Shouldn't this be `on_experiment_begin` ?

teoparvanov (approved):
Awesome, thanks for a crystal-clear implementation, @kta-intel !

openfl/component/collaborator/collaborator.py
kminhta: Actually yes, I'm glad you brought this up. 

It's funny - I originally had it `on_experiment_begin` then moved it in to `on_round_begin` in https://github.com/securefederatedai/openfl/pull/1598/commits/81035c35488483730f049eeb968351b9386ee14a since the Task Runner shuts off the interop server after the round is complete. My mindset at the time was to bookend the start and stop of the interop server within a single round... but I've now realized that this callback just prepares it for the Task Runner which handles the start and stop on its own anyway. Sometimes first instincts are usually the best :facepalm:

openfl-workspace/workspace/plan/defaults/tasks_connector.yaml
```diff
@@ -1,4 +1,4 @@
-start_client_adapter:
+prepare_for_interop:
   function : start_client_adapter
   kwargs   :
-    local_server_port : 0 # local grpc server, 0 to dynamically allocate
+    interop_server_port : 0 # interop server port, 0 to dynamically allocate
```
kminhta, line 4: Working on addressing making dynamic allocation of gRPC ports consistent across OpenFL in https://github.com/securefederatedai/openfl/pull/1599

The issue with this particular case is that for a local simulation, we need the collaborators to have different ports but hashing the plan would result in the same port which will lead to binding issues. I'm considering adding a `local_simulation` flag where the `Collaborator` can dynamically allocate based on its common name during the `_prepare_interop_server` - or the `Task Runner` can also handle this which may be a cleaner delineation from the `Collaborator`

It's the actually same reason for this messy handling of the `client_port` too: https://github.com/securefederatedai/openfl/blob/develop/openfl-workspace/flower-app-pytorch/src/runner.py#L51

PR #1602, opened 2025-05-09
Fix keras model initialization after restart

psfoley:
## Summary
Fixes keras model initialization after collaborator restart corner case

## Type of Change (Mandatory)
- Bug fix  

## Description (Mandatory)
If the collaborator restarts after `aggregated_model_validation`, the initial model will be rebuilt. The collaborator would then receive the `train` task, and might not have the right metrics returned by the `get_metrics_result` function. This PR will run evaluation on a single dataset prior to training if this corner case occurs.

## Testing
With the fix:
image

Without the fix:
image


openfl/federated/task/runner_keras.py
```diff
@@ -169,6 +169,15 @@ def train_task(
 
         return global_tensor_dict, local_tensor_dict
 
+    def _initialize_metrics_result(self,kwargs):
+        if "batch_size" in kwargs:
+            batch_size = kwargs["batch_size"]
+        else:
+            batch_size = 1
```
MasterSkepticista: Simplification: `batch_size = kwargs.get("batch_size", 1)`

openfl/federated/task/runner_keras.py
```diff
@@ -189,13 +198,10 @@ def train_(self, batch_generator, metrics: list = None, **kwargs):
         #  defined) then the model must be recompiled.
         try:
             results = self.model.get_metrics_result()
+            if len(results) == 0:
+                results = self._initialize_metrics_result(kwargs)
```
MasterSkepticista: Looking at the def, `_initialize_metrics_result` only requires `batch_size` from `kwargs`.

```python
batch_size = kwargs.get("batch_size", 1)
results = self._initialize_metrics_result(batch_size)  # Need not pass kwargs
```

openfl/federated/task/runner_keras.py
```diff
@@ -189,13 +198,10 @@ def train_(self, batch_generator, metrics: list = None, **kwargs):
...
         except ValueError:
-            if "batch_size" in kwargs:
-                batch_size = kwargs["batch_size"]
-            else:
-                batch_size = 1
-            # evaluation needed before metrics can be resolved
-            self.model.evaluate(self.data_loader.get_valid_loader(batch_size), verbose=1)
+            self._initialize_metrics_result(kwargs)
```
MasterSkepticista: `kwargs` can be replaced with `batch_size`, as per prev. comment

MasterSkepticista (changes requested):

kminhta (approved):

openfl/federated/task/runner_keras.py
psfoley: Thanks for the suggestion - fixed

openfl/federated/task/runner_keras.py
psfoley: Fixed in latest commit

openfl/federated/task/runner_keras.py
psfoley: Fixed in latest commit

MasterSkepticista (approved):
Formatter check is failing, could you run that before merge?

psfoley:
> Formatter check is failing, could you run that before merge?

Thanks for catching this - done!

PR #1603, opened 2025-05-11
Datasets: support Azure Blob data sources

Efrat1:
## Summary
Support Azure blob data sources.

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
Support Azure blob data sources, including needed changes in `verifiableDatasetInfo`, serializing and deserializing to JSON (v1 and v2).

## Testing
Add unittests for the new class, with `verifiableDatasetInfo` and with `VerifiableMapStyleDataset`/`VerifiableImageFolder`.
Test with OpenFl-Security using Azurite.

### DIFF
--- a/openfl/federated/data/sources/azure_blob_data_source.py
+++ b/openfl/federated/data/sources/azure_blob_data_source.py
@@ -1,0 +1,60 @@
+# Copyright 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import hashlib
+from typing import Callable
+
+from azure.storage.blob import BlobServiceClient
+
+from openfl.federated.data.sources.data_source import DataSource, DataSourceType
+
+
+class AzureBlobDataSource(DataSource):
+    """Class for Azure Blob data"""
+
+    def __init__(
+        self,
+        connection_string: str,
+        container_name: str,
+        folder_prefix="",
+        hash_func: Callable[..., "hashlib._Hash"] = hashlib.sha384,
+    ):
+        super().__init__(DataSourceType.AZURE_BLOB)
+        self.connection_string = connection_string
+        self.container_name = container_name
+        self.folder_prefix = folder_prefix
+        if not super().is_valid_hash_function(hash_func):
+            raise ValueError(
+                f"Data source {self.name}: Invalid hash function: {hash_func.__name__}."
+                " Must be a hashlib function."
+            )
+        self.hash_func = hash_func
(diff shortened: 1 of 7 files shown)


setup_azure_blob_containers.py
```diff
@@ -0,0 +1,50 @@
+import sys
```
payalcha: Please add copyright info.

setup_azure_blob_containers.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+from azure.storage.blob import BlobServiceClient
+
+# Azurite connection string
+AZURITE_CONNECTION_STRING = (
+    "DefaultEndpointsProtocol=http;"
+    "AccountName=devstoreaccount1;"
+    "AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;"
```
payalcha: Please don't add AccountKey and account details here. Add this in Github Secrets.

setup_azure_blob_containers.py
Efrat1: Thanks, done

setup_azure_blob_containers.py
Efrat1: The AccountKey in that script is of Azurite (an emulator for azure blobs), so the key in there is not really secure, it is the same key for all Azurite users

setup_azure_blob_containers.py
teoparvanov, line 1: The unit tests under `/tests/openfl/federated/data` are sufficient for the purposes of the PR itself.

Therefore, I suggest removing the `setup_azure_blob_containers.py` script from the PR, and instead sharing it offline with @payalcha, as it may be useful for E2E testing.

teoparvanov (approved):
LGTM!

setup_azure_blob_containers.py
teoparvanov: +1 for @payalcha's comment. See my suggestion below.

payalcha:
@Efrat1 - we added Azure and S3 data sources, but there is no documentation update for any Verifiable dataset. Can you please add documentation and README.md so that one can understand how to use the same. 
Thanks.

setup_azure_blob_containers.py
Efrat1: OK, I removed the file. @payalcha - I'll share it with you if you need it

PR #1634, opened 2025-05-19
Fix fedprox MNIST tutorial for pytorch

### LINKED ISSUE #823: FedProx: getting same model depsite different mu values
Hi, I'm using the FedProx optimizer and following the PyTorch MNIST demos. I passed different mu values (I tried 0, 0.001, 0.01, 0.1, 1, 2, 5, 10) but still got the same trained client and server models for the same seed (but different mu's). 

In the original paper, mu should control the degree of personalization, where mu=0 is equivalent to FedAvg. In this implementation, how does mu affect the model training? Could you point me to the relevant files in which mu is used in optimization?   

I found Line 93 in openfl/utilities/optimizers/torch/fedprox.py, `d_p.add_(p - w_old_p, alpha=mu)`. I verified that my mu values had not been overwritten and were different. Other than that, I couldn't find anything else that directly uses mu in optimization.

Thanks!

nisha987:
## Summary
Models trained with different mu values in the FedProx optimizer appear to have identical weights, despite the expectation that different mu values should produce different models. This suggests that the proximal term in the FedProx algorithm is not being applied correctly.

## Type of Change (Mandatory)
- Bug fix  : https://github.com/securefederatedai/openfl/issues/823

### 1. Incorrect Timing of the `set_old_weights` Call

In the current notebook implementation, `set_old_weights` is called after gradient computation but right before the optimizer step. This is problematic because:
- It sets the reference weights (`w_old`) to be nearly identical to the current weights
- When the optimizer step is executed, the proximal term `mu * (param - w_old_param)` becomes effectively zero
- This nullifies the effect of different mu values, resulting in identical training outcomes

### 2. Missing Initial `w_old` Value in Optimizer

The optimizers (`FedProxOptimizer` and `FedProxAdam`) don't initialize the `w_old` parameter in their constructors. If `step()` is called before `set_old_weights`, there may be an error accessing the uninitialized `w_old`
.
## Testing
- Tested locally.
- 
Screenshot 2025-05-19 113316


openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -115,7 +116,11 @@ def step(self, closure=None):
             nesterov = group["nesterov"]
             mu = group["mu"]
             w_old = group["w_old"]
-            for p, w_old_p in zip(group["params"], w_old):
+            
+            # Skip FedProx regularization if w_old is not set or mu is 0
+            apply_proximal = w_old is not None and mu > 0
```
kminhta: IMO, `FedProx` should always attempt to apply the proximal term. 

If a user sets `mu==0` regardless of `w_old`, it'll mathematically cancel out the proximal term anyway (`mu*(p-w_old_p)`). 

`w_old is not None and mu > 0` should be the expected default condition if a user is trying to use FedProx

If `w_old is None and mu > 0`, we should raise an error or warning that the user needs to set the old weights prior to training. This condition indicates that a user wanted to use FedProx so just omitting the proximal term if the user forgets to set the old weights could lead to a silent bug for them. 

We also probably shouldn't let m < 0, though technically it isn't mathematically wrong, it's just counterintuitive since it'll reward deviations (so, maybe a warning)

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -75,6 +75,7 @@ def __init__(
             "mu": mu,
```
kminhta, line 90: General: we should probably rename `FedProxOptimizer` to `FedProxSGD` to maintain naming convention, but maybe we can save that for a future PR

kminhta (changes requested):
Good catch @nisha987 ! Thanks for taking up that long standing issue

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -75,6 +75,7 @@ def __init__(
...
             "nesterov": nesterov,
             "weight_decay": weight_decay,
+            "w_old": None,  # Initialize w_old as None
```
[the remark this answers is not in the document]
nisha987: Added more detailed docstring and comments.

openfl/utilities/optimizers/torch/fedprox.py
nisha987: I have added suggested conditions to fedprox.py.

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -115,34 +172,53 @@ def step(self, closure=None):
             nesterov = group["nesterov"]
             mu = group["mu"]
             w_old = group["w_old"]
-            for p, w_old_p in zip(group["params"], w_old):
+
+            # Validate old weights for FedProx
+            self._validate_old_weights(mu, w_old)
+
+            # Apply proximal term when mu != 0
+            apply_proximal = w_old is not None and mu != 0
```
kminhta: Can you remove `and mu != 0` from this conditional so that we only check for `w_old is not None`?
It is more accurate to let `mu=0` cancel out the proximal term mathematically

kminhta:
Thanks @nisha987 This is looking good. I had one more minor comment regarding `mu`
Can you also rebase and signoff on your commits? https://github.com/securefederatedai/openfl/pull/1634/checks?check_run_id=42564995654

openfl/utilities/optimizers/torch/fedprox.py
nisha987: fixed. modified the condition.

nisha987:
> Thanks @nisha987 This is looking good. I had one more minor comment regarding `mu` Can you also rebase and signoff on your commits? https://github.com/securefederatedai/openfl/pull/1634/checks?check_run_id=42564995654

added signoff to all commits.

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -67,14 +75,22 @@ def __init__(
         if weight_decay < 0.0:
             raise ValueError(f"Invalid weight_decay value: {weight_decay}")
         if mu < 0.0:
```
MasterSkepticista, line 77: One liner `assert mu >= 0.0, f"FedProx regularizer coefficient must be greater than or equal to 0, got {mu}"` is sufficient. 
I don't think there are any scenarios where negative `mu` is used. It is OK to raise exception here instead of a warning.

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -20,6 +20,14 @@ class FedProxOptimizer(Optimizer):
...
     reduce the impact of devices with outlying updates.
 
+    IMPORTANT: This optimizer requires a reference to the original (global) model parameters
+    to calculate the proximal term. These must be set explicitly using the set_old_weights()
+    method before training begins. The old weights (w_old) must match the order and structure
+    of the model's parameters. Typically, w_old should be set to the initial global model
+    parameters received from the aggregator at the beginning of each round.
+
+    If mu > 0 and w_old is not set, the optimizer will raise a ValueError.
+
```
MasterSkepticista, line 30: Is it possible to record the first value of weights supplied to the optimizer as `w_old` at the end of iteration? User need not set this.

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -93,6 +109,47 @@ def __setstate__(self, state):
...
+            w_old: Old weights reference
+
+        Raises:
+            ValueError: If mu > 0 and w_old is None
+        """
+        if mu > 0 and w_old is None:
+            raise ValueError(
+                "FedProx requires old weights to be set when mu > 0. "
+                "Please call set_old_weights() before optimization step."
+            )
```
MasterSkepticista, line 126: This is assumed to be verified during init, no?

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -115,34 +172,53 @@ def step(self, closure=None):
             nesterov = group["nesterov"]
             mu = group["mu"]
             w_old = group["w_old"]
-            for p, w_old_p in zip(group["params"], w_old):
+
+            # Validate old weights for FedProx
+            self._validate_old_weights(mu, w_old)
+
+            # Apply proximal term when mu != 0
+            apply_proximal = w_old is not None
```
MasterSkepticista, line 180: Overall, consider a simpler approach: 

`apply_proximal` should always be true. We guarantee that `mu` will be `>=0.0` during initialization. Which means negative values never appear. As for `0.0` case, it means "no contribution" of the regularizer term. It becomes implied that it makes no effect on the weights.

openfl/utilities/optimizers/torch/fedprox.py
```diff
@@ -115,34 +172,53 @@ def step(self, closure=None):
...
+
+            for i, p in enumerate(group["params"]):
                 if p.grad is None:
                     continue
+
                 d_p = p.grad
+
+                # Apply weight decay
                 if weight_decay != 0:
                     d_p = d_p.add(p, alpha=weight_decay)
```
MasterSkepticista, line 190: Similar case: `weight_decay` can be guaranteed to be `>=0.0` during initialization and this check can be avoided

openfl/utilities/optimizers/torch/fedprox.py
kminhta: This is a good suggestion. I hadn't considered this. 

@nisha987 - the optimizer is supplied with the model parameters when it is first initialized. These parameters are updated during `optimizer.step()` . This is the global model and it should be possible to also record these weights (before updating) as `w_old` internally rather than having the user explicitly do it, as @MasterSkepticista suggests

Then we won't risk running into the issue that we saw in the fedprox mnist tutorial

PR #1635, opened 2025-05-19
Modify process_local_sources to support mount paths

Efrat1:
## Summary
Modify `process_local_sources` to support mount paths

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
Use common `base_dir` for all local data sources, and `source_path`s relative to that base.
The reason is to simplify path management in containerized environments, such as Docker. By using a common `base_dir`, we can ensure that paths remain consistent when mounting volumes, as only the `base_dir` needs to be adjusted to point to the mount path inside the container. This way, we only need to adjust the `base_dir` to point to the mount path.

- Cosmetic changes in histology_s3 dataloader.
- Change types names: "local" to "fs" and "azure_blob" to "ab"

## Testing
Run full experiment in OpenFL and in OpenFL-Security.

### DIFF
--- a/openfl/federated/data/sources/data_sources_json_parser.py
+++ b/openfl/federated/data/sources/data_sources_json_parser.py
@@ -15,7 +15,9 @@ from openfl.utilities.path_check import is_directory_traversal
 
 class DataSourcesJsonParser:
     @staticmethod
-    def parse(json_string: str) -> VerifiableDatasetInfo:
+    def parse(
+        json_string: str, label="", metadata="", check_dir_traversal=False
+    ) -> VerifiableDatasetInfo:
         """
         Parse a JSON string into a dictionary.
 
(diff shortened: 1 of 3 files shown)


teoparvanov (approved):
LGTM!

EDIT: additional changes and testing requested below, so please do not merge in the meantime.

openfl/federated/data/sources/data_sources_json_parser.py
```diff
@@ -31,48 +33,65 @@ def parse(json_string: str) -> VerifiableDatasetInfo:
...
+            label=label,
+            metadata=metadata,
         )
 
     @staticmethod
-    def process_data_sources(data):
+    def process_data_sources(data, check_dir_traversal=False):
         """Process and validate data sources."""
-        cwd = os.getcwd()
+        os.getcwd()
```
MasterSkepticista: This only returns `cwd`. If `cwd` is unused, line also can be removed.

openfl/federated/data/sources/data_sources_json_parser.py
```diff
@@ -31,48 +33,65 @@ def parse(json_string: str) -> VerifiableDatasetInfo:
...
-        path = params.get("path", None)
-        if not path:
-            raise ValueError(f"Missing 'path' parameter for local data source '{source_name}'")
-        abs_path = os.path.abspath(path)
-        rel_path = os.path.relpath(abs_path, cwd)
-        if rel_path and not is_directory_traversal(rel_path):
-            return LocalDataSource(source_name, rel_path, base_path=Path("."))
-        else:
-            raise ValueError(f"Invalid path for local data source '{source_name}': {path}.")
+    def process_local_sources(local_datasources, datasources, check_dir_traversal=False):
```
MasterSkepticista, line 70: I guess this is also supposed to be `staticmethod`?

openfl/federated/data/sources/data_sources_json_parser.py
```diff
@@ -31,48 +33,65 @@ def parse(json_string: str) -> VerifiableDatasetInfo:
...
+        """Process and validate local data sources."""
+        # The reason we use common base_dir and source_path relative to that base
+        # is to simplify path management in containerized environments, such as Docker.
+        # By using a common base_dir, we can ensure that paths remain consistent
+        # when mounting volumes, as only the base_dir needs to be adjusted to point
+        # to the mount path inside the container.
+        # This way, we only need to adjust the base_dir to point to the mount path.
+        absolute_paths = {
+            source_name: os.path.realpath(params.get("path", None))
```
MasterSkepticista: `realpath` would raise an exception if `path` is `None`. If we know params will always have `path`, better to set it as `params['path']` IMO

openfl/interface/collaborator.py
```diff
@@ -231,9 +231,11 @@ def register_data_path(collaborator_name, data_path=None, silent=False):
...
     help=(
         "Path to directory containing sources.json file defining the data sources of the dataset. "
-        "This file should contain a JSON object with the data sources to be registered. For 'local'"
-        " type, 'params' must include: 'path'. For 's3' type, 'params' must include: 'uri', "
-        "'access_key_env_name', 'secret_key_env_name', 'secret_name', and optionally 'endpoint'."
+        "This file should contain a JSON object with the data sources to be registered. For local "
+        "data source, 'type' is 'fs', and 'params' must include: 'path'. For 's3' type, 'params' "
+        "must include: 'uri', 'access_key_env_name', 'secret_key_env_name', 'secret_name', and "
+        "optionally 'endpoint'. For azure_blob, 'type' is 'ab', and 'params' must include: "
+        "'connection_string', 'container_name', and optionally 'folder_prefix'."
```
MasterSkepticista, line 238: My suggestion would be to point to OpenFL documentation URL that describes data sources in detail with examples, in addition to the format of JSON files.

openfl/interface/collaborator.py
```diff
@@ -258,7 +260,7 @@ def calchash(data_path):
         sys.exit(1)
     with open(datasources_json_path, "r", encoding="utf-8") as file:
         data = file.read()
-    vds = DataSourcesJsonParser.parse(data)
+    vds = DataSourcesJsonParser.parse(data, check_dir_traversal=True)
```
MasterSkepticista, line 263: Instead, can we `check_dir_traversal` before calling this function? I don't think the class needs to know about whether it is in a valid subdirectory...

teoparvanov:
Hi @Efrat1, after addressing @MasterSkepticista's comments, I'd request that you additionally test those changes by running a [containerized OpenFL federation](https://openfl.readthedocs.io/en/latest/about/features_index/taskrunner.html#docker-container-approach) locally.

openfl/interface/collaborator.py
Efrat1: OK. @teoparvanov what doc file should we insert that info to?

openfl/interface/collaborator.py
Efrat1: OK. That means we go over the json one more time to look for the local data sources. Makes sense? currently happens only in `calchash`

openfl/federated/data/sources/data_sources_json_parser.py
Efrat1: Right, fixed.

openfl/federated/data/sources/data_sources_json_parser.py
Efrat1: Yes, right.

openfl/federated/data/sources/data_sources_json_parser.py
Efrat1: sorry right. removed of course

PR #1649, opened 2025-05-22
Add function to deserialize the flower message for further introspection

kminhta:
## Summary

In the long term, we will want a mechanism that provides some control over the type of information that Flower is allowed to send across the network for security purposes.

In the short term, we introduce a mechanism to deserialize the flower message. If the flower message cannot be deserialized successfully, we will log a warning

## Type of Change (Mandatory)
- Feature enhancement  

## Description (Mandatory)
- introduce `deserialize_flower_message` function in `openfl/transport/grpc/interop/flower/message_conversion.py` to dynamically import the specified module and class from Flower message metadata, deserialize the message content, and return the deserialized object, logging errors if any step fails.
- when converting between a flower message to an openfl message (i.e. when the openfl interop component receives a message from the flower compoennt), first check if the Flower message can be deserialized and log a warning if not -> can potentially escalate into an error as we tighten the security (long term)

## Testing
Manual testing

### DIFF
--- a/openfl/transport/grpc/interop/flower/message_conversion.py
+++ b/openfl/transport/grpc/interop/flower/message_conversion.py
@@ -1,11 +1,13 @@
 # Copyright 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-from flwr.proto import grpcadapter_pb2
-from openfl.protocols import aggregator_pb2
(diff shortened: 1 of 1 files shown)


teoparvanov (approved):
LGTM!

PR #1651, opened 2025-05-26
Memory Optimizations

porteratzo:
**Title**: Optimize Memory Usage in Aggregator Task Runner Flow

**Summary**:  
Improve memory efficiency in the aggregator task runner flow to reduce resource usage and costs.

**Type of Change (Mandatory)**:  
- Feature enhancement

**Description (Mandatory)**:  
This PR applies several optimizations to reduce memory usage in the aggregator task runner flow. By reducing memory usage, we can use fewer resources when using OpenFL, resulting in lower costs.  
- **Efficient weighted averaging**: Instead of creating a tensor of tensors to get the average, we use a for loop to operate with one array at a time, saving substantial memory.
- **Close cursor**: Small effect but should be done anyway.
- **Reduce nested methods in send_local_task_results**: Persistent db logic is put into its own function. This allows liberating this memory once Persistent db tasks are completed instead of having it in memory until the end of send_local_task_results
- **Use np.asarray in get_tensors_by_filter**: Change `np.array` to `np.asarray` because `np.array` creates a copy instead of a view.
- **Load only keys in _save_model**: `og_tensor_dict` doesn’t utilize the values, only keys. Changed `utils.deconstruct_model_proto` so it can return a dict without the keys resulting in not creating them and having them in memory.

The combination of these changes results in lower peak memory usage, particularly during tensor aggregation, which can scale out of control with the number of collaborators and model size.

**Further analysis**
- **Save persistent db in chunks**: Originally, the persistent db saves task results by first creating a list of serialized tensors. Instead, we use a for loop to save results for each tensor, avoiding the creation of a large list of tensors. This optimization comes with a large latency penalty so it's implementation needs further polishing
- **Reduce nested methods in process_task_results**: `_end_of_round_with_stragglers_check` is called inside `process_task_results` which might leave memory from objects in `process_task_results` that are no long necessary until _end_of_round_with_stragglers_check is finished. It is not implemented because `process_task_results` has multiple paths that need to be analyzed.
- Aggregation still takes up most of the memory spike because it creates 3 versions of the model(previous round, current round, aggregated) which represents the largest increase in memory usage, perhaps a per layer clean up would keep this step smaller

**Testing**:  
Experiments where run on torch/mnist and torch/histology workflows, extensive profiling was run to detect regressions and memory usage. 
An exeriment with a custom model with two large layers (linear(10,000 X 10,000)) was conducted to compare improvements in memory usage the original implementation used 22486 MB vs 20531 MB with improvements, a 8.69% reduction in peak memory usage


openfl/protocols/utils.py
```diff
@@ -212,7 +212,7 @@ def construct_model_proto(tensor_dict, round_number, tensor_pipe):
     return base_pb2.ModelProto(tensors=named_tensors)
 
 
-def deconstruct_model_proto(model_proto, compression_pipeline):
+def deconstruct_model_proto(model_proto, compression_pipeline, just_keys=False):
```
kminhta: add `just_keys` arg to the docstrings

openfl/interface/aggregation_functions/sequentialweightedaverage.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+"""Federated averaging module."""
+
+import numpy as np
+
+from openfl.interface.aggregation_functions.core import AggregationFunction
+
+
+class SequentialWeightedAverage(AggregationFunction):
```
kminhta: general discussion: is there a scenario where we wouldn't want to to average weights sequentially? i.e. could we just eventually replace `WeightedAverage` with the `SequentialWeightedAverage` logic in a future release?

openfl/protocols/utils.py
kminhta: NIT: what do you think of the name `keys_only` instead? it would have a consistent nomenclature to `weights_only` which is used in torch

kminhta (approved):
LGTM - great additions towards scalability!

openfl/interface/aggregation_functions/sequentialweightedaverage.py
```diff
@@ -0,0 +1,51 @@
...
+            tags: tuple of tags for this tensor
+        Returns:
+            np.ndarray: aggregated tensor
+        """
+        accumulated = 0
+        total_weight = 0
+        for lt in local_tensors:
+            accumulated += lt.tensor * lt.weight
+            total_weight += lt.weight
+        return accumulated / total_weight
```
teoparvanov: You could equivalently write:
```python
    total_weight = sum(lt.weight for lt in local_tensors)
    weighted_sum = sum((lt.tensor * lt.weight for lt in local_tensors))
    return weighted_sum / total_weight
```

This seems to improve readability, and doesn't use additional memory.

openfl/interface/aggregation_functions/sequentialweightedaverage.py
teoparvanov: Good point, @kta-intel, I'm also in favor of replacing the `WeightedAverage` implementation directly. In fact, it's also "sequential", just with a different syntax (and unnecessary memory overhead due to the list comprehension used there). So, I don't see any advantage in keeping the "original", except probably that it's been very well tested.

openfl/databases/persistent_db.py
```diff
@@ -118,6 +118,7 @@ def save_task_results(
             )
             new_id = cursor.lastrowid
             self.conn.commit()
+            cursor.close()
```
teoparvanov: Maybe use `try/finally` to make sure the `cursor` object is properly disposed of, even in case of exceptions?

teoparvanov (approved):
LGTM, with a couple of additional suggestions!

openfl/databases/persistent_db.py
porteratzo: Included try/finally

openfl/interface/aggregation_functions/sequentialweightedaverage.py
porteratzo: Changed to suggestion

openfl/protocols/utils.py
porteratzo: Changed to keys_only

porteratzo: Added

psfoley (approved):

openfl/interface/aggregation_functions/weighted_average.py
```diff
@@ -9,14 +9,7 @@
 from openfl.interface.aggregation_functions.core import AggregationFunction
 
 
-def weighted_average(tensors, weights):
-    """Compute average."""
-    return np.average(tensors, weights=weights, axis=0)
-
-
```
psfoley: @porteratzo FYI - this function is used by other aggregation functions (geometric median): https://github.com/securefederatedai/openfl/actions/runs/15303941672/job/43051657843?pr=1651#step:4:275 . Would suggest adding this back in

porteratzo: The function was returned

openfl/interface/aggregation_functions/sequentialweightedaverage.py
porteratzo: Sequential weighted average replaced the original implmentation

porteratzo:
The deconstruct model proto just_keys is not working, will merge and include that one later

